PR #1, opened 2018-03-12
update authors with @gidden

gidden:


gidden:
Looks like CI needs to be set up here as well, @danielhuppmann ?

gidden:
CircleCI Added

PR #15, opened 2018-05-30
Install model

gidden:
This PR does lots of stuff:

- removes `ixmp` submodule
- installs model code with python package
- provides CLI for using user-defined directory for model code
- removes dynamic writing of `default_paths.py`
- adds a logger
- bumps version to 1.0.0


message_ix/cli.py
```diff
@@ -0,0 +1,13 @@
+import argparse
+
+from message_ix.config import config
+
+
+def config():
```
danielhuppmann: Would it make sense to also copy the tutorial folder to the new location, in addition to the `model` folder? And pack of folders within a `message_ix` folder?

gidden:
Hmm, it looks like CI stopped here..

gidden:
Now has a dl cli!

PR #20, opened 2018-06-02
remove ixmp-install from `install.bat`

danielhuppmann:
This PR removes the installation of ixmp (previously a submodule) from the installation using `install.bat`.

Also, it updates the path to the MESSAGE-GAMS version (folder moved for installation of GAMS code) in the sphinx documentation `config.py`.

Note that the GAMS code has a hard-coded version number (see `message_ix/model/version.gms`), which must match the version number in the compiled `ixmp.jar`.  This should be integrated with the repository version tag.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -64,7 +64,7 @@ copyright = u'2018, IIASA, Energy Program'
 author = u'IIASA, Energy Program'
 
 # retrieve MESSAGEix version number from model/version.gms
-fname = os.path.join('..', '..', 'model', 'version.gms')
+fname = os.path.join('..', '..', 'message_ix', 'model', 'version.gms')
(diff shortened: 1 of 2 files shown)


gidden:
Hey @danielhuppmann - I think part of this is done in #15. Should we pull that then rebase this one?

danielhuppmann:
Hey @gidden - this PR is going into the branch of #15, and it’s cleaning up two things that aren’t done there yet...

gidden:
ah, it looks like its on `iiasa/install-model` which I was using for the release, etc. `gidden/install-model` has more updates, but I can push it to either..

gidden:
Ok made sure the two are at the same commit.

gidden:
(which means conflicts.. sorry!)

PR #26, opened 2018-06-04
WIP: (do not merge!) beginning of new tutorial suite

gidden:
I'll fill this in later, but I wanted to give you an idea of what I'm thinking here @danielhuppmann

### DIFF
--- a/message_ix/utils.py
+++ b/message_ix/utils.py
@@ -1,6 +1,11 @@
+import collections
+import copy
+import itertools
 import logging
+import six
 
 import numpy as np
+import pandas as pd
 
 # globally accessible logger
 _LOGGER = None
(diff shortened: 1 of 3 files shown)


message_ix/utils.py
```diff
@@ -16,6 +21,105 @@ def logger():
...
+    """Returns True if x is a string"""
+    return isinstance(x, six.string_types)
+
+
+def isscalar(x):
+    """Returns True if x is a scalar"""
+    return not isinstance(x, collections.Iterable) or isstr(x)
+
+
+def add_spatial_sets(scenario, data):
```
danielhuppmann: I think that this syntax will lead to confusion in the medium term - why is it `scenario.function(...)` in some cases and `message_ix.function(scenario, ...)` in others?

message_ix/utils.py
```diff
@@ -16,6 +21,105 @@ def logger():
...
+    data = {'year': [2010, 2020]}
+    data = {'year': [2010, 2020], 'firstmodelyear': 2020}
+    """
+    if 'year' not in data:
+        raise ValueError('"year" must be in temporal sets')
+    horizon = data['year']
+    scenario.add_set("year", horizon)
+
+    first = data['firstmodelyear'] if 'firstmodelyear' in data else horizon[0]
+    scenario.add_set("cat_year", ["firstmodelyear", first])
```
danielhuppmann: This notation is deprecated, better to use `scenario.add_cat('year', 'firstmodelyear', first)`. By setting the kwarg `unique=True`, this will also remove any other entries from the mapping set `cat_year`, making sure that you don't have two first model years...

message_ix/utils.py
```diff
@@ -16,6 +21,105 @@ def logger():
...
+
+
+def vintage_and_active_years(scenario):
```
danielhuppmann: This approach creates (in most cases) far more data points than required - verbose and not not elegant. As an alternative, I'd suggest to use the function `years_active(node, tec, vintage)`, which returns the list of years in which this `technology` at a particular `node` built in year `vintage` can be active. On the contra side, this requires the parameter `technical_lifetime` to be set first - on the plus side, it takes into account changing period duration.

tests/test_utils.py
```diff
@@ -0,0 +1,111 @@
...
+def test_add_spatial_single(test_mp):
+    scen = test_mp.Scenario(*msg_args, version='new', scheme='MESSAGE')
+    data = {'country': 'Austria'}
+    utils.add_spatial_sets(scen, data)
+
+    exp = ['World', 'Austria']
+    obs = scen.set('node')
+    npt.assert_array_equal(obs, exp)
+
+    exp = ['World', 'global', 'country']
```
danielhuppmann: should `'World'` be really in here? I'd think that `World` is the node at the `global` level...

message_ix/utils.py
gidden: Possible, but this is the same as in Pandas. Other options exist, like subclassing `Scenario`, make a `ScenarioBuilder` object, etc. 

For this function, it may even be possible to just add it directly to `ixmp`, as I don't think there's anything message specific here. Or is there?

message_ix/utils.py
gidden: What happens if you add a first model year when one already exists?

message_ix/utils.py
gidden: We use this all throughout the tutorial. My goal here was simply reducing complexity there. I believe you yourself said this was distracting when message was taught in italy, right?

tests/test_utils.py
gidden: I was surprised too, but this passes the tests =)

gidden:
Ok, answers to your questions inline. I would propose the following:

1. move `add_spatial_sets` to `ixmp.Scenario`
2. add default required sets/params to be set in `ixmp` before solve (e.g., `firstmodelyear`)
3. *either* keep `vintage_and_active_years` years and `make_df` here or move to `ixmp`
   - `vintage_and_active_years` can be expanded in the future with kwargs that subset the years to return

Thoughts?

message_ix/utils.py
gidden: The more I think about it, the more I like subclassing..

message_ix/utils.py
danielhuppmann: then you have two, which obviously doesn't make sense - so you have to remove the previously existing one or use the `unique=True` flag.

message_ix/utils.py
danielhuppmann: When you implemented the first version of the Austria tutorial, the `years_active()` feature wasn't implemented yet.

gidden:
closing in favor of #27

PR #32, opened 2018-06-06
add notes on how to start tutorials

gidden:
@benham2015 does this align with your experiences?

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,14 +116,15 @@ opening a command prompt and typing
     conda install nb_conda
     ```
 
-2. Open Jupyter Notebooks from Anaconda's "Home" Tab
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -106,6 +106,32 @@ opening a command prompt and typing
    messageix-dl --local_path /path/to/tutorials
    ```
 
+## Running Tutorials
+
+### Using Anaconda
```
danielhuppmann: I'd think that the easiest way is to launch `Jupyter notebook` directly (not through Anaconda), then navigate to the tutorial folder.

gidden: I didn't see that option on my Mac after installing Anaconda (only the navigator). So you have to launch it through that...

gidden: Ok I think I addressed this now.

PR #46, opened 2018-06-12
Westeros

gidden:
Still needs a lot of touching up, but here's a first look @danielhuppmann


tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "slideshow": {
+     "slide_type": "slide"
+    }
+   },
+   "source": [
+    "The `COMMODITY_BALANCE` equation ensures that `input`, `output`, and `demand` are met at each `level` in the energy system\n",
```
danielhuppmann: Might also want to mention commodities here. Also note that `input` is the parameter, `input` times `ACT` is the amount of commodity created/produced/generated at that node. Better to make that distinction clear also for users working through the tutorial.

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+    "\n",
+    "light_demand = pd.DataFrame({\n",
+    "        'node': country,\n",
+    "        'commodity': 'light',\n",
+    "        'level': 'useful',\n",
+    "        'year': horizon,\n",
+    "        'time': 'year',\n",
+    "        'value': demand_per_year * demand_profile,\n",
+    "        'unit': 'GWa',\n",
+    "    })\n",
```
danielhuppmann: show the dataframe before adding it the scenario?

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+   "source": [
+    "The model has a number of \"reality\" constraints, which relate built capacity to available power.\n",
+    "\n",
+    "The `CAPACITY_CONSTRAINT` \n",
+    "\n",
+    "$\\sum_{m} ACT_{n,t,y^V,y,m,h}\n",
+    "    \\leq duration\\_time_{h} \\cdot capacity\\_factor_{n,t,y^V,y,h} \\cdot CAP_{n,t,y^V,y}\n",
+    "    \\quad t \\ \\in \\ T^{INV}$\n",
+    "    \n",
+    "TODO: needed? asked daniel, wait for reply\n",
```
danielhuppmann: `OPERATION_CONSTRAINT` is only needed if you use sub-annual timeslices, to distinguish between required down-time per `time` step compared to flexible down-time throughout the year - e.g., when you can decide whether to do maintenance for a plant in summer or winter.

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+   "cell_type": "markdown",
+   "metadata": {
+    "slideshow": {
+     "slide_type": "fragment"
+    }
+   },
+   "source": [
+    "The simplest way to model a pre-defined energy mix is to use equality constraints on activity or capacity.\n",
+    "\n",
+    "More elegant ways include using parameters to define `historical` values, but in this simple example, we choose a starting point for the Westerosi energy system.\n",
```
danielhuppmann: I'd go for the elegant way right away - it's one additional step, but it makes the results a lot more intuitive because it removes the big spike of new capacity in the first year

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "slideshow": {
+     "slide_type": "fragment"
+    }
+   },
+   "source": [
+    "However, `COST_NODAL` is comprised of a variety of costs related to the use of different technologies."
```
danielhuppmann: I'd remove the 'however'...

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "slideshow": {
+     "slide_type": "slide"
+    }
+   },
+   "source": [
+    "### Fixed O&M Costs\n",
```
danielhuppmann: You could add a line like 'Fixed cost are only relevant as long as the capacity is active. This formulation allows to include the potential cost savings from early retirement of installed capacity.'

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "slideshow": {
+     "slide_type": "slide"
+    }
+   },
+   "source": [
+    "## Commit the datastructure and solve the model"
```
danielhuppmann: Legacy term 'datastructure', should be 'scenario'

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
...i8gzXHsDygvGadNpWkgYDvSLi140tSNJESfMkzautrc2+UjMzswzlGa6qZ1psnSmVAT8CLm1qQRExJSIqI6KyvLw8wxLNzMyyl2e41gC9CsZ7Aq8WjO8NDAQeklQNHAHM8klNZmbW2uUZrk8DfSX1kdQOGAvMqpsZEW9GRLeIqIiICuAJ4NSImJdjTWZmZrnLLVwjYjNwITAHWATcGRELJE2WdGpe6zUzMyu13H6KAxARs4HZRdOubKDtMXnWYmZm1lJ8hSYzM7OMOVzNzMwy5nA1MzPLmMPVzMwsYw5XMzOzjDlczczMMuZwNTMzy5jD1czMLGMOVzMzs4w5XM3MzDLmcDUzM8uYw9XMzCxjDlczM7OMOVzNzMwy5nA1MzPLmMPVzMwsY7neLH1XUPHujFKXkKvqUhdgZrYLcs/VzMwsYw5XMzOzjDlczczMMuZwNTMzy5jD1czMLGO5hqukkZKqJC2RdHk98/9V0kJJz0v6X0kH5VmPmZlZS8gtXCW1AW4CTgQGAOMkDShq9hxQGRGHAXcDP8irHjMzs5aSZ891KLAkIl6KiE3AHcCowgYR8YeI2JCOPgH0zLEeMzOzFpFnuPYAlheM16TTGjIBuL++GZImSponaV5tbW2GJZqZmWUvz3BVPdOi3obSWUAl8MP65kfElIiojIjK8vLyDEs0MzPLXp6XP6wBehWM9wReLW4k6XjgCmBERGzMsR4zM7MWkWfP9Wmgr6Q+ktoBY4FZhQ0kDQZuBk6NiNU51mJmZtZicgvXiNgMXAjMARYBd0bEAkmTJZ2aNvsh0Am4S9J8SbMaWJyZmVmrketdcSJiNjC7aNqVBcPH57l+MzOzUvAVmszMzDLmcDUzM8uYw9XMzCxjDlczM7OMOVzNzMwy5nA1MzPLmMPVzMwsYw5XMzOzjDlczczMMuZwNTMzy5jD1czMLGMOVzMzs4w5XM3MzDLmcDUzM8uYw9XMzCxjDlczM7OMOVzNzMwy5nA1MzPLmMPVzMwsYw5XMzOzjDlczczMMuZwNTMzy1iu4SpppKQqSUskXV7P/PaSfpHOf1JSRZ71mJmZtYTcwlVSG+Am4ERgADBO0oCiZhOAtRFxCPAj4Oq86jEzM2spefZchwJLIuKliNgE3AGMKmozCrgtHb4bOE6ScqzJzMwsd21zXHYPYHnBeA3wyYbaRMRmSW8CXYH
```
danielhuppmann: The `plot_prices()` function is using the marginal of the `COMMODITY_BALANCE` equation - these prices are discounted. You could switch to the auxiliary variable 'PRICE_COMMODITY`, where GAMS does the 'undiscounting' for you.

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+       "<Figure size 432x288 with 1 Axes>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "beta = 0.7 # an elasticity of demand\n",
+    "gdp = pd.Series([1., 1.21631, 1.4108], index=pd.Index(horizon, name='Time'))\n",
```
danielhuppmann: do you know how these values were generated? might be good to be explicit here...

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+    "slideshow": {
+     "slide_type": "slide"
+    }
+   },
+   "source": [
+    "First we establish demand. Let's assume\n",
+    "\n",
+    "- 40 million people in [300 AC](https://atlasoficeandfireblog.wordpress.com/2016/03/06/the-population-of-the-seven-kingdoms/)\n",
+    "- similar population growth to Earth in the same time frame [(~factor of 12)](https://en.wikipedia.org/wiki/World_population_estimates)\n",
+    "- similar per capita demand for lighting as a developed country today (e.g., ~6000 GWh in Austria)\n",
```
danielhuppmann: could include a reference here to the public IEA energy statistics at http://www.iea.org/statistics/statisticssearch/report/?country=AUSTRIA&product=electricityandheat&year=2015 to give users an easy starting point to adapt to other countries

tutorial/westeros/westeros.ipynb
```diff
@@ -0,0 +1,1268 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "slideshow": {
+     "slide_type": "slide"
+    }
+   },
+   "source": [
+    "And finally, our power plants which are not resource constrained in this simple model thus have no `input`"
```
danielhuppmann: I wouldn't call this "not resource constrained", but "the model does not include the fossil resources used as inputs for coal_ppl". The costs of coal extraction are included in the variable costs." And wind obviously doesn't have an input, no need to phrase it in such a complicated way.

gidden:
note to self:

- [ ] add to test_tutorials.py
- [ ] update other `tools.py` to use top-level stuff

tutorial/westeros/westeros.ipynb
gidden: done (these are comments are out of order for some reason, so I'm just adding a note that I've addressed it)

tutorial/westeros/westeros.ipynb
gidden: done

tutorial/westeros/westeros.ipynb
gidden: yes, will replace with global gdp trajectory for clarity

tutorial/westeros/westeros.ipynb
gidden: done!

tutorial/westeros/westeros.ipynb
gidden: done!

tutorial/westeros/westeros.ipynb
gidden: done!

tutorial/westeros/westeros.ipynb
gidden: done

tutorial/westeros/westeros.ipynb
gidden: will do!

tutorial/westeros/westeros.ipynb
gidden: done!

tutorial/westeros/westeros.ipynb
gidden: done!

tutorial/westeros/westeros.ipynb
gidden: done!

danielhuppmann:
one more comment (working through the Westeros tutorial again) - the statement "While demand must be met, it can also be exceeded allowing the model to plan for future periods of demand for storable commodities." is wrong, the formulation as inequality has nothing to do with intertemporal storage. It's just a convention for easier parametrization.

gidden:
@danielhuppmann this should be good to go except for your final comment about demand. How would you suggest describing why its a geq than strict ==?

To note, this also needs to be updated in the presentation:

- [ ] update demand description
- [ ] update model solve api call

danielhuppmann:
how about the following (or a shorter version of it):
The ` COMMODITY_BALANCE` equation is formulated as an inequality. This implies that ` demand`  must be met, but supply of a commodity can exceed demand. The formulation implicitly assumes " zero cost of disposal", as is common in economics. This implementation simplifies calibration and is in line with conventions in energy systems modelling.

However, for modelling electricity systems, a formulation with equality is usually chosen, because electricity cannot be discarded easily.

gidden:
It looks like windows is giving us an interesting answer... namely 3.687727158588087e+23 vs.  3.600316973564438e+23 only on Python 2.7... no idea why.

gidden:
This is good to go now @danielhuppmann

danielhuppmann:
thanks @gidden!

One more suggestion - can we switch the cost parameters to " mUSD", to get the numerical values to a more human-intelligible range. This could hopefully also improve the numerical issues. You may need `mp.add_unit()` 

And one request - can you replace your email address "message_ix@iiasa.ac.at"?

gidden:
Ok, it appears that division was indeed the issue. Once all CI goes, should be good to merge @danielhuppmann

danielhuppmann:
thanks @gidden!

PR #50, opened 2018-06-15
Rename

gidden:
This one does a bit more than I expected:

- adds `rename_technology` function as discussed
- moves solve and clone into `message_ix.Scenario`
   - this is messier than I would have hoped given the pretty complicated logistics between the java objects, `ixmp.Scenario` and `ixmp.Platform.Scenario`
- moves all message-related database testing into `message_ix`


gidden:
@adrivinca this may interest you as well..

message_ix/core.py
```diff
@@ -107,3 +116,73 @@ def vintage_and_active_years(self):
         year_pairs = [(y_v, y_a) for y_v, y_a in combinations if y_v <= y_a]
         v_years, a_years = zip(*year_pairs)
         return v_years, a_years
+
+    def solve(self, **kwargs):
+        """Solve a Message Scenario. See ixmp.Scenario.solve() for arguments"""
```
danielhuppmann: please change to "MESSAGE Scenario" for consistency.

message_ix/core.py
```diff
@@ -9,7 +9,8 @@
 
 class Scenario(ixmp.Scenario):
 
-    def __init__(self, platform, model, scen, version=None, annotation=None, cache=False):
+    def __init__(self, platform, model, scen, version=None, annotation=None,
+                 cache=False, clone=None):
```
danielhuppmann: Is there a particular reason for passing the to-be-cloned-Scenario as a kwarg?

message_ix/core.py
```diff
@@ -107,3 +116,73 @@ def vintage_and_active_years(self):
...
+            ('slicing', only available for MESSAGE-scheme scenarios)
+        """
+        self._keep_sol = keep_sol
+        self._first_model_year = first_model_year or 0
+        model = self.model if not model else model
+        scen = self.scenario if not scen else scen
+        return Scenario(self.platform, model, scen, annotation=annotation,
+                        cache=self._cache, clone=self)
+
+    def rename_technology(self, from_tech, to_tech):
```
danielhuppmann: Suggest to refactor to `rename(set, mapping)`, where `set` is the name of the set in which you want to do the renaming and `mapping` is a dictionary...

message_ix/core.py
```diff
@@ -107,3 +116,73 @@ def vintage_and_active_years(self):
...
+                continue
+            df = self.par(item, filters={'technology': from_tech})
+            if not df.empty:
+                df['technology'] = to_tech
+                self.add_par(item, df)
+
+        # this removes all instances of from_tech in the model
+        self.remove_set('technology', from_tech)
+
+        # recommit
```
danielhuppmann: why `recommit`?

message_ix/core.py
gidden: Yes, this is necessary to hook everything up with `ixmp.Scenario`. It's messy because there's a strange pattern in `ixmp` which has two `Scenario` classes, one on the module, and one on the `Platform`.

message_ix/core.py
gidden: Sorry, is this question for why commit at all, or why the comment says recommit?

danielhuppmann: just the docstring

gidden:
But we are not just renaming sets, right? also parameters.. hence trying to
be more "general"

danielhuppmann:
well, not quite - we are renaming index set elements (elements of `technology`), and then the renaming is pulled through all mapping sets (e.g., `cat_tec`) and parameters.

PR #52, opened 2018-06-19
set model config in install.bat

danielhuppmann:
This PR adds the command `messageix-config --model_path` to `install.bat`, so that users cloning the GitHub repository and installing it (on Windows) use the GAMS code as cloned from Git and not as "installed" in `C:/ProgramData/Anaconda3/Lib/...`.

The output of this line is quite verbose because the CLI tries to copy files back to the folder where they came from, and then spits out a warning message that the file already exists. Still, an improvement for now...


gidden:
close in favor of #54

PR #53, opened 2018-06-20
add windows ci

gidden:


gidden:
Hold on this I need to rm some stuff, but wifi just failed

gidden:
looks like this will pass, should be good to pull after @danielhuppmann

danielhuppmann:
nice @gidden!

Issue #41, opened 2018-06-11
Cannot run messageix-config on Windows10 public desktop


ypwong22:
I am trying to install message_ix following the Install from Conda steps. Everything was fine until I ran the config model files command, i.e. 

messageix-config --model_path /path/to/model

The following error message is printed: 

Traceback (most recent call last):
  File "C:\Users\wangy\AppData\Local\Continuum\Anaconda2\envs\py36\Scripts\messageix-config-script.py", line 10, in <module>
    sys.exit(config())
  File "C:\Users\wangy\AppData\Local\Continuum\Anaconda2\envs\py36\lib\site-packages\message_ix\cli.py", line 65, in config
    do_config(model_path=args.model_path, overwrite=args.overwrite)
  File "C:\Users\wangy\AppData\Local\Continuum\Anaconda2\envs\py36\lib\site-packages\message_ix\cli.py", line 53, in do_config
    with open(CONFIG_PATH, mode='w') as f:
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\wangy\\.local\\message_ix\\config.json'

I could not find a .local folder in C:\\Users\\wangy.

gidden:
To the best of my knowledge, we need a 

```
path = os.dirname(CONFIG_PATH)
if not os.path.exists(path):
    os.makedirs(path)
```

Should be a good issue for a new person to tackle.

PR #54, opened 2018-06-20
add full ci support for messageix-config

gidden:
Hey @danielhuppmann, was already most of the way through this when I saw your #52... apologies.


gidden:
This should be good to go now, @danielhuppmann

appveyor.yml
```diff
@@ -39,6 +39,7 @@ build: false
 test_script:
   - conda install --yes -c conda-forge pytest
   - pytest -v tests
+  - messageix-config --model_path=message_ix\\model
```
danielhuppmann: shouldn't this first set the model folder, then run the tests?

.circleci/config.yml
```diff
@@ -8,14 +8,18 @@ jobs:
...
       - setup_remote_docker
       - run:
-          name: Build App Docker Image
+          name: Build App Docker Image for Python
           command: docker build -t message_ix/message_ix:latest .
       - run:
           name: Python2 Tests
-          command: docker run message_ix/message_ix:latest python2 -m pytest -v /message_ix/tests
+          command: |
+                docker run message_ix/message_ix:latest python2 -m pytest -v /message_ix/tests
```
danielhuppmann: shouldn't this first set the model folder, then run the tests?

gidden:
We actually already have unit tests for changing the model_path folder. What #41 identified was that the CLI on a fresh machine was broken. Thus, we simply here want to make sure that the CLI works.

PR #55, opened 2018-06-21
update manuscript reference, etc. in GAMS code

danielhuppmann:
I just noticed that we didn't update the reference to the manuscript pre-print in the entrypoint GAMS code files - mea culpa. This PR updates the reference and brings the GAMS code headers in line with the NOTICE.


gidden:
lgtm!

PR #56, opened 2018-06-21
clarify readme about config

gidden:


README.md
```diff
@@ -96,13 +96,20 @@ Then, open a command prompt in the new `ixmp` directory and type
 
 ## Configure Model Files
 
-By default, the model files are stored with your installation. If you would like
-to store them some place locally, open a command prompt and type
+By default, the model files are stored with your installation. Many users will
```
danielhuppmann: Does it make sense to add an example path here - I would not have known what "stored with your installation" means a little while ago...
Also, should this mention GAMS - again, it might not be obvious what "model files" are...

README.md
```diff
@@ -96,13 +96,20 @@ Then, open a command prompt in the new `ixmp` directory and type
...
+simply want to run MESSAGEix, and will never need to see the these files;
+however, some users will want to edit the files directly to add new parameters,
```
danielhuppmann: I'd rephrase that to "new types of parameters" to reduce (but not eliminate, unfortunately) the confusion between a parameter and its elements. Or maybe say "change the mathematical formulation"?

gidden:
I think the new changes address your suggestions

danielhuppmann:
lgtm, thanks @gidden

PR #57, opened 2018-06-21
removing deprecated "`ixmp` as submodule" from installation instructions

danielhuppmann:
Just noticed that the readme installations instructions still mention `ixmp` as a submodule. Updated the instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,18 +81,18 @@ the ix modeling platform for scientific publications or technical reports:
 
 ## Install from Source (Advanced Users)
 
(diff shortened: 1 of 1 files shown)


danielhuppmann:
@gidden, would it work out of the box if a user installs `ixmp` via `conda` and then uses `python setup.py install` for `message_ix`? Could there be some conflicts between Anaconda and other Python installations?

gidden:
No, this should work out of the box and is in fact how the new windows ci works. You can see that in `appveyor.yml`

PR #60, opened 2018-07-04
add ability to dump data directly to excel and read it back

gidden:
This depends on iiasa/ixmp#64

- [x] Tests Added
- [x] Documentation Added

# Adding to RELEASE_NOTES.md

```
```


danielhuppmann:
thanks @gidden, this looks like a really nice feature (for a short-term fix)... I'm sure @alexkoberle would very much appreciate this.

One idea to make it more stable - write out first sets that do not have index dimensions, then mapping sets, then parts - this way, no need to specify which sets to read in first, which is currently hard-coded...

alexkoberle:
thanks @gidden and @danielhuppmann. This does interest me. Seems not quite ready to go yet, but let me know how/if I can help.

gidden:
@danielhuppman, sadly I don't think that would work. When you read in an
excel file, you get a dictionary of sheetname: dataframe. This is unordered
and not related to how you wrote it in the first place.

behnam2015:
thanks @gidden for this. This i similar to what we did earlier in May. I have a suggestion:
I don't know if this is the case for this feature, but sometime the user may want to build a scenario from scratch using this read_excel feature. In thiscase, it's good to have the option of checking if sets and parameters in the Excel alredy exist in the newly created scenario, and if not, to initiate them (init_set and init_par) before importing the data.

gidden:
Hey Behnam - yes, I agree. I'm not sure of the exact use case, since we're
using the MESSAGE scheme, I think all basic sets and pars are already
init'd. I would be happy to see a PR that adds this feature though =)

behnam2015:
@gidden, you are right, it may be relevant to specific cases, for example, when calling MACRO-related parameters such as "aeei" and "grow" into a new scenario. Or it can be any sets and parameters added by the user to their original scenario.

behnam2015:
@gidden, Another thought is that you commit the sets first and then parameters. For example, sets and parameters can be stored originally in two different Excel files. It makes it easier to browse the Excel sheets as the number of sheets decreases in each file. In any case, sets can be called first and committed to the new scenario before commiting parameters.

behnam2015:
Something like:
----------------------------------------------------
set_list = ds.set_list()
xlsSetNames = xls_set.sheet_names               # List of the sets in the Excel file

sets = {}
sets['first'] = [x for x in xlsSetNames if len(xls_set.parse(x).columns) == 1]    # To check if set is not a datafarme
sets['second'] = [x for x in xlsSetNames if x not in sets['first']]

for set_name in sets['first']:
    df = xls_set.parse(set_name)         
    if set_name not in set_list:
        self.init_set(set_name, idx_sets= None, idx_names= None)
    self.add_set(set_name, df.iloc[:,0].values.tolist())

for set_name in sets['second']:
    df = xls_set.parse(set_name)
    if set_name not in set_list:
        self.init_set(set_name, idx_sets=df.columns.tolist(), idx_names=None) 
    self.add_set(set_name, df)

self.commit('sets added')

gidden:
Hey @behnam2015 thanks for the suggestion. I really think here we want everything stored as one file. This is more for interoperability and migration of data, not so much for ease of use for the user (this is more what @OFR-IIASA has been implementing, because users going through these sheets is already too high a bar).

I am still a little confused however. My understanding is that all of the pars/sets that could be initialized using the GAMS formulation already is when using `scheme='MESSAGE'`. I am almost 100% confident this is true for pars (otherwise, you would need new definitions of the par in the GAMS code, right?). Sets I do believe can be generated on the fly.

behnam2015:
@gidden, I see your point, and I agree, there is no need for initializing sets or pars in this workflow. So, we can store data in one Excel file, and for getting rid of those hardcoded set names, we may consider commiting the index sets that are lists and with one column in Excel first, then commiting datafarmes?

gidden:
@behnam2015, sorry, I don't quite follow your suggestion. I think this is fine as implemented, or should there be something further done?

gidden:
mind doing a review @behnam2015 ?

behnam2015:
Thanks @gidden for the cleanups. I checked it for one of the country-level models and it works flawlessly.

gidden:
then feel free to merge, my friend!

PR #63, opened 2018-07-09
added model='MESSAGE' to solve command to run

ClaraLuisa:


gidden:
Hi @ClaraLuisa, thanks for making a PR. I do not think this is needed, however, as the [`message_ix.Scenario.solve()`](https://github.com/iiasa/message_ix/blob/master/message_ix/core.py#L120) automatically adds this for you. Do you see some different behavior?

ClaraLuisa:
HEy @gidden , thanks for the fast reply. Without the model statement it runs into an error message asking for the model as an input. Do I need to update messageix?

gidden:
yes, this uses the latest version

gidden:
closing in favor of #65

Issue #66, opened 2018-07-09
Macro: solve method for message doesnt correspond with cplexopt


OFR-IIASA:
cplex.opt is configured as follows:
_* CPLEX LP solution method
* see https://www.gams.com/help/topic/gams.doc/solvers/cplex/index.html#CPLEXlpmethod
* 0     Automatic
* 1     Primal Simplex
* 2     Dual Simplex
* 3     Network Simplex
* 4     Barrier
* 5     Sifting
* 6     Concurrent

lpmethod = 4
threads = 4
epopt 1e-6_

When message initially solves, the barrier method is applied and solves quickly. After macro solves and runs message with updated demands, the primal simplex method is applied despite being configured otherwise. Maybe i dont understand how the barrier algorithm works, but the current solving of macro-initiated solving is very slow and may benefit from using a different solution method.  

See log below:
_--- MESSAGE-MACRO_run.gms(4019) 778 Mb
--- LOOPS iteration = 2
---   469,900 rows  398,378 columns  3,171,485 non-zeroes
--- Executing CPLEX (Solvelink=1): elapsed 0:02:32.683

IBM ILOG CPLEX   24.9.1 r63795 Released Aug 30, 2017 WEI x86 64bit/MS Windows 
--- GAMS/Cplex Link licensed for continuous and discrete problems.
Cplex 12.7.1.0

Reading parameter(s) from "C:\Repo\message_ix\message_ix\model\cplex.opt"
>>  * CPLEX LP solution method
>>  * see https://www.gams.com/help/topic/gams.doc/solvers/cplex/index.html#CPLEXlpmethod
>>  * 0     Automatic
>>  * 1     Primal Simplex
>>  * 2     Dual Simplex
>>  * 3     Network Simplex
>>  * 4     Barrier
>>  * 5     Sifting
>>  * 6     Concurrent
>> 
>>  lpmethod = 4
>>  threads = 4
>>  epopt 1e-6
Finished reading from "C:\Repo\message_ix\message_ix\model\cplex.opt"
*** Basis supplied by GAMS will be used.
*** Use option 'advind 0' to discard basis information.
Reading data...
Starting Cplex...
Space for names approximately 47.34 Mb
Use option 'names no' to turn use of names off
CPXPARAM_Advance                                 2
CPXPARAM_TimeLimit                               1000000
CPXPARAM_LPMethod                                4
CPXPARAM_Threads                                 4
CPXPARAM_Parallel                                1
[...]

PR #67, opened 2018-07-16
advanced basis turned off as default option to avoid conflicts with b…

OFR-IIASA:
…arrier method

- [ ] Tests Added
- [ ] Documentation Added

# Adding to RELEASE_NOTES.md

```
```

### DIFF
--- a/message_ix/model/cplex.opt
+++ b/message_ix/model/cplex.opt
@@ -8,6 +8,8 @@
 * 5     Sifting
 * 6     Concurrent
 
+* advanced basis turned off as default to avoid conflicts wehen using barrier method
+advind = 0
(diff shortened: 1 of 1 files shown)


danielhuppmann:
closes #66 

as discussed at the group meeting right now, being merged; documentation for other advanced GAMS options to be added as part of the documentation drive (see #71 )

volker-krey:
Switching off advanced basis in combination with the default lpmethod = 6 (concurrrent) is not too useful, because simplex (primal and dual) are usually part of the methods used by the concurrent threads. I would therefore suggest to make lpmethod = 4 (barrier) the default in cplex.opt.

PR #70, opened 2018-07-17
WIP - do not merge: generic MESSAGEix reporting class

danielhuppmann:

- [ ] Tests Added
- [ ] Documentation Added

# Adding to RELEASE_NOTES.md

```
```

This PR is intended to restart the discussion concerning the API of a generic MESSAEGix reporting class, based on existing work byn @OFR-IIASA and @ClaraLuisa (but as a new clean class to make the discussion more focused).

The use of this class is illustrated in the Westeros tutorial, and looks like

```
pp = message_ix.PostProcess(scenario)
pp.activity({'Wind': {'technology': 'wind_ppl'}})
pp.finalize()
```

where the first line initializes the postprocessing class, the second line is an example of how to translate results to an IAMC-style timeseries variable (only activity implemented now), and the third line commits the generated reporting to the database.

The `finalize()` function could also be used to perform aggregations over the nodes or across variables (e.g, from `Primary Energy|*` to `Primary Energy`).

One open question for me is whether all timeseries existing in the scenario (for example historical data before the first model year) should be loaded to the `PostProcess` instance at the beginning.

### DIFF
--- a/message_ix/__init__.py
+++ b/message_ix/__init__.py
@@ -5,6 +5,7 @@ import ixmp.model_settings as model_settings
 
 from message_ix import default_paths
 from message_ix.core import *
+from message_ix.reporting import *
 
 
 model_file = os.path.join(default_paths.model_path(), '{model}_run.gms')
--- a/message_ix/reporting.py
+++ b/message_ix/reporting.py
@@ -2,9 +2,11 @@
 
 import pandas as pd
 import numpy as np
+import collections
+import six
+
 import message_ix
 import ixmp
-import pyam
 
 
 IAMC_IDX = ['region', 'variable', 'unit', 'year']
@@ -76,6 +78,16 @@ class PostProcess(object):
 def _apply_filters(data, filters):
     keep = np.array([True] * len(data))
     for col, values in filters.items():
-        values = [values] if pyam.isstr(values) else values
+        values = values if islistable(values) else [values]
         keep &= [i in values for i in data[col]]
     return data[keep].copy()
+
+
+def isstr(x):
+    """Returns True if x is a string"""
+    return isinstance(x, six.string_types)
+
+
+def islistable(x):
+    """Returns True if x is a list but not a string"""
+    return isinstance(x, collections.Iterable) and not isstr(x)
--- a/message_ix/reporting.py
+++ b/message_ix/reporting.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+
+import pandas as pd
+import numpy as np
+import message_ix
+import ixmp
+import pyam
+
+
+IAMC_IDX = ['region', 'variable', 'unit', 'year']
+
+
+class PostProcess(object):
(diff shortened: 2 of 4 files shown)


danielhuppmann:
Strategic question here - should we make `pyam` a hard requirement for the postprocessing? For now, I'm only using `isstr`, but thinking ahead, we'll probably soon want nice features like auto-summation over the variable tree (like (`Primary Energy|*`), or even the plots.

@gidden, any thoughts?

ClaraLuisa:
Hey, I tried running the new postprocessing westeros example but the pp.finalize() command seems to be not working. Is this an error on my side?

image

gidden:
I see the same error as Clara.

gidden:
From an implementation point of view, is there anything other than `technology` that has an activity? If not, I would suggest simply providing a mapping and do

```
df = s.var('ACT')
df = df[df.technology.isin(mapping.from)
df['technology'] = df['technology'].map(mapping)
```

danielhuppmann:
I don't see how this syntax would work - can you give an example of how the mapping should look like?

gidden:
```
In [1]: import pandas as pd

In [2]: df = pd.DataFrame({
   ...: 'technology': ['a', 'a', 'b', 'b'],
   ...: 'value': [0, 1, 2, 3],
   ...: })

In [3]: df
Out[3]: 
  technology  value
0          a      0
1          a      1
2          b      2
3          b      3

In [4]: mapping = pd.DataFrame({
   ...: 'from_technology': ['a', 'b'],
   ...: 'to_technology': ['Wind', 'Solar'],
   ...: })

In [5]: mapping
Out[5]: 
  from_technology to_technology
0               a          Wind
1               b         Solar

In [9]: (df
           .merge(mapping, left_on='technology', right_on='from_technology')
           .drop(['technology', 'from_technology'], axis=1)
           .rename(columns={'to_technology': 'technology'})
           )

Out[9]: 
   value technology
0      0      Wind
1      1      Wind
2      2     Solar
3      3     Solar

```

danielhuppmann:
thanks @gidden for the detailed example - I thought that your comment was about the API, not the implementation. Your suggestion is indeed a lot nicer - but how would that work if one wants to filter by mode, too?

e.g., a mapping for a wind-power plant with a "high-flexibility" mode...
```
{
 'Wind|Standard operation': {'technology': 'wind_ppl', 'mode': 'standard'},
 'Wind|Flexibility operation': {'technology': 'wind_ppl', 'mode': 'flex'}
}
```

And I think the `finalise()` function should then add up these two variables to `Wind|Total`...

gidden:
It's hard to discuss the API with only one example (activity). Perhaps add a few more and we can flesh out the discussion a bit.

danielhuppmann:
fixed the bug identified by @ClaraLuisa and @gidden , removed `pyam` as dependency for now to get CI to pass

gidden:
closing for now (since there are multiple PRs with the same name), @danielhuppmann will reopen the correct one

PR #79, opened 2018-07-23
changed name of reticulate R package to retixmp consistent with name in ixmp 

volker-krey:


danielhuppmann:
thanks @volker-krey and @vruijven - but looking to the long term, what would be a good name for the `reticulate-ixmp` package? In particular if `rixmp` may be phased out `retixmp` sounds a bit odd...

volker-krey:
@danielhuppmann You are of course right that revisiting the R package discussion is important. Let's have that discussion when Adriano is back. The objective of the current pull request ist just to get this working again.

danielhuppmann:
ok, thanks y'all, going ahead and merging

PR #87, opened 2018-07-31
new formulation of the capacity maintenance 

ClaraLuisa:
The capacity maintenance was seperated into three seperate equations to ensure that capacity that is built during one periode also exists as cap by the end of the periode.


ClaraLuisa:
This PR changes the capacity maintenance equation to omit technologies that have a CAP_NEW not showing up in CAP by the end of the periode.

danielhuppmann:
thanks @ClaraLuisa, nice work! but for easier testing and review, can you please make a new branch from the latest master and cherry-pick this commit, and then make a PR directly into the master branch? After all, this change does not depend on any of the hackathon developments...

It seems that CircleCI fails because of an infeasibility, and as the PR now mixes the all-modes extension and your new-capacity formulation, it is difficult to tell why that happens...

ClaraLuisa:
sure, no problem!

ClaraLuisa:
closed in favour of #88

PR #89, opened 2018-07-31
full implementation of the renewables implementation (Sullivan et al.)

danielhuppmann:
# Please confirm that this PR has done the following:

- [x] Tests Added
- [x] Documentation Added
- [ ] Description in RELEASE_NOTES.md Added

# Adding to RELEASE_NOTES.md

Please add a single line in the release notes similar to the following:

```
- (#XX)[http://link-to-pr.com] Added feature which does something
```

PR #88, opened 2018-07-31
new formulation of the capacity maintenance

ClaraLuisa:
The capacity maintenance was seperated into three seperate equations to ensure that capacity that is built during one periode also exists as cap by the end of the periode.

PR #84, opened 2018-07-24
only return years greater than horizon

gidden:
# Please confirm that this PR has done the following:

- [x] Tests Added
- [x] Documentation Added
- [x] Description in RELEASE_NOTES.md Added

# Adding to RELEASE_NOTES.md

Please add a single line in the release notes similar to the following:

```
- (#XX)[http://link-to-pr.com] Added feature which does something
```

PR #83, opened 2018-07-24
addon-technology formulation

gidden:
# Please confirm that this PR has done the following:

- [x] Tests Added
- [x] Documentation Added
- [x] Description in RELEASE_NOTES.md Added

# Adding to RELEASE_NOTES.md

Please add a single line in the release notes similar to the following:

```
- (#XX)[http://link-to-pr.com] Added feature which does something
```

PR #91, opened 2018-08-01
Hackathon prs

gidden:
This is the main branch for all hackathon PRs. This is comprised of:

- #74 
- #80 
- #81 
- #83 
- #89 (#92 in actuality)
- #84 (#94 in actuality)
- #88 (#96 in actuality)
- #95
- #97
- #125 (updating compatibility with new `ixmp`)


gidden:
#106 included now in 7931fbf

gidden:
#107 included now in 14c3e4a

gidden:
#110 included now in e6aea82

gidden:
Hi @zikolach and @danielhuppmann I think I fixed the first issue. Now I get the following:

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
usr/local/lib/python2.7/site-packages/ixmp/core.py:81: in __init__
    self._jobj = java.ixmp.Platform("Python", dbprops)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <jpype._jclass.at.ac.iiasa.ixmp.Platform object at 0x7ff1a8e98410>
args = ('Python', '/tmp/kH07wz/test.properties')

    def _javaInit(self, *args):
        object.__init__(self)
    
        if len(args) == 1 and isinstance(args[0], tuple) \
           and args[0][0] is _SPECIAL_CONSTRUCTOR_KEY:
            self.__javaobject__ = args[0][1]
        else:
            self.__javaobject__ = self.__class__.__javaclass__.newClassInstance(
>               *args)
E           org.flywaydb.core.api.FlywayExceptionPyRaisable: org.flywaydb.core.api.FlywayException: Validate failed: Migration checksum mismatch for migration 1
E           -> Applied to database : 1588531206
E           -> Resolved locally    : 822227094
```

What is the right way to fix this? Do I need to regenerate the testdb? What will users have to do? How can they migrate their existing DBs to work with this version?

gidden:
Hi @zikolach / @danielhuppmann I am not at a loss. I updated the `properties` file, which seemed to fix that issue. I then tried to regenerate the test db with `testdb_setup.py`: this basically builds an HSQLDB from scratch and fills it with our test scenarios. This did not work, as I still get the checksum error. Could either of you please advise?

gidden:
Copying from #125 

Per my conversation with @zikolach, we have decided to proceed as follows:

```
Ok

So just to clarify my understanding

I'm not sure what users may or may not have applied

We know that the schema files had an issue between the two releases

But we now have a system in place to prevent that (notably they live in ixmp source)

So for this release, users may observe migration errors

And given that we have low confidence in whether they will be uniform across the user base

We should probably suggest they simply regenerate their databases
```

PR #80, opened 2018-07-23
Share constraints for level/commodity

gidden:
# Please confirm that this PR has done the following:

- [x] Tests Added
- [x] Documentation Added
- [x] Description in RELEASE_NOTES.md Added

# Adding to RELEASE_NOTES.md

Please add a single line in the release notes similar to the following:

```
- (#XX)[http://link-to-pr.com] Added feature which does something
```

PR #93, opened 2018-08-02
WIP - initial thoughts on generic share constraint

gidden:

- [ ] Tests Added
- [ ] Documentation Added

# Adding to RELEASE_NOTES.md

```
```


gidden:
for @danielhuppmann and @OFR-IIASA to comment. I know names of pars/etc with change, but what about the general formulation? can it be simplified? 

@danielhuppmann, the main issues here are that we need the equation to go across regions, so I keyed the is_input/output on both node types. I wanted to use one parameter, but saw that gams will see a 0 as false..

danielhuppmann:
Not sure how this is supposed to work and whether the mapping to `input` and `output` is necessary... But it seems to me like if this would want to implement a constraint for each region, the region info would have to be hardcoded into the name `shares`, like `['renewables_share_AFR', 'renewables_share_EUR', ...]`. This is probably not a good solution, because one would need to parse and decode the elements of the `shares` set. 

Suggestion for an alternative approach, building off the approach in #80:

add explicit mapping sets that state which type_tecs, nodes, etc. belong to the numerator and which belong to the denominator, like
```
share_factor_up('shares', 'node', 'year', 'time')
map_shares_commodity_level_share('shares', 'commodity', 'level', 'type_tec_share')
map_shares_commodity_level_total('shares', 'commodity', 'level', 'type_tec_share')
``` 
This would allow to have multiple entries into the mapping `map_shares_commodity_level_share` and it breaks the direct connection between commodity and level in the numerator and denominator. So that we could have constraints like "all electricity generated by renewables compared to all methane used by cars".

This formulation could be extended to include nodes and modes - but I'm not sure we really want to go that far?

gidden:
closing in favor of #95

PR #98, opened 2018-08-22
Redoc

gidden:
Ok, added API docs for message and ixmp. Also did a bit of cleanup to index.html so readers aren't given a complete information dump.


gidden:
I think so - I was building locally and browsing and it looked fine.

danielhuppmann:
yes, it looks fine, and there are hyperlinks to the API pages, so you can navigate to them. Seems that sphinx is just concerned if not all pages are included in toctrees...

PR #103, opened 2018-08-24
JUST A TEST Fix emission price reporting

gidden:
trying after rebasing


gidden:
looks like this one works @danielhuppmann I think you need to actually rebase yours

gidden:
also, feel free to just merge this one if you want after it passes

PR #106, opened 2018-08-29
updated addon formulation to exclude default conversion factor set to 1

OFR-IIASA:

- [ ] Tests Added
- [ ] Documentation Added

# Adding to RELEASE_NOTES.md

```
```

### DIFF
--- a/message_ix/model/MESSAGE/data_load.gms
+++ b/message_ix/model/MESSAGE/data_load.gms
@@ -139,8 +139,6 @@ capacity_factor(node,tec,year_all2,year_all,'year') =
 operation_factor(node,tec,year_all2,year_all)$( map_tec(node,tec,year_all)
     AND map_tec_lifetime(node,tec,year_all2,year_all) AND NOT operation_factor(node,tec,year_all2,year_all) ) = 1 ;
 
-* addon-conversion factor is not set to 1 by default by default to avoid modes for technologies being set to 1 although they shouldn't
-
 * set the upper bound on addon-technology activity to 1 by default
 addon_up(node,tec,year_all,mode,time,type_addon)$(
     map_tec_addon(tec,type_addon)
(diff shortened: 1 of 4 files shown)


gidden:
@OFR-IIASA This should be included in the [add-on commit](https://github.com/iiasa/message_ix/pull/91/commits/c5093cc1ffcde5644d520442be809044aa1a05b9) correct? Once this passes, I will rebase it into that commit to keep every feature addition to one commit (assuming @danielhuppmann agrees).

message_ix/model/MESSAGE/data_load.gms
```diff
@@ -139,12 +139,7 @@ capacity_factor(node,tec,year_all2,year_all,'year') =
 operation_factor(node,tec,year_all2,year_all)$( map_tec(node,tec,year_all)
     AND map_tec_lifetime(node,tec,year_all2,year_all) AND NOT operation_factor(node,tec,year_all2,year_all) ) = 1 ;
 
-* set the addon-conversion factor to 1 by default
-addon_conversion(node,tec,vintage,year_all,mode,time,type_addon)$(
-    map_tec_addon(tec,type_addon)
-    AND map_tec_act(node,tec,year_all,mode,time)
-    AND map_tec_lifetime(node,tec,vintage,year_all)
-    AND NOT addon_conversion(node,tec,vintage,year_all,mode,time,type_addon) ) = 1 ;
+* addon-conversion factor is not set to 1 by default by default to avoid modes for technologies being set to 1 although they shouldn't
```
danielhuppmann: @OFR-IIASA , "although they shouldn't" is really helpful to understand why this shouldn't be done

also, it would make more sense to insert this in the `parameter_def.gms` file where all parameters are described, rather than having a comment in `data_load.gms` that something is not done

OFR-IIASA: @danielhuppmann done

gidden:
So does this mean that for an addon technology to be "active" it must be declared and have a conversion factor?

OFR-IIASA:
yes

PR #109, opened 2018-09-04
Scenci

### LINKED ISSUE #108: Scenario Specific Unit Tests
We want to implement some tests that confirm changes in the code base do not affect basic baseline scenarios. This will require a number of steps.

First, we will want to make a `ci` folder inside of `message_ix/tests`. Inside, we want the following files:

- `scenarios.csv`: lists models, scenarios, expected objective function value
- `generate_scenario_db.[py|sh]`: connects to production database, creates new local database, copies models/scenarios in `scenarios.csv` to local database, uploads local database to internal IIASA hosted filesystem accessible from outside. This will require the `ixmp` branch `gidden:newjar`.
- `test_scenarios.py`: 
  - all tests should have a pytest CLI decorator to skip unless a specific flag is provided (e.g., pytest --ci)
  - all tests should have a check that downloads the localdb if not already there
  - single test for each scenario (either explicitly or generated on the fly, we can talk about the best strategy here)
- a `README.md` which explains what each file is for, how to regenerate the data, etc.

We will want to have all passwords required stored in environment variables. This can then be updated in `circleci`.

Finally, we will want to update `.circleci/config.yaml` to update the line which does `pytest tests --ci`

gidden:
Closes #108

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,14 @@ jobs:
                 docker run message_ix/message_ix:py2 /bin/bash -c 'source activate root && pytest -v /message_ix/tests'
                 docker run message_ix/message_ix:py2 messageix-config --model_path=/tmp/
       - run:
+          name: Python2 Scenario Tests
+          command: |
+                docker run \
+                -e MESSAGE_IX_CI_USER=$MESSAGE_IX_CI_USER \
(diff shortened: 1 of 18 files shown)


gidden:
@OFR-IIASA I rebased this into a single commit for easier managing. Please take note if you are building more commits on top of it.

PR #128, opened 2018-11-21
find version in setup.py the same way as in __init__.py

gidden:
needed for packaging of new release


gidden:
note circle is actually running [here](https://circleci.com/gh/gidden/message_ix/468)

PR #139, opened 2018-12-04
Add requirements file + update readme

zikolach:


khaeru:
Hi, so again the question is whether the dependencies should *also* be specified in the 'install_requires' kwarg to setup() in setup.py. This seems common in other packages.

E.g. pandas depends on python-dateutil:
1. [in a file used to set up the test environment](https://github.com/pandas-dev/pandas/blob/master/tox.ini) **and**
2. [in `setup.py`](https://github.com/pandas-dev/pandas/blob/master/setup.py#L29).

We seem to be missing (2).

khaeru:
Of the remaining items in install_requires:
1. `pandas`: this is genuinely a requirement of `message_ix`.
2. `jpype`: AFAICT this is a dependency of `ixmp`, not of `message_ix`, so probably remove.
3. `Xlsxwriter`: likewise, I think ([see here](https://github.com/iiasa/ixmp/blob/e84d154ce607fc3af0e9bf9025a992ced334325f/ixmp/utils.py#L38)).
4. `cython`, `xlrd`: I can't tell where these enter as dependencies. Can someone please explain?

We should probably have another PR adding the dependencies for ixmp in the same way.

khaeru:
One of the Travis failures looks like this:
```
Traceback (most recent call last):
  File "/Users/travis/miniconda/lib/python3.7/site-packages/pkg_resources/__init__.py", line 578, in _build_master
    ws.require(__requires__)
  File "/Users/travis/miniconda/lib/python3.7/site-packages/pkg_resources/__init__.py", line 895, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/Users/travis/miniconda/lib/python3.7/site-packages/pkg_resources/__init__.py", line 786, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (ixmp 0.1.3-1-g784c75e (/Users/travis/miniconda/lib/python3.7/site-packages/ixmp-0.1.3_1_g784c75e-py3.7.egg), Requirement.parse('ixmp>=0.1.3'), {'message-ix'})
```

When ixmp is installed from source (e.g. `git clone …` and then `pip install --editable .`), then the version string produced by versioneer is in the "git-describe" style: 0.1.3-1-g784c75e. ([List of all the styles](https://github.com/warner/python-versioneer/blob/master/details.md#how-do-i-select-a-version-style).)

In setup.cfg, I see the comment:
```
# this has to be git-describe (or some other varient without pluses), due to
# issues on loading java objects in windows with '+' in the path
style = git-describe
```

The issue here is that "0.1.3-1-g784c75e" is not PEP 440-compliant. So when setuptools/pip are trying to parse the requirement—added by this PR—for "ixmp>=0.1.3", then that dev version string does not actually satisfy the requirement.

Another of the versioneer styles that could be used for ixmp is "pep440-pre", which would produce strings like "0.1.3.post0.dev3" *and* still not include the `+` character. Would that break anything, or can I make the change in the other, ixmp PR?

khaeru:
> `Xlsxwriter` … `xlrd`: I can't tell where these enter as dependencies.

These are required to use `pandas.to_excel()` and `pandas.read_excel()` respectively (see [1](https://pandas.pydata.org/pandas-docs/stable/io.html#excel-writer-engines), [2](https://pandas.pydata.org/pandas-docs/stable/io.html#excel-files)). They are *optional* deps of pandas, but *required* by both ixmp and message_ix, since both use these methods.

> Another of the versioneer styles that could be used for ixmp is "pep440-pre" […] Would that break anything, or can I make the change in the other, ixmp PR?

I ended up making that change.

> (We shouldn't expect these checks to pass until [iiasa/ixmp#103](https://github.com/iiasa/ixmp/pull/103) is merged, so that installing the ixmp master branch pulls in its dependencies.)

Merging this caused the Travis and CircleCI tests to pass again :ok_hand: The Appveyor test can't be rerun without poking the branch somehow.

PR #161, opened 2019-02-04
Adding new years to a MESSAGE scenario

behnam2015:
**PR checklist**
- [x] Tests added.
- [x] Documentation added.
- [x] Description in RELEASE_NOTES.md added.


OFR-IIASA:
@behnam2015 please review scripts to make these pep8 conform. Once this is done, you will see that stickler-ci tests pass.

OFR-IIASA:
@behnam2015 If I understand correctly, then this script essentially makes a new scenario from an existing scenario by copying all individual sets and parameters. In my opinion, that seems like a large workload as opposed to just adding datapoints, where required to an existing scenario; and should also save a lot of time in terms of processing and probably also reduce the number of lines of code. Also, as opposed to having this as a standalone tool, I think it would be desirable to have this part of the message_ix class, so that you can simply call e.g. "scen.toolbox.add_year([2025])".  I also think that changing the firstmodelyear or changing the macro initial year shouldn't be part of the script's job.  The script should be limited to simply adding additional datapoints.  @gidden and @danielhuppmann would you agree? Are there any technical specifics that wouldn't allow this?

OFR-IIASA:
Talking to Daniel it seems that due to the implementation in gams and java, it is not possible to insert a year into an existing model. Everything needs to be defined from the beginning, as otherwise the new year being added would end up being inserted after the last year, as both of the above dont know about the correct sequence of years.

behnam2015:
@OFR-IIASA thank you for the feedback and evaluating the code. There was one pep8 issue I resolved it, hopefully it passes the test now.

behnam2015:
@OFR-IIASA as you mentioned, the year sequence was the reason that the new years are added to an empty scenario and all parameters are then copied from the reference scenario to the new one.

gidden:
would you mind taking a look here @khaeru? I will dig into #190

RELEASE_NOTES.md
```diff
@@ -115,3 +115,4 @@ jdbc.pwd = ixmp
 - [#65](https://github.com/iiasa/message_ix/pull/65): Bugfix for downloading tutorials. Now downloads current installed version by default.
 - [#60](https://github.com/iiasa/message_ix/pull/60): Add basic ability to write and read model input to/from Excel
 - [#59](https://github.com/iiasa/message_ix/pull/59): Added MacOSX CI support
+- [#161](https://github.com/iiasa/message_ix/pull/161): A feature for adding new time steps to a scenario
```
khaeru: This line should go under the section for "Next Release", not for "v1.1.0 / Updates".

behnam2015:
Thanks a lot @khaeru for the comments and for your great additions! The use of `click` for comman-line handling looks very good and concise. I reformatted the class to a function, now called `add_year`. Originally this was a class to reduce the number of arguments passed over different functions, but it seems it's not needed anymore.

RELEASE_NOTES.md
```diff
@@ -1,6 +1,7 @@
 
 # Next Release
 
+- [#161](https://github.com/iiasa/message_ix/pull/161): A feature for adding new time steps to a scenario
```
gidden: Can we please make the order numeric here? Both this one and 187 below. Thanks!

khaeru:
> 1. This introduces a new "standard" for calling CLIs (namely via the `-m` option). My preference here would be we choose a standard and stick to it. See e.g., `setup.py`:
> 
> ```
>    entry_points = {
>         'console_scripts': [
>             'messageix-config=message_ix.cli:config',
>             'messageix-dl=message_ix.cli:dl',
>         ],
> }
> ```
> 
> I have no strong preference between the options here, but would suggest we maintain consistency. Let me know what you think!

@behnam2015 and I used `click` in this PR to have a tidy CLI; but replacing the existing CLI was out of scope.

I think it should be a target for a future release to have a single entry point (`message-ix`) and then use `click` to dispatch to different commands. Once the overall CLI uses click, then the add_year tool/CLI can be converted to just one of the commands, and it can be called with `message-ix add_year [options] [args]`. For the moment, IMO it makes sense to keep it separate.

message_ix/tools/add_year/__main__.py
```diff
@@ -0,0 +1,170 @@
+"""Add additional years to a MESSAGE model/scenario instance
+
+\b
+Examples:
+$ python f_addYear.py --model_ref Austria_tutorial --scen_ref test_core \
```
gidden: is it still `f_addYear.py`?

gidden:
Ok, great advance here! Thanks all who helped out!

Issue #47, opened 2018-06-13
add option for strict equality of `COMMODITY_BALANCE` equation


danielhuppmann:
There are certain technologies/levels where an inequality for the `COMMODITY_BALANCE` (i.e., SUPPLY > DEMAND) is not applicable, like electricity or pollutants.

We need to define a new equation in the GAMS formulation that implements strict equality as an option.

The question at this point is how to "activate" the strict equality. Should this be contingent on the `commodity` only, or `(commodity, level)`, or also be specific on `node`, `year` or sub-annual `time`.

adrivinca:
for now, in our specific application with water technologies, we found it useful to have the equation contingent on `commodity,level`

adrivinca:
Shall we close this?

danielhuppmann:
this will be closed as part of release 1.2 (merged PR #172 into the release-candidate branch)

khaeru:
Closing as this is now in `master` and 1.2 will be released shortly.

PR #164, opened 2019-02-06
add =L= commodity balance

adrivinca:
#47 
addition to core model, with related set, slack variable and documentation

- [ ] Tests Added
- [ ] Documentation Added

# Adding to RELEASE_NOTES.md

```
```
For sure the nomenclature requires some more thinking.

NOTES: I did not edit the message_master_template.gms file
I did not test the documentation, I get errors when building it (also on master branch). Maybe I am not update with some latest requirements

### DIFF
--- a/message_ix/model/MESSAGE-MACRO_run.gms
+++ b/message_ix/model/MESSAGE-MACRO_run.gms
@@ -184,7 +184,9 @@ import_cost(node2, commodity, year) =
 * import into node2 from other nodes
           input(node2,tec,vintage,year,mode,node,commodity,level,time2,time)
         * duration_time_rel(time,time2) * ACT.L(node2,tec,vintage,year,mode,time2)
-        * COMMODITY_BALANCE.M(node,commodity,level,year,time) / discountfactor(year) )
+        * (COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) + 
+        COMMODITY_BALANCE_GT.m(node,commodity,level,year,time) - COMMODITY_BALANCE_LT.m(node,commodity,level,year,time))
+         / discountfactor(year) )
 ;
 
 * calculation of commodity export costs by node, commodity and year
@@ -194,7 +196,9 @@ export_cost(node2, commodity, year) =
 * export from node2 to other market
           output(node2,tec,vintage,year,mode,node,commodity,level,time2,time)
         * duration_time_rel(time,time2) * ACT.L(node2,tec,vintage,year,mode,time2)
-        * COMMODITY_BALANCE.M(node,commodity,level,year,time) / discountfactor(year) )
+        * (COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) + 
+        COMMODITY_BALANCE_GT.m(node,commodity,level,year,time) - COMMODITY_BALANCE_LT.m(node,commodity,level,year,time))
+         / discountfactor(year) )
 ;
 
 * net commodity trade costs by node and year
--- a/message_ix/model/MESSAGE-MACRO_run.gms
+++ b/message_ix/model/MESSAGE-MACRO_run.gms
@@ -184,9 +184,9 @@ import_cost(node2, commodity, year) =
 * import into node2 from other nodes
           input(node2,tec,vintage,year,mode,node,commodity,level,time2,time)
         * duration_time_rel(time,time2) * ACT.L(node2,tec,vintage,year,mode,time2)
-        * (COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) + 
-        COMMODITY_BALANCE_GT.m(node,commodity,level,year,time) - COMMODITY_BALANCE_LT.m(node,commodity,level,year,time))
-         / discountfactor(year) )
+        * ( COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) + 
+            COMMODITY_BALANCE_GT.m(node,commodity,level,year,time) - COMMODITY_BALANCE_LT.m(node,commodity,level,year,time) )
+            / discountfactor(year) )
 ;
 
 * calculation of commodity export costs by node, commodity and year
@@ -196,9 +196,9 @@ export_cost(node2, commodity, year) =
 * export from node2 to other market
           output(node2,tec,vintage,year,mode,node,commodity,level,time2,time)
         * duration_time_rel(time,time2) * ACT.L(node2,tec,vintage,year,mode,time2)
-        * (COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) + 
-        COMMODITY_BALANCE_GT.m(node,commodity,level,year,time) - COMMODITY_BALANCE_LT.m(node,commodity,level,year,time))
-         / discountfactor(year) )
+        * ( COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) + 
+            COMMODITY_BALANCE_GT.m(node,commodity,level,year,time) - COMMODITY_BALANCE_LT.m(node,commodity,level,year,time) )
+            / discountfactor(year) )
 ;
 
 * net commodity trade costs by node and year
--- a/message_ix/model/MESSAGE/model_core.gms
+++ b/message_ix/model/MESSAGE/model_core.gms
@@ -548,7 +548,7 @@ RESOURCE_HORIZON(node,commodity,grade)$( SUM(year$map_resource(node,commodity,gr
 * ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 *
 * Equation COMMODITY_EQUIVALENCE
-* """"""""""""""""""""""""""
+* """"""""""""""""""""""""""""""
 * This constraint accounts for the nodal and temporal balance of outputs/imports and inputs/exports for all commodities.
 *
 *  .. math::
@@ -589,7 +589,7 @@ COMMODITY_EQUIVALENCE(node,commodity,level,year,time)$( map_commodity(node,commo
     =E= COMM(node,commodity,level,year,time) ;
 
 * Equation COMMODITY_BALANCE_GT
-* """"""""""""""""""""""""""
+* """""""""""""""""""""""""""""
 * This constraint ensures the supply to be lower than or equal to the demand for every commodity and level
 * considered in Equation COMMODITY_EQUIVALENCE
 
@@ -601,7 +601,7 @@ COMMODITY_BALANCE_GT(node,commodity,level,year,time)..
         COMM(node,commodity,level,year,time) =G= 0;
 
 * Equation COMMODITY_BALANCE_LT
-* """"""""""""""""""""""""""
+* """""""""""""""""""""""""""""
 * This constraint ensures the supply to be greater than or equal to the demand for specific commodities and levels
 * defined with `balance_equality` at the nodal and temporal level
 *
--- a/message_ix/model/MESSAGE/model_core.gms
+++ b/message_ix/model/MESSAGE/model_core.gms
@@ -222,6 +222,8 @@ Set
 Positive variables
     SLACK_COMMODITY_BALANCE_UP(node,commodity,level,year_all,time) slack variable for commodity balance (upwards)
     SLACK_COMMODITY_BALANCE_LO(node,commodity,level,year_all,time) slack variable for commodity balance (downwards)
+    SLACK_COMMODITY_BALANCE_FULL_UP(node,commodity,level,year_all,time) slack variable for full commodity balance (upwards)
+    SLACK_COMMODITY_BALANCE_FULL_LO(node,commodity,level,year_all,time) slack variable for full commodity balance (downwards)
     SLACK_CAP_NEW_BOUND_UP (node,tec,year_all)        slack variable for bound on new capacity (upwards)
     SLACK_CAP_NEW_BOUND_LO (node,tec,year_all)        slack variable for bound on new capacity (downwards)
     SLACK_CAP_TOTAL_BOUND_UP (node,tec,year_all)      slack variable for upper bound on total installed capacity
@@ -251,7 +253,8 @@ Equations
     EXTRACTION_BOUND_UP             upper bound on extraction (by grade)
     RESOURCE_CONSTRAINT             constraint on resources remaining in each period (maximum extraction per period)
     RESOURCE_HORIZON                constraint on extraction over entire model horizon (resource volume in place)
-    COMMODITY_BALANCE               commodity supply-demand balance constraint
+    COMMODITY_BALANCE               commodity supply-demand lower balance constraint
+    COMMODITY_BALANCE_FULL          commodity supply-demand upper balance constraint
     STOCKS_BALANCE                  commodity inter-temporal balance of stocks
     CAPACITY_CONSTRAINT             capacity constraint for technology (by sub-annual time slice)
     CAPACITY_MAINTENANCE_HIST       constraint for capactiy maintainance  historical installation (built before start of model horizon)
@@ -422,6 +425,8 @@ COST_ACCOUNTING_NODAL(node, year)..
     + SUM((commodity,level,time)$( map_commodity(node,commodity,level,year,time) ), ( 0
 %SLACK_COMMODITY_BALANCE%   + SLACK_COMMODITY_BALANCE_UP(node,commodity,level,year,time)
 %SLACK_COMMODITY_BALANCE%   + SLACK_COMMODITY_BALANCE_LO(node,commodity,level,year,time)
+%SLACK_COMMODITY_BALANCE_FULL%   + SLACK_COMMODITY_BALANCE_FULL_UP(node,commodity,level,year,time)
+%SLACK_COMMODITY_BALANCE_FULL%   + SLACK_COMMODITY_BALANCE_FULL_LO(node,commodity,level,year,time)
         ) * 1e6 )
     + SUM((tec)$( map_tec(node,tec,year) ), ( 0
 %SLACK_CAP_NEW_BOUND_UP%    + 10 * SLACK_CAP_NEW_BOUND_UP(node,tec,year)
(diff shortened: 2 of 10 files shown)


gidden:
Hey @adrivinca it indeed looks like some tests fail here. You're also getting errors on the master branch? Let me know if I can assist.

adrivinca:
@gidden Do you mean for the documentation or the whole installation?
with building the documentation it does not work on the master branch either.

While the failure in tests is normal as I have not initialized the new set on the java side. 
I see a temporary solution you used for addon and share constraints you wrote the initialization in `core.py` , is it still a good way to follow?
Also, does the new equation need to be initialized?

gidden:
Yes, we will need to initialize the set in core.py for the message
scenario. I don't think the equation needs to be initialized.

Regarding tests, for the moment, let's just make sure the unit tests pass
(we can look at the docs later)

adrivinca:
ok, any suggestion on the names to be changed? we can do it later, but there will be more scripts to modify then

gidden:
I'll put some initial thoughts in line comments

message_ix/model/MESSAGE/model_core.gms
```diff
@@ -251,7 +253,8 @@ Equations
     EXTRACTION_BOUND_UP             upper bound on extraction (by grade)
     RESOURCE_CONSTRAINT             constraint on resources remaining in each period (maximum extraction per period)
     RESOURCE_HORIZON                constraint on extraction over entire model horizon (resource volume in place)
-    COMMODITY_BALANCE               commodity supply-demand balance constraint
+    COMMODITY_BALANCE               commodity supply-demand lower balance constraint
+    COMMODITY_BALANCE_FULL          commodity supply-demand upper balance constraint
```
gidden: `COMMODITY_BALANCE_FULL` -> `COMMODITY_BALANCE_LO` (to match other names)

message_ix/model/MESSAGE/model_core.gms
```diff
@@ -222,6 +222,8 @@ Set
 Positive variables
     SLACK_COMMODITY_BALANCE_UP(node,commodity,level,year_all,time) slack variable for commodity balance (upwards)
     SLACK_COMMODITY_BALANCE_LO(node,commodity,level,year_all,time) slack variable for commodity balance (downwards)
+    SLACK_COMMODITY_BALANCE_FULL_UP(node,commodity,level,year_all,time) slack variable for full commodity balance (upwards)
```
gidden: Here it gets tricky since there is already `SLACK_COMMODITY_BALANCE_UP`. Perhaps `SLACK_COMMODITY_BALANCE_LT_UP`? @danielhuppmann thoughts?

message_ix/model/MESSAGE/model_core.gms
```diff
@@ -222,6 +222,8 @@ Set
...
+    SLACK_COMMODITY_BALANCE_FULL_LO(node,commodity,level,year_all,time) slack variable for full commodity balance (downwards)
```
gidden: same here with whatever is decided above

message_ix/model/MESSAGE/sets_maps_def.gms
```diff
@@ -243,6 +246,7 @@ Sets
     type_emission                           types of emission aggregations
     cat_emission(type_emission,emission)    mapping of emissions to respective categories
     type_tec_land(type_tec)                 dynamic set whether emissions from land use are included in type_tec
+    full_balance(commodity)                 subset of 'commodity' to mark all commodities in the full commodity balance
```
gidden: just for clarity, I would suggest the name `commodity_balance_eq` here instead of `full` just because it's a bit more exact

gidden:
Ok, suggestions are in. @danielhuppmann I would highly suggest you provide input on the naming convention to make sure we are consistent with prior work here.

message_ix/model/MESSAGE/sets_maps_def.gms
adrivinca: or maybe commodity_balance_lt
and stick with '_gt' and '_lt' for the two commodity balances all over the model, also concerning the equation/slack variable names

gidden: The tricky thing here is that if a commodity belongs to this set it will be placed in *both* gt and lt at the moment. Perhaps it is best to get @danielhuppmann's thoughts on this.

message_ix/model/MESSAGE/model_core.gms
danielhuppmann: there is no need to duplicate the slack equations, they can be used for both the upper and lower part of the `COMMODITY_BALANCE` equation.

message_ix/model/MESSAGE/sets_maps_def.gms
danielhuppmann: agree with @gidden, whatever commodity (or probably better commodity-and-level?) is specified here must balance with equality.

message_ix/model/MESSAGE/model_core.gms
```diff
@@ -583,6 +588,47 @@ COMMODITY_BALANCE(node,commodity,level,year,time)$( map_commodity(node,commodity
 %SLACK_COMMODITY_BALANCE%   - SLACK_COMMODITY_BALANCE_LO(node,commodity,level,year,time)
     =G= 0 ;
 
+* Equation COMMODITY_BALANCE_FULL
```
danielhuppmann: It may at this point be easier to have an equation `COMMODITY_EQUIVALENCE` which equates the commodity balance to an auxiliary variable (similar to the variable `REL`), and then have much shorter lower and upper bounds on that.

gidden:
Thanks @danielhuppmann for your input! @adrivinca let me know if I can assist with making/looking at changes.

message_ix/model/MESSAGE/model_core.gms
adrivinca: I think this is a good idea. So we don't have to worry about exclusion-purpose sets. Also any possible future case of using only LT, EQ or GT balances are embedded.

adrivinca:
@gidden @danielhuppmann Updates:

One single commodity accounting equation + two inequalities
- Regarding the `PRICE_COMMODITY` reporting in `model_solve.gms`, it should still work the same now, correct me if I'm wrong

- I opted for inclusive subsets of commodities, rather than exclusion ones, it seems more straight-forward. Would you like to also add subsets of level? or a commodity and level mapping?

- Please comment on  nomenclature, I used '_GT' and '_LT' because UP and LO might be confusing with the slack variables

- STILL MISSING:

1. mathematical part of doc for the new equations
2. initialization on message_ix core.py

gidden:
> 
> 
> @gidden @danielhuppmann Updates:
> 
> One single commodity accounting equation + two inequalities
> 
>     * Regarding the `PRICE_COMMODITY` reporting in `model_solve.gms`, it should still work the same now, correct me if I'm wrong
need @danielhuppmann to comment on this one
> 
>     * I opted for inclusive subsets of commodities, rather than exclusion ones, it seems more straight-forward. Would you like to also add subsets of level? or a commodity and level mapping?
> 
I think in theory this is ideal; however, will it cause backwards compatibility issues with existing scenarios? My assumption is yes. I think either we need to use an exlcusion-style set for this equation **or** we do some operation before running that adds everything to the inclusion set. Perhaps others have a thought?
>     * Please comment on  nomenclature, I used '_GT' and '_LT' because UP and LO might be confusing with the slack variables
I think consistency with other equations (e.g., `BOUND_ACTIVITY_UP`) is still preferred, but happy to be overruled here. I personally find `GT` `LT` less cognitively difficult than `UP` `LO` FWIW.
> 
>     * STILL MISSING:
> 
> 
>     1. mathematical part of doc for the new equations
> 
>     2. initialization on message_ix core.py
Once we agree on your above points, let's prioritize this (over docs) to get the tests working.

danielhuppmann:
thanks @adrivinca for the efforts!

> Regarding the `PRICE_COMMODITY` reporting in `model_solve.gms`, it should still work the same now, correct me if I'm wrong

This is not correct. If you have an equality constraint, the marginals can be positive or negative, so you need to include the marginals from both the upper and lower bounds.

danielhuppmann:
> I opted for inclusive subsets of commodities, rather than exclusion ones, it seems more straight-forward. Would you like to also add subsets of level? or a commodity and level mapping?

Agree with @gidden: by default, the commodity should be an inequality (supply greater demand), and only if actively chosen should this be a strict equality. I don't think that we need to consider the use case demand greater than supply for now.

gidden:
1. change `commodity_balance` to `commodity_euivalence`
2. on `COMMODITY_BALANCE_GT` remove set `commodity_surplus`
3. on `COMMODITY_BALANCE_LT` change set name `commodity_shortage` to `balance_equality`
3. change `balance_equality` to be indexed on `commodity` **and** `level`
4. update https://github.com/iiasa/message_ix/blob/master/message_ix/core.py#L15 to

```
    inits = (
        # {
        #  'test': False  # some test,
        #  'exec': [(pass, {'args': ()}), ],
        # },
        {
          'test': 'balance_equality' not in s.set_list(),
          'exec': [(s.init_set, {'args': ('balance_equality', ['commodity', 'level'])})],
        },
```

5. Update `price_commodity` to include all appropriate marginals

gidden:
after CI passes, this LGTM. Should we try to think of a short test to run? or are we happy with this going in as is @danielhuppmann ?

message_ix/model/MESSAGE-MACRO_run.gms
```diff
@@ -184,7 +184,9 @@ import_cost(node2, commodity, year) =
 * import into node2 from other nodes
           input(node2,tec,vintage,year,mode,node,commodity,level,time2,time)
         * duration_time_rel(time,time2) * ACT.L(node2,tec,vintage,year,mode,time2)
-        * COMMODITY_BALANCE.M(node,commodity,level,year,time) / discountfactor(year) )
+        * (COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) +
```
danielhuppmann: please format similar to the other GAMS code (space around mathematical brackets, indenting)

message_ix/model/MESSAGE-MACRO_run.gms
```diff
@@ -194,7 +196,9 @@ export_cost(node2, commodity, year) =
 * export from node2 to other market
           output(node2,tec,vintage,year,mode,node,commodity,level,time2,time)
         * duration_time_rel(time,time2) * ACT.L(node2,tec,vintage,year,mode,time2)
-        * COMMODITY_BALANCE.M(node,commodity,level,year,time) / discountfactor(year) )
+        * (COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) +
```
danielhuppmann: formatting as above

message_ix/model/MESSAGE/model_solve.gms
```diff
@@ -47,7 +47,8 @@ EMISSION_CONSTRAINT.m(node,type_emission,type_tec,type_year)$(
 
 * assign auxiliary variables DEMAND, PRICE_COMMODITY and PRICE_EMISSION for integration with MACRO and reporting
     DEMAND.l(node,commodity,level,year,time) = demand_fixed(node,commodity,level,year,time) ;
-    PRICE_COMMODITY.l(node,commodity,level,year,time) = COMMODITY_BALANCE.m(node,commodity,level,year,time)
+    PRICE_COMMODITY.l(node,commodity,level,year,time) = (COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) +
```
danielhuppmann: please follow the GAMS formatting in the remaining code

message_ix/model/MESSAGE/model_core.gms
```diff
@@ -547,7 +547,7 @@ RESOURCE_HORIZON(node,commodity,grade)$( SUM(year$map_resource(node,commodity,gr
 * Constraints on commodities and stocks
 * ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 *
-* Equation COMMODITY_BALANCE
+* Equation COMMODITY_EQUIVALENCE
 * """"""""""""""""""""""""""
```
danielhuppmann: please add correct number of `"`

message_ix/model/MESSAGE/model_core.gms
```diff
@@ -583,31 +583,32 @@ COMMODITY_BALANCE(node,commodity,level,year,time)$( map_commodity(node,commodity
...
 * relaxation of constraints for debugging
-%SLACK_COMMODITY_BALANCE%   + SLACK_COMMODITY_BALANCE_UP(node,commodity,level,year,time)
-%SLACK_COMMODITY_BALANCE%   - SLACK_COMMODITY_BALANCE_LO(node,commodity,level,year,time)
+%SLACK_COMMODITY_EQUIVALENCE%   + SLACK_COMMODITY_EQUIVALENCE_UP(node,commodity,level,year,time)
+%SLACK_COMMODITY_EQUIVALENCE%   - SLACK_COMMODITY_EQUIVALENCE_LO(node,commodity,level,year,time)
 * equality to auxiliary variable COMM
     =E= COMM(node,commodity,level,year,time) ;
 
 * Equation COMMODITY_BALANCE_GT
 * """"""""""""""""""""""""""
```
danielhuppmann: please add correct number `"`

message_ix/model/MESSAGE/model_core.gms
```diff
@@ -583,31 +583,32 @@ COMMODITY_BALANCE(node,commodity,level,year,time)$( map_commodity(node,commodity
...
+*
 *
 ***
 
-COMMODITY_BALANCE_GT(node,commodity,level,year,time)$( commodity_surplus(commodity) )..
+COMMODITY_BALANCE_GT(node,commodity,level,year,time)..
         COMM(node,commodity,level,year,time) =G= 0;
 
 * Equation COMMODITY_BALANCE_LT
 * """"""""""""""""""""""""""
```
danielhuppmann: please add correct number `"`

message_ix/model/MESSAGE/sets_maps_def.gms
```diff
@@ -202,11 +202,9 @@ Alias(mode, mode2);
 *    * - type_tec_land (type_tec) [#type_tec_land]_
 *      - :math:`\widehat{t} \in \widehat{T}^{LAND} \subseteq \widehat{T}`
 *      - Mapping set of technology types and land use
-*    * - commodity_surplus (commodity)
-*      - :math:`c \in C^{FULL} \subseteq C`
-*      - Commodities related to `Equation COMMODITY_BALANCE_GT`_
-*    * - commodity_shortage (commodity)
+*    * - balance_equality (commodity,level)
```
danielhuppmann: please add description as with the other mapping sets in that table

message_ix/model/MESSAGE/sets_maps_def.gms
```diff
@@ -202,11 +202,9 @@ Alias(mode, mode2);
...
 *      - :math:`c \in C^{FULL} \subseteq C`
+*      - :math:`l \in L^{FULL} \subseteq L`
 *      - Commodities related to `Equation COMMODITY_BALANCE_LT`_
```
danielhuppmann: if this is related to the mapping set `balance_equality`, please write accordingly

message_ix/model/MESSAGE/sets_maps_def.gms
```diff
@@ -249,8 +247,7 @@ Sets
     type_emission                           types of emission aggregations
     cat_emission(type_emission,emission)    mapping of emissions to respective categories
     type_tec_land(type_tec)                 dynamic set whether emissions from land use are included in type_tec
-    commodity_surplus(commodity)            subset of 'commodity' to mark all commodities in Equation COMMODITY_BALANCE_GT
-    commodity_shortage(commodity)           subset of 'commodity' to mark all commodities in Equation COMMODITY_BALANCE_LT
+    balance_equality(commodity,level)       subset of 'commodity' and 'level' to mark all commodities in Equation COMMODITY_BALANCE_LT
```
danielhuppmann: rather than write "to mark all commodities in Equation COMMODITY_BALANCE_LT", why not write "mapping  of commodities-level where the supply-demand balance must be maintained with equality"

message_ix/model/MESSAGE_run.gms
```diff
@@ -81,7 +81,9 @@ import_cost(node2, commodity, year) =
 * import into node2 from other nodes
           input(node2,tec,vintage,year,mode,node,commodity,level,time2,time)
         * duration_time_rel(time,time2) * ACT.L(node2,tec,vintage,year,mode,time2)
-        * COMMODITY_BALANCE.M(node,commodity,level,year,time) / discountfactor(year) )
+        * (COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) +
```
danielhuppmann: please do formatting

message_ix/model/MESSAGE_run.gms
```diff
@@ -91,7 +93,9 @@ export_cost(node2, commodity, year) =
 * export from node2 to other market
           output(node2,tec,vintage,year,mode,node,commodity,level,time2,time)
         * duration_time_rel(time,time2) * ACT.L(node2,tec,vintage,year,mode,time2)
-        * COMMODITY_BALANCE.M(node,commodity,level,year,time) / discountfactor(year) )
+        * (COMMODITY_EQUIVALENCE.m(node,commodity,level,year,time) + 
+        COMMODITY_BALANCE_GT.m(node,commodity,level,year,time) - COMMODITY_BALANCE_LT.m(node,commodity,level,year,time))
```
danielhuppmann: please do formatting

message_ix/model/MESSAGE/sets_maps_def.gms
adrivinca: the description is below, do you mean something else

message_ix/model/MESSAGE/sets_maps_def.gms
adrivinca: of course it's related to the set `balance_equality`, it is its description. I also don't understand this comment, I will just add `level` to the description, if it's ok

gidden:
Hey @adrivinca which editor/IDE do you use to edit python code? We can try to find an auto-pep8 for that so you don't have to worry about stickler (or mostly don't have to worry..)

adrivinca:
I was using spyder with autopep, but was not highlighting this one that stickler says. Also i don't clearly understand what change it wants..

gidden:
Ok, talked with @adrivinca and he will fix this last stickler issue. 

The tests are now failing due to licensing issues. @danielhuppmann do you have any thoughts here? The basic problem is that we have now included more data (I guess?) which brings the austria tutorial over the size allowed by the trial GAMS license. 

While we could try some tricks here, I think we will continue to run into this problem. I see two primary options:

1. add a flag to nosetests that skips these tests if no license is found
2. move these tests to our "nightly" suite which already requires a license

Let me know what you think

gidden:
Ok, we now pass stickler. I can address the test failure, will try to do so later today. @danielhuppmann will do a final review of style/other issues, then we should be good to go!

gidden:
@adrivinca I just pushed to your branch adding a check for the gams license file

gidden:
closing in favor of #171

PR #170, opened 2019-02-22
remove cleaning from setup.py

gidden:
stop removing build/egg dirs as part of install, which makes pip fail on building wheels as seen by @volker-krey


khaeru:
:+1:, if only for simplification…

PR #181, opened 2019-03-22
no copying gdx files during install and setting of model directory

danielhuppmann:
This PR changes the setup and model-directory config such that `gdx` files are no longer copied back and forth.

### DIFF
--- a/message_ix/cli.py
+++ b/message_ix/cli.py
@@ -12,13 +12,11 @@ from six.moves.urllib.request import urlretrieve
 from message_ix.default_path_constants import CONFIG_PATH, DEFAULT_MODEL_PATH
 from message_ix.utils import logger
 
-GMS_EXT = ['.gms', '.opt', '.md']
 
-
-def recursive_copy(src, dst, overwrite=False, valid_ext=GMS_EXT):
+def recursive_copy(src, dst, overwrite=False, skip_ext=[]):
     """Copy src to dst recursively"""
     for root, dirs, files in os.walk(src):
-        for f in [_f for _f in files if os.path.splitext(_f)[1] in valid_ext]:
+        for f in [f for f in files if os.path.splitext(f)[1] not in skip_ext]:
             rel_path = root.replace(src, '').lstrip(os.sep)
             dst_path = os.path.join(dst, rel_path)
 
@@ -50,7 +48,8 @@ def do_config(model_path=None, overwrite=False):
         if not os.path.exists(model_path):
             logger().info('Creating model directory: {}'.format(model_path))
             os.makedirs(model_path)
-        recursive_copy(DEFAULT_MODEL_PATH, model_path, overwrite=overwrite)
+        recursive_copy(DEFAULT_MODEL_PATH, model_path, overwrite=overwrite,
+                       skip_ext=['gdx'])
         config['MODEL_PATH'] = model_path
 
     # overwrite config if already exists
--- a/message_ix/cli.py
+++ b/message_ix/cli.py
@@ -2,7 +2,6 @@ import argparse
 import json
 import message_ix
 import os
-import pathlib
 import shutil
 import tempfile
 import zipfile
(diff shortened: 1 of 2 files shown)


danielhuppmann:
@gidden, does it make sense to remove the logging here?

message_ix/cli.py
```diff
@@ -12,11 +13,13 @@
...
 from message_ix.utils import logger
 
+GMS_EXT = ['.gms', '.opt', '.md']
+
 
 def recursive_copy(src, dst, overwrite=False):
     """Copy src to dst recursively"""
     for root, dirs, files in os.walk(src):
-        for f in files:
+        for f in [_f for _f in files if splitext(_f)[1] in GMS_EXT]:
```
gidden: here I would prefer adding a new kwarg to the function like `skip_ext` and pass `GAMS_EXT` on the function call

gidden: and if we want to follow @khaeru's lead, we might want to start using the `Path` library, like

```
pathlib.Path(f).suffix
```

gidden: (same below)

gidden:
comments in line. remove which logging?

gidden:
@danielhuppmann can you rebase this one? @khaeru has fixed up the tests so they all should now pass with recent ixmp master

danielhuppmann:
rebased

message_ix/cli.py
danielhuppmann: done

danielhuppmann:
re @gidden and logging: `recursive_copy()` in `cli.py` writes a line to logger for every file copied. Seems excessive (even when not copying `gdx` files)...

danielhuppmann:
@khaeru gidden, seems that Python 2.7 tests were failing with `pathlib` and when I started digging, I got confused between `pathlib` and `pathlib2`... So I just went back to the previous file extension approach for now.

message_ix/cli.py
```diff
@@ -12,11 +12,13 @@
 from message_ix.default_path_constants import CONFIG_PATH, DEFAULT_MODEL_PATH
 from message_ix.utils import logger
 
+GMS_EXT = ['.gms', '.opt', '.md']
 
-def recursive_copy(src, dst, overwrite=False):
+
+def recursive_copy(src, dst, overwrite=False, valid_ext=GMS_EXT):
```
gidden: ah, here you have added a white list instead of a black list. I would instead suggest changing this to something like

```
def recursive_copy(src, dst, overwrite=False, skip_ext=[]): # now a black list
```

And then wherever this function is being called, specifically set the blacklist

```
recursive_copy(src, dst, skip_ext=['gdx'])
```

setup.py
```diff
@@ -33,12 +33,14 @@
     'tutorial': ['jupyter', 'matplotlib'],
 }
 
+GMS_EXT = ['.gms', '.opt', '.md']
+
 
 def all_subdirs(path, strip=None):
```
gidden: should rename this function if the behavior is changing

setup.py
```diff
@@ -33,12 +33,14 @@
...
     paths = []
-    for root, dirnames, filenames in os.walk(path):
-        for dirname in dirnames:
-            paths.append(os.path.join(root, dirname, '*'))
+    for root, dirnames, files in os.walk(path):
+        for f in [_f for _f in files if os.path.splitext(_f)[1] in GMS_EXT]:
```
gidden: and here since you only use `GMS_EXT` here, I would just make it a list in the function.

setup.py
```diff
@@ -64,8 +66,7 @@ def main():
         # for some reason the model/ directory had to be added separately
```
gidden: and you can remove this comment as I don't think it is applicable if you are explicitly adding filenames instead of directory globs

danielhuppmann: done

setup.py
danielhuppmann: done

setup.py
danielhuppmann: renamed to `all_gams_files()`

danielhuppmann:
thanks for the comments, @gidden - should be good to go!

gidden:
Looks great @danielhuppmann thanks!

PR #186, opened 2019-04-12
cleanup of Westeros tutorials

danielhuppmann:
# Description of this PR

This PR cleans up the Westeros suite of tutorials, in particular:
 - remove legacy elements of summer school presentation, use more generic header and summary instead
 - replace unit `%` by `-` where the values are absolute (0.05, not 5%)
 - consistency of section headers
 - miscellaneous rewriting

The PR also updates the reference to the MESSAGEix South Africa publication in the user guidelines.


tests/test_tutorials.py
```diff
@@ -8,6 +8,8 @@
 
 from conftest import here
 
+skip_py2 = True if sys.version_info[0] == 2 else False
```
gidden: This can simply be:

```
skip_py2 = sys.version_info[0] == 2
```

and would suggest renaming to `is_py2` or similar

tests/test_tutorials.py
```diff
@@ -74,6 +77,7 @@ def get_cell_by_name(nb, name):
     raise ValueError("no cell named '{}'".format(name))
 
 
+@pytest.mark.skipif(skip_py2, reason=py2_deprecated)
```
gidden: should we just add a check in `__init__.py` that raises an error if trying to run from py2? any thoughts on this @khaeru ?

gidden:
Hey @danielhuppmann. Content looks good. I have an open question on the python2/3 stuff.

tests/test_tutorials.py
danielhuppmann: done

tests/test_tutorials.py
danielhuppmann: This is just a quick fix to avoid the `from __future__ import division` in the tutorial.

I'd suggest to add proper warning messages when we remove py2-tests completely after release 1.2.

danielhuppmann:
@thanks Gidden, implemented first suggested, commented on why not to worry about second suggestion. Should be good to go once the unit tests pass.

gidden:
looks great, thanks @danielhuppmann

PR #189, opened 2019-04-26
A tutorial for representing sub-annual time steps in a MESSAGEix model

behnam2015:

- [ ] Tests Added
- [ ] Documentation Added

# Adding to RELEASE_NOTES.md

```
```


gidden:
Hi @behnam2015 this looks great! I really like the question/discussion section as well =). Can you please coordinate with @OFR-IIASA as discussed in the meeting and open a single PR with all of the new tutorials?

My goal will also be to get them explicitly on our website at the same time (similarly to how we do it in pyam: https://pyam-iamc.readthedocs.io/en/latest/tutorials.html)

behnam2015:
Thank you @gidden for the feedback! Isn't it better to have a separate PR for each tutorial to track the comments and commits more easily?

gidden:
Normally you want one PR per "feature" or "bugfix". However, given that there are ~8ish new tutorials, I think it would be easier on the reviewer to look through them all at once. It will also give us the opportunity to update our docs/CI, hence my preference.

Probably the best way to get started is simply pushing this branch to `upstream` so that both you and @OFR-IIASA can commit to it and it can easily be reviewed. At the same time I can do the doc refactor. Sound good?

behnam2015:
Thanks @gidden for the points! These tutorials are independent of each other and developed by different contributors. I'm afraid piling them all in one PR may not ease the process of review and approval. Also, the approval of all tutorials then has to wait until the review and approval of each and every of them. If the goal is harmonization between tutorials, I suggest we prepare a guide for that.

Issue #162, opened 2019-02-04
Implementation of explicit storage equations in MESSAGE


behnam2015:
**Motivation**:
There are different ways to represent storage (such as electricity storage, water storage, heat storage, etc.) using the current mathematical model in message_ix. However, it may be more efficient to add explicit equations represnting storage technologies and their functionality. This issue is opened to discuss the best way of doing this so that the added equations will serve as many as possible storage technologies and applications.

**Requirements**:
In this context, the desired equations should meet the following requirements:
- providing the possibility to invest in some storage technologies for shifting a specified commoditiy (like electricity) between subannual time steps
- showing the state of storage at any sub-annual time interval
- ensuring the state of storage is the same at the beginning and end of the year (any desire for inter-year storage?)
- represnting the storage losses ideally relative to the duration of storage (number of sub-annual time steps)
- being consistent with other equations, like COMMODITY_BALANCE

Please add your views and comments in this thread, and we try to synthesize the best possible way forewards.

behnam2015:
**Adriano's suggestion**:
This method is based on the changes in commodity balance equation and adding a new block of equations for storage:
1. Adding the following lines to COMMODITY_BALANCE to represnt quantity taken out from ( >0 ) or put into ( <0 ) inter-period stock (storage):
    _+ STOCK_CHG(node,commodity,level,year,time)$( map_stocks(node,commodity,level,year) )
    - STORAGE_CHG(node,commodity,level,year,time)$( map_storage(node,commodity,level,year,time) AND bound_storage_up(node,commodity,level,year,time) )_
![image](https://user-images.githubusercontent.com/30926636/52222384-b13f3280-28a3-11e9-9963-61c149e4e1f5.png)

2. Adding storage equations (including CAP variable and a possible mapping in the BOUND equations):
_STORAGE_BALANCE(node,commodity,level,year,time)$( map_storage(node,commodity,level,year,time) AND bound_storage_up(node,commodity,level,year,time) )..
    STORAGE(node,commodity,level,year,time)_
* fixed level of storage (i.e. in first period)
    _- commodity_storage(node,commodity,level,year,time) =E=
    ( storage_to_activity(time)*STORAGE_CHG(node,commodity,level,year,time) )_
* the next line is just saying STORAGE(t-1)
    _+ SUM((time2,year2)$seq_year_time(year2,year,time2,time), STORAGE(node,commodity,level,year2,time2) * 
[...]

PR #190, opened 2019-04-27
Implementation of generic storage equations

behnam-zakeri:
This PR is to summarize suggestions collected through discussions with @adrivinca and @JulianHunt4 in the issue <https://github.com/iiasa/message_ix/issues/162> for representing storage equations explicitly in the model. The concept and parameters presented in this PR are illustrated in the picture below (for more info, please see the accompanying documentation in this PR):
image

This PR has done the following:
- [x] Tests Added
- [x] Documentation Added


behnam2015:
I created a document to illustrate the storage concept presented in this PR and discuss the new equations suggested to be added to the model. I put it under a "temporary" folder in the doc folder. Please feel free to put in the right place.

RELEASE_NOTES.md
```diff
@@ -119,4 +119,5 @@ jdbc.pwd = ixmp
 - [#65](https://github.com/iiasa/message_ix/pull/65): Bugfix for downloading tutorials. Now downloads current installed version by default.
 - [#60](https://github.com/iiasa/message_ix/pull/60): Add basic ability to write and read model input to/from Excel
 - [#59](https://github.com/iiasa/message_ix/pull/59): Added MacOSX CI support
-- [#187](https://github.com/iiasa/message_ix/pull/187): Test for cumulative bound on emissions
\ No newline at end of file
+- [#187](https://github.com/iiasa/message_ix/pull/187): Test for cumulative bound on emissions
```
gidden: can you put these at the top please

tests/test_storage.py
```diff
@@ -0,0 +1,235 @@
...
+    scen.add_set('type_year', years)
+    scen.add_set('technology', ['tec1', 'tec2'])
+    scen.add_set('mode', 'mode')
+    output_specs = ['node', 'electr', 'level', 'year', 'year']
+    dict_var_cost = {'tec1': 1, 'tec2': 2}
+
+    for yr in years:
+        scen.add_par('demand', ['node', 'electr', 'level', yr, 'year'],
+                     1, 'GWa')
+        for t in dict_var_cost.keys():
```
gidden: everywhere you iterate through the keys, can be done like:

```
for tech, cost in dict_var_cost.items()
```

and then you can use the values explicitly

gidden: and if you want to get **real** fancy, these loops can be done like

```
import itertools # up at the top
for year, (tech, cost) in itertools.product(years, dict_var_cost):
```

tests/test_storage.py
```diff
@@ -0,0 +1,235 @@
...
+            scen.add_par('output', tec_specs + output_specs, 1, 'GWa')
+            scen.add_par('var_cost',
+                         tec_specs + ['year'], dict_var_cost[t], 'USD/GWa')
+
+
+# A function for adding sub-annual time steps to a MESSAGEix model
+def add_seasonality(scen, seasons_dict):
+    scen.add_set('time', list(set(seasons_dict.keys())))
+    scen.add_set('lvl_temporal', 'season')
+    for h in seasons_dict.keys():
```
gidden: updating to `items()`

tests/test_storage.py
```diff
@@ -0,0 +1,235 @@
...
+    model_setup(scen, [2020])
+
+    # Initiating missing sets/parameters if not yet in ixmp backend
+    if not storage_in_java:
+        init_storage(scen)
+    scen.commit('initialized test model')
+    scen.solve(case='no_seasonality')
+
+    # Second, adding seasonality
+    seasons_dict = {'a': 0.25, 'b': 0.25, 'c': 0.25, 'd': 0.25}
```
gidden: these are evenly valued - maybe it is worth testing non-even valued season durations?

tests/test_storage.py
```diff
@@ -0,0 +1,235 @@
...
+def add_seasonality(scen, seasons_dict):
+    scen.add_set('time', list(set(seasons_dict.keys())))
+    scen.add_set('lvl_temporal', 'season')
+    for h in seasons_dict.keys():
+        scen.add_set('map_temporal_hierarchy', ['season', h, 'year'])
+        scen.add_par('duration_time', [h], seasons_dict[h], '-')
+
+
+# A function for modifying model parameters after adding sub-annual time steps
+def year_to_time(scen, parname, dict_par, filters=None):
```
gidden: can you explain a bit more what this function is doing/why it is needed?

also, it looks like the `filters` argument is not used - can it be removed?

and finally, could you please format standard docstrings? so this would be:

```
def year_to_time(scen, parname, dict_par):
 """A function for modifying model parameters after adding sub-annual time steps (but now adding more explanation)"""
```

tests/test_storage.py
```diff
@@ -0,0 +1,235 @@
...
+                     1, 'GWa')
+        for t in dict_var_cost.keys():
+            tec_specs = ['node', t, yr, yr, 'mode']
+            scen.add_par('output', tec_specs + output_specs, 1, 'GWa')
+            scen.add_par('var_cost',
+                         tec_specs + ['year'], dict_var_cost[t], 'USD/GWa')
+
+
+# A function for adding sub-annual time steps to a MESSAGEix model
+def add_seasonality(scen, seasons_dict):
```
gidden: @behnam2015 do you think it would be worth it to make this a top-level `message_ix.Scenario` function? Right now we have an [`add_horizon()`](https://github.com/iiasa/message_ix/blob/master/message_ix/core.py#L209) function which helps you set up a model. I think it would be great to have this somewhere callable by others as well! It would need to do all the required work though, so something like the `year_to_time` below as well.

tests/test_storage.py
```diff
@@ -0,0 +1,235 @@
...
+        for index in time_idx:
+            new[index] = h
+        new['value'] = dict_par[h] * old['value']
+        scen.add_par(parname, new)
+
+
+# A function for initiating sets/parameters required for storage equations
+# NOTICE: this function can be deleted when storage parameters/sets are added
+# to ixmp java backend
+def init_storage(scen):
```
gidden: So we actually have support for this already in the code base~! Check out [this line](https://github.com/iiasa/message_ix/blob/master/message_ix/core.py#L26). I would suggest putting this there (feel free to stop by and I can show you how). Then when it gets into `ixmp`, we can take it out again. 

On the other hand, I would also be happy if we wanted to move this work all into python, but I'm not sure that's in the scope of this PR =)

adrivinca:
Thanks @behnam2015 for the PR, I only have some comments and suggestions on the GAMS side.

period_parameter_assignment.gms
I tried the model with different naming for the time elements (i.e non alphabetic order and with different numbers of letters) and there are two problems.
On the Python side, `list(set(seasons_dict.keys()))` is re-ordering the set alphabetically, this should not happen, so that one can have seasons like winter, summer and so on.
Fixing that (but I did it hard coding), then the new ordering is correct on the GAMS side.
However, I would suggest making time_seq a set and defining some basic sets that can be helpful in the future, such as first_time(time) last_time(time).

This is an example (not tested), but if we agree I can push changes to your branch:
```
Sets
first_time(time)
last_time(time)
;

first_time(time)$( ORD(time)$time_seq(time) = 1) = yes ;
last_time(time)$( ORD(time)$time_seq(time) = CARD(time_seq(time) ) = yes ;

* mapping of first and last time steps in a period AV
map_time_first_last(year_all,year_all2,time,time2)$( ( ORD(year_all) = ORD(year_all2)) AND
    first_time(time) AND  last_time(time) ) = yes;
```

Concerning `STORAGE_CHANGE ` equation, if the `level_storage` is always different from the normal commodity levels (e.g charger output = discharger input but different than charger input) then we don't need all the charger and discharger mappings and we can write a single line:

```
STORAGE_CHANGE(node,storage_tec,level,year,time)$( SUM( (mode,tec,commodity), map_tec_storage_level(node,tec,storage_tec,level,year,time) AND
    (map_tec_charge(node,tec,mode,commodity,level,year,time) OR map_tec_discharge(node,tec,mode,commodity,level,year,time) ) ) )..
* change in the content of storage in the examined timestep
    STORAGE_CHG(node,storage_tec,level,year,time) =E=
* increase in the content of storage due to the activity of charging technologies - decrease due to discharging
        SUM((location,vintage,mode,tec,commodity,time2)$(map_tec_lifetime(node,tec,vintage,year) AND ) AND map_tec_storage_level(node,tec,storage_tec,level,year,time) ),
      (  output(location,tec,vintage,year,mode,node,commodity,level,time2,time) - input(location,tec,vintage,year,mode,node,commodity,level,time2,time) ) * ACT(location,tec,vintage,year,mode,time) )
```
 but if we don't want to be so strict, and keep a level of freedom, I think the current implementation is ok.

in `STORAGE_BALANCE `Eq the if condition should not be with `storage_loss` otherwise the equation does not work then no losses is defined, probably `map_tec_storage_level` works well

Finally `bound_storage_up` and lo have the should have the same indexes as the `STORAGE` variable. 
So we should decide whether keeping the commodity index in STORAGE or removing it from the boundary parameters

if we agree on some of these points I can also contribute to the code

tests/test_storage.py
behnam2015: Thanks @Matt for the explanation!

RELEASE_NOTES.md
behnam2015: Thanks for the reminder! Done.

tests/test_storage.py
behnam2015: Thanks a lot @Matt for the suggestions! I improved the code.

tests/test_storage.py
behnam2015: Thanks Matt! Done.

tests/test_storage.py
behnam2015: Thanks @Matt for the suggestion, it's indeed needed! I added another test for inequal time steps.

tests/test_storage.py
behnam2015: @Matt! This function takes a MESSAGE parameter with the time index of 'year' and modifies the parameters for new "time" indexes, let's say months. It modifies the 'value' if needed too, based on the sahre of each time step from the yearly values. _filters_ could be used to load only a part of a parameter not all the table. As you noticed it's not used in this test and I removed it. Thanks!

tests/test_storage.py
behnam2015: @ Matt, thanks for the suggestion! I agree it will be a good addition. @adrivinca also suggested that the order of each time step can be specified through a set. Now, we specifiy the time order through a parameter: `time_seq`. So, `add_season` or something similar can be optionally used to wrap the information related to 
1. a subannual time step (set `time`)
2. hirerachy of that time step (set `lvl_temporal` and `map_temporal_hierarchy`)
3. Duration of that time step (parameter `duration_time`)
4. The order of that time step in a year (in this PR, parameter `time_seq`)

behnam2015:
Thank @gidden for the feedback and useful suggestions. I will move the documentation to the GAMS files as soon as we are happy with the final equations. 
As for `STORAGE_EQUALITY`, based on the feedback from you and others, I replaced this with an optional `RELATION_STORAGE` that can relate the content of storage between two time steps (and even between two periods). This is now represented as an inaquality to be more flexible (instead of equality). I will continue improving the PR if you have further thoughts.

behnam2015:
Thanks a lot @adrivinca for feedback and suggestions. I try to address them one by one:
- related to period_parameter_assignment and `time_seq`, I'm quite open if we decide to go for a set for adding the time order. Now, after removing `STORAGE_EQUALITY' we don't need to specify or map the first and last time steps anymore.
- related to `list(set(seasons_dict.keys()))`, the order doesn't affect the results, as the duration will be specified for each time step directly, and this is only for the test_storage.py. I modified the test to sort the time steps alphabetically, but if a user wants to use this test for declaring other time names, they should do that explicitly, and not through `dict.keys()`.
- related to 'STORAGE_BALANCE', it is now mapped based on `map_tec_storage_level`.
_ commodity was removed from the bounds and `storage_loss`.
Please have a look again and suggest improvements if needed.

gidden:
Hi all - just getting back and catching up. @behnam2015, does this need another review now?

behnam2015:
@gidden, nothing from my side, @adrivinca may want to suggest some additions.

gidden:
Hi folks, it looks like we have some conflicts in the GAMS code. I would suggest @behnam2015 @adrivinca and @danielhuppmann decide whether this is worth trying to get into the next release or should be a first PR into master after the release is finalized. I'm happy either way if all agree.

behnam2015:
Thanks @gidden! I rebased and resloved the GAMS related conflicts. Also I modified the equations for the following changes:
- Shifting commodity between two sub-annual time steps that are only in one temporal level
- Initialization of storage (adding free content at any time step needed), follwoing discussions with @adrivinca and @JulianHunt4

gidden:
@danielhuppmann can you please take a look at the GAMS code here to make sure it is in line with our preferred style/syntax? Also, specifically at 90faefe which adjust `test_legacy_scenario.py`?

tests/test_storage.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # II.4. Energy balance: storage change + losses = storage content
+        storage = scen.var('STORAGE').set_index(['year', 'time'])['lvl']
+        assert storage[(2020, 'b')] * (1 - loss[(2020, 'b')]
+                                       ) == -change[(2020, 'c')]
+
+
+def test_storage(test_mp):
+    '''
+    Testing storage setup with equal and inaqual duration of seasons"
```
danielhuppmann: Typo: this should be "unequal"

message_ix/model/MESSAGE/model_core.gms
```diff
@@ -1981,6 +1992,146 @@ RELATION_CONSTRAINT_LO(relation,node,year)$( is_relation_lower(relation,node,yea
...
+STORAGE_BALANCE(node,storage_tec,level,year,time2)$ (
+    SUM(tec, map_tec_storage_level(node,tec,storage_tec,level,year,time2) )
+    AND NOT (time_seq(time2) = 1 ) )..
+* Showing the content of storage at each timestep
+    STORAGE(node,storage_tec,level,year,time2) =E=
+* initial content of storage and change in the content of storage in the examined timestep
+    init_storage(node,storage_tec,level,year,time2)
+    + STORAGE_CHG(node,storage_tec,level,year,time2)
+* storage content in the previous subannual timestep
+    + SUM((time,year2)$map_time_period(year,year2,time,time2),
```
danielhuppmann: I believe that this line is wrong - why would you include in the sum the `STORAGE` variable from the current year but include `year2` in the summation? Also, `year2` is not mentioned in the documentation...

tests/test_storage.py
behnam-zakeri: Thanks @Daniel, corrected.

message_ix/model/MESSAGE/model_core.gms
behnam-zakeri: @danielhuppmann, sorry for being late in coming back to this. The second index of year (`year2`) was initially intended for cases of shiftIing a commodity between two periods. Later, we decided using storage only for inter temporal commodity shift in one period. So, thanks for noticing this, `year2` was extra and I removed it.

behnam-zakeri:
@khaeru, the tests in this PR fail because the new sets and parameters defined here are not initialized. Before rebasing this initialization was done in _core.py_. Should we wait for the PR [ixmp#122](https://github.com/iiasa/ixmp/pull/212) before refactoring the initialization here?

khaeru:
Rebased now that iiasa/ixmp#212 is merged.

behnam-zakeri:
@khaeru, thanks editing `model.py` for new sets and parameters in this PR. Looking at the failed tests, it seems whenever a scenario is cloned in tests, the test is failing. In other words, cloning doesn't include the initialization of new sets and parameters. Because of that the cloned scenario fails to solve. Have you tested this in `ixmp.Model.initialize_items()`?

khaeru:
> Looking at the failed tests here, it seems whenever a scenario is cloned in a test, the test is failing. In other words, cloning a scenario doesn't include the initialization of new sets and parameters introduced in this PR through `model.py`. Because of that the cloned scenario fails to solve.

Thanks for investigating and diagnosing. That's a good catch.

> Have you tested this behaviour in `ixmp.Model.initialize_items()`?

No. To explain, the `message_ix.Scenario.clone` behaviour is, for now, implemented in the closed-source Java code behind `ixmp.JDBCBackend`.

Intuitively (I think) cloning any Scenario (either ixmp.Scenario or message_ix.Scenario) should copy *all* of its contents.

However, the Java code (see [here](https://github.com/iiasa/ixmp_source/blob/6544c5a9f1d2a37d3270c9de9298fcbba209b608/src/main/java/at/ac/iiasa/ixmp/objects/MsgScenario.java#L149-L264)) does a lot of other stuff, most of which is not publicly documented. In the process, I worry that it may be using some internal/hidden list, and throwing away items (parameters) that are not in that list—such as the ones added by this PR.

This kind of issue is why we have #254. To use our analogy of the messy closet, I think you have identified that my pulling out iiasa/ixmp#212 caused something else to fall on the floor, which we must now pick up.

What I can do is:
1. Make a small, separate PR that tests whether the Java clone code is excluding some data.
2. Move the 'special' message_ix clone behaviour out of Java into Python, and then fix it so it doesn't do this.

@danielhuppmann and @zikolach know the Java code best, so might be able to confirm (1).

behnam-zakeri:
@khaeru thanks for the detailed reply. I may have missed something here. When this PR was made initially, it could pass the tests. At that time, the initialization of new items (sets and parameters) of this PR was handled in _message_ix.core.py_. The idea was that when initiating a MESSAGE scenario (either by loading an existing one and cloning, or creating a new one) a function in _core.py_ would initialize the new items and scenario was solved without a problem. This meant to me the backend didn't filter any parameter/set names, as the GDX from the scenario was passed to GAMS with the new items. In fact, this function in _core.py_, now in _model.py_, for adding new items work as an interim solution until the PR is accepted. After accpeting PR, these new items cacn be added to any default list, either in python or backend, which is your concern I guess.

behnam-zakeri:
@khaeru, to address the remaining issues here, I still can't pass the tests in which a scenario is cloned. It seems the changes you suggested in `message_ix.models.py` under this PR, i.e., the initialization of new sets and parameters specific to this PR, is applied only when a new scenario is initialized. However, these new sets and parameters are not added to a scenario if cloned from an old one. This is the case in `test_solve_legacy_scenario` and `test_macro`. can you please have a look?

khaeru:
> @khaeru, to address the remaining issues here, I still can't pass the tests in which a scenario is cloned. It seems the changes you suggested in `message_ix.models.py` under this PR, i.e., the initialization of new sets and parameters specific to this PR, is applied only when a new scenario is initialized. However, these new sets and parameters are not added to a scenario if cloned from an old one. This is the case in `test_solve_legacy_scenario` and `test_macro`. can you please have a look?

Yes, will do. Thanks for the summary!

khaeru:
FYI all, I rebased this branch on `master` and cleaned up duplicate commits. It's down from 77 to 39 commits.

adrivinca:
thanks all for pushing this forward.
I went through the implementation again. form the GAMS point of view I think it is fine to merge.

Only have minor comments on things that perhaps could be simplify:

- What is a practical use-case for the `STORAGE_RELATION`, has it ever been used and tested? if not, I would keep it out for the moment

- does the mapping set `map_tec_storage_level(node,tec,storage_tec,level,year,time)` need to have all those dimensions? I think mapping sets should be as simple as possible, so if we map a technology to a level, are there reasons why this mapping would differ by node, year and time? 
if not I would edit those sets, and check it does not break the if conditions.

I did not have time to work on a tutorial, but interested in it

behnam-zakeri:
Thanks @adrivinca for the review. I made some cleanup and removed some extra mappings. 
- I combined _map_tec_storage_level_ with _map_tec_storage_ and removed indexes of _time_ and _year_. But it's useful to map storage technologies with the _level_ and _commodity_ of storage.
- I removed extra sets and mappings, incl. _charge_tec_, _discharge_tec_, _map_charge_tec_, and _map_discharge_tec_.
- I removed _bound_storage_lo_ and _bound_storage_up, and related the state of charge of storage (_STORAGE_) to activity of the corresponding technology. This way we can use the available bounds in current formulation.
- _STORAGE_RELATION_ is used to relate the state of charge in two different time steps. It can be used, for example, to specify that the content at the last time step should be greater than 40% of initial value. The use of this parameter is optional.
- I improved the documentation for clarity, and improved the tests to reflect these changes.

khaeru:
Per discussion in today's MESSAGE meeting:

- [ ] @khaeru: address failure of test_legacy_version.
- [ ] @francescolovat: review/copyedit documentation of new sets and parameters.

khaeru:
> - [ ] @khaeru: address failure of test_legacy_version.

@behnam-zakeri I had to make the modifications in iiasa/ixmp#315 in order to resolve this. Can you (1) please check that the code there, in combination with this branch, allows test_legacy_version to pass (it does for me), and (2) if so, leave an approving review there?

Thanks!

behnam-zakeri:
> @behnam-zakeri I had to make the modifications in [iiasa/ixmp#315] ... Can you (1) please check that the code there, in combination with this branch, allows test_legacy_version to pass (it does for me), and (2) if so, leave an approving review there?
> 
Thanks @khaeru for resolving this. Yes, both tests (storage and legacy) pass after these changes.

khaeru:
> Unfortunately a test from `message_ix/tests/test_tutorials.py::test_tutorial[austria_multiple_policies-answers]` from the MacOS image is still failing

@francescolovat good catch—see further down [at line 1827](https://travis-ci.com/github/iiasa/message_ix/jobs/325874948#L1827):
```
com.zaxxer.hikari.pool.HikariPool.PoolInitializationException: com.zaxxer.hikari.pool.HikariPool$PoolInitializationException: Failed to initialize pool: Database lock acquisition failure: lockFile: org.hsqldb.persist.LockFile@5035d5ba[file =/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/pytest-of-travis/pytest-0/localdb/default.lck, exists=true, locked=false, valid=false, ] method: checkHeartbeat read: 2020-04-30 15:54:05 heartbeat - read: -9763 ms.
com.zaxxer.hikari.pool.HikariPool.PoolInitializationException: com.zaxxer.hikari.pool.HikariPool$PoolInitializationException: Failed to initialize pool: Database lock acquisition failure: lockFile: org.hsqldb.persist.LockFile@5035d5ba[file =/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/pytest-of-travis/pytest-0/localdb/default.lck, exists=true, locked=false, valid=false, ] method: checkHeartbeat read: 2020-04-30 15:54:05 heartbeat - read: -9763 ms.
```

This is a failure in the Java database layer that happens randomly once in a while; it is spurious (doesn't signal a code error). I will restart this job.

francescolovat:
Ok @khaeru thanks! 🙂

behnam-zakeri:
Thanks @francescolovat for organizing the documentation very nicely and @khaeru for imprving the model initialization workflow. I have no further additions or edits.

PR #176, opened 2019-03-01
WIP: Improve reporting (phase 2)

khaeru:
This will implement the second phase of the [reporting update](https://github.com/iiasa/message_ix/wiki/Reporting#phase-2).

**See also:**

- iiasa/ixmp#144 for corresponding changes in `ixmp`
- iiasa/message_data#41 for corresponding changes in the MESSAGE global model.
- #142 for phase 1
- #149 master issue for both phases.

**Progress**
Address requirements from wiki:
- [ ] B1. Provide equal support for any type of model that can be built in MESSAGEix, i.e. do not restrict based on specific features of any one model:
  - [ ] e.g. Clara's "hospital" model, i.e. where the 'node' set does not map to countries or geographical regions.
  - [ ] the Westeros and Austria tutorial models.
  - [ ] Single-country models, e.g. South Africa, Israel, others built using Behnam's tools.
  - [ ] Models with 'extra' features/detail added: MESSAGE-Access, MESSAGE-Transport.
  - [ ] the MESSAGE global model(s).
- [x] B2. Provide zero-config reporting, i.e. reporting that is automatically configured by introspection of the model structure. Including: calculate all outputs of commodities at all levels.
- [ ] B3. Support user-defined operations:
  - [ ] B3i. …on exogenous sub-sets of model sets.
  - [ ] B3ii. …that exclude quantities from aggregation operations.
  - [x] B3iii. …that compute the same quantity in different ways.
  - [ ] B3iv. Cumulative quantities with defined start and end periods.
- [ ] B4. Support user-defined reports.
  - [x] B4i. …defined programmatically.
  - [ ] B4ii. …loaded from a configuration file.
- [ ] B5. Provide (either automatically or per user input) operations related to model features including:
  - [ ] B5i. …sub-annual time steps.
  - [ ] B5ii. …bi-directional commodity flows.
  - [ ] B5iii. …user-defined relations. (Probably not automatic; user will need to define reporting.)

PR checklist
- [x] Tests added (ongoing)
- [ ] Tests merged with Phase 1 tests.
- [x] Documentation added — **preview [here](https://message-ix-khaeru.readthedocs.io/en/reporting2/reporting.html).**
- [x] Description in RELEASE_NOTES.md added.

Miscellaneous
- [x] Choose target release: ~1.3 or 2.0?~ 1.2

PR #196, opened 2019-05-24
WIP: Improve testing by re-using ixmp apparatus

khaeru:
Testing is substantially simplified by re-using methods and fixtures from `ixmp.testing`.

As well, this PR:
- [x] Adds methods to create Dantzig's canning/transport problem and the Westeros model to `message_ix.testing`. This will make it easier to write tests for reporting against these example models, using the code in #176.
- [x] Depends on iiasa/ixmp#137; tests will not pass on CI until that is merged.

**PR checklist**
- [x] Tests added
- [x] Documentation added
- [x] Description in RELEASE_NOTES.md added


gidden:
@khaeru do you want to restart CI since ixmp has been updated?

khaeru:
I think @danielhuppmann is the owner of the Appveyor project; I don't have permission to trigger a new build.

gidden:
Can you just `git commit --amend` and repush?

khaeru:
Okay, this is good to go.

Note that in test/tutorials.py, there is now a single test function that is parametrized with values from a list. Any additional tutorials can simply be added to the list. At some point, we should expand our contributor guidelines to include this step, as well as other actions to be taken by people developing new features/additions to `message_ix.tools`.

gidden:
Sounds good. Would you mind making an issue for that so we don't forget in the future?

RELEASE_NOTES.md
```diff
@@ -119,4 +120,4 @@ jdbc.pwd = ixmp
 - [#65](https://github.com/iiasa/message_ix/pull/65): Bugfix for downloading tutorials. Now downloads current installed version by default.
 - [#60](https://github.com/iiasa/message_ix/pull/60): Add basic ability to write and read model input to/from Excel
 - [#59](https://github.com/iiasa/message_ix/pull/59): Added MacOSX CI support
-- [#187](https://github.com/iiasa/message_ix/pull/187): Test for cumulative bound on emissions
\ No newline at end of file
+- [#187](https://github.com/iiasa/message_ix/pull/187): Test for cumulative bound on emissions
```
gidden: I know it's not you here, but can we move this back up?

message_ix/__init__.py
```diff
@@ -3,8 +3,11 @@
 
 import ixmp.model_settings as model_settings
 
-from message_ix import default_paths
-from message_ix.core import *
+from . import default_paths
+from .core import (  # noqa: F401
```
gidden: This seems like an API change (based on how users can import things) - do you think it is sufficiently important to document? Was there a reason to make this change?

tests/test_utils.py
```diff
@@ -15,3 +17,16 @@ def test_make_df():
 
 def test_make_df_raises():
     pytest.raises(ValueError, utils.make_df, 42, baz=[42, 42])
+
+
+def test_testing(test_mp):
```
gidden: Is this used somewhere? a ctrl+f doesn't find it anywhere else.

gidden: can we expand the name here, e.g. `test_testing_instance_generation`?

gidden:
Not sure why, but it seems stickler isn't going over this one?

tests/test_tutorials.py
```diff
@@ -1,176 +1,67 @@
...
+AT = 'Austrian_energy_system'
+
+# Argument values to parametrize test_tutorial
+# Each item is a 3-tuple:
+# 1. Path fragments under tutorials directory,
+# 2. Name or index of cell containing objective value (or None),
+# 3. Expected objective value.
+tutorials = [
+    (('westeros', 'westeros_baseline'),
+     'solve-objective-value', 207544.09375),
```
gidden: for future expansion, should the expected value actually be iterable here? I can imagine wanting to check multiple cells at some point. what do you think?

gidden:
Looks great @khaeru, a few small suggestions inline

tests/test_tutorials.py
khaeru: Sure, easily done.

gidden:
Ok looks great! I'll merge after CI passes.

message_ix/__init__.py
khaeru: If I do something like this:
```py
>>> import message_ix
>>> dir(message_ix)
>>> len(dir(message_ix))
```

…then there are 39 items before the change, and 28 afterwards. The excluded items are: `collections, copy, isscalar, itertools, ixmp, logger, np, pd, pd_read, pd_write, warnings`.

Some of these are reimports. We should encourage the user to do `import numpy as np; np.ones(...)` and never `import message_ix; message_ix.np.ones(...)` and so on.

The only two defined by message_ix itself are `message_ix.utils.isscalar` and `message_ix.utils.logger`. Do we mean to promise that those are available as `message_ix.isscalar` and `message_ix.logger`? If so, I can add tests to that effect, and they can be imported directly from `message_ix.utils`. IMHO it's clearer to do this explicitly than implicitly/indirectly with `from message_ix.core import *`.

gidden: I agree in general - thus would not consider this blocking for the PR. AFAIK, this is common practice in python libraries (i.e., importing submodules in their intirety in `__init__.py`). This is abused in the usage of some libraries though (looking at `from pyomo.environ import *`), but within the namespace I think its ok.

In any case, no problem with the change =)

gidden:
Thanks @khaeru !!

PR #203, opened 2019-06-03
Fix wrong electricity consumption in tutorial

lumbric:
The old link does not work any longer (404 error), but 17000 GWh
electricity consumption per year is probably a typo and should be 71000
GWh (in 2016 it rose to 72000 GWh).

(I guess this fix is too small to be mentioned in the release notes.)


gidden:
Hah, great find. Thanks @lumbric !

PR #206, opened 2019-06-13
Improve reporting (phase 2)

khaeru:
This will partially implement the second phase of the [reporting update](https://github.com/iiasa/message_ix/wiki/Reporting#phase-2).

The goal is to have a basic subset of requirements met in message_ix 1.2 / ixmp 0.2 in an experimental state, and then proceed with the remainder for the next release.

Supersedes #176, which was from a branch in a fork; all can commit to this branch.

**See also:**

- iiasa/ixmp#150 for corresponding changes in `ixmp`.
- iiasa/message_data#41 for corresponding changes in the MESSAGE global model.
- #149 master issue.

**Progress**
Address requirements from wiki:
- [x] B1. Provide equal support for any type of model that can be built in MESSAGEix, i.e. do not restrict based on specific features of any one model:
  - ~e.g. Clara's "hospital" model, i.e. where the 'node' set does not map to countries or geographical regions.~ deferred; we lack a concrete `message_ix.testing` function that sets up this model.
  - [x] the Westeros and ~Austria~ tutorial models.
  - ~Single-country models, e.g. South Africa, Israel, others built using Behnam's tools.~ deferred, ditto.
  - ~Models with 'extra' features/detail added: MESSAGE-Access, MESSAGE-Transport.~ deferred, ditto.
  - [x] the MESSAGE global model(s).
    - Basic reporting runs in a cleaned-up version of MESSAGEix-GLOBIOM; see iiasa/message_data#41.
- [x] B2. Provide zero-config reporting, i.e. reporting that is automatically configured by introspection of the model structure. Including: calculate all outputs of commodities at all levels.
- [x] B3. Support user-defined operations:
  - ~B3i. …on exogenous sub-sets of model sets.~ deferred, requires a specific example, perhaps drawn from iiasa/message_data.
  - ~B3ii. …that exclude quantities from aggregation operations.~ deferred, ditto.
  - [x] B3iii. …that compute the same quantity in different ways.
  - ~B3iv. Cumulative quantities with defined start and end periods.~ deferred, ditto.
- [x] B4. Support user-defined reports.
  - [x] B4i. …defined programmatically.
  - ~B4ii. …loaded from a configuration file.~ deferred, ditto.
- ~B5. Provide (either automatically or per user input) operations related to model features including:~ deferred, ditto.
  - ~B5i. …sub-annual time steps.~
  - ~B5ii. …bi-directional commodity flows.~
  - ~B5iii. …user-defined relations. (Probably not automatic; user will need to define reporting.)~

[...]

PR #211, opened 2019-06-23
initial draft of first standard message reporting

gidden:
Hi @khaeru and @danielhuppmann. Here is my crack at the second bullet point on #206. I think it gets us most of the way there structurally, but please feel free to further edit the internals as desired. There are a number of `TODO`s I left open, as I didn't have internet access on the plane to Korea and now will be busy with the meeting. You're more than welcome to address them (or not) as needed to get #206 in for the release.

There are a few key points I ran across:

1. I didn't have a chance to generalize the collapse functions using `eval`. I think this can go in as-is with an issue stating it should be cleaned up later if you prefer. It is a bit ugly at the moment..
2. When making a collection of tasks like `rep.add('super-task', ['set', 'of', 'subtasks'])`, I think it would be useful to either explicitly support or have helper functions to chain results back into a single returned list/container. At the moment it returns a list of return types, which gets cumbersome when they themselves are lists, etc. I tried playing around with `itertools.chain()` and `np.ravel()` which are my usual go-tos, but neither worked off the shelf for me... again, no internet.
3. **most important** there is currently a test failure in `test_reporter_from_westeros`. I explain in comments what the observed behavior is, which is related to unit registration. This definitely needs to be fixed before #206 can go into the release.


gidden:
Also to note, I made the PR so you can cleanly see the diff. I can't promise to have time to make further changes on my end, given the time difference and meeting. But I'll do my best to respond to feedback as needed - you may have to tarry on without me =)

khaeru:
@gidden thanks for handling the bulk of this while in the :airplane:! Will carry on from here.

gidden:
And sorry for the stickler, I really need to go back to my `.emacsrc` on the new laptop..

khaeru:
> **most important** there is currently a test failure in `test_reporter_from_westeros`. I explain in comments what the observed behavior is, which is related to unit registration. This definitely needs to be fixed before #206 can go into the release.

Good catch, thanks. I quickly did iiasa/ixmp#155 to resolve this issue. Restarting tests…

khaeru:
> 1. I didn't have a chance to generalize the collapse functions using `eval`. I think this can go in as-is with an issue stating it should be cleaned up later if you prefer. It is a bit ugly at the moment..

This is now done.

khaeru:
> 2. When making a collection of tasks like `rep.add('super-task', ['set', 'of', 'subtasks'])`, I think it would be useful to either explicitly support or have helper functions to chain results back into a single returned list/container. At the moment it returns a list of return types, which gets cumbersome when they themselves are lists, etc. I tried playing around with `itertools.chain()` and `np.ravel()` which are my usual go-tos, but neither worked off the shelf for me... again, no internet.

Since the test expresses that you'd eventually want to `concat()` these IamDataFrames, I just added a one-line computation that does that and made each `message:<foo>` report and `message:default` return a single object.

Maybe that will be problematic (too big?) when it comes to the global model, but seems tidy for now.

khaeru:
Stickler here is complaining about things that no longer exist, so I'm just going to merge this.

PR #220, opened 2019-07-19
Adjust standard set of reported products

gidden:
Adds more reporting capability per (private repo): https://github.com/iiasa/message_data/wiki/Reporting

**PR checklist:**

- [x] Tests added.
- [x] Documentation added. (N/A)
- [x] Release notes updated (N/A)

### DIFF
--- a/message_ix/reporting/__init__.py
+++ b/message_ix/reporting/__init__.py
@@ -72,7 +72,7 @@ PRODUCTS = (
         ('var_cost', 'ACT')),
     ('land_out',
         ('land_output', 'LAND')),
-    ('land_use',  # TODO: better name!
+    ('land_use_qty',  # TODO: better name!
         ('land_use', 'LAND')),
     ('land_emi',
         ('land_emission', 'LAND')),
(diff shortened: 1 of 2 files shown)


gidden:
@khaeru should now pass. we should probably still check in the `ixmp.Reporter` or maybe `message_ix.Reporter` that the `scenario.has_solution()` before doing anything. Otherwise I guess this is good to go and we can continue progress on the wiki?

khaeru:
@gidden what behaviour should depend on the presence of a solution?

gidden:
@khaeru, there is an inherent implication that any "reporting" is done on a model *with* a solution. To keep things simple, I will add this here briefly and we can decide later if we want to push it up into `ixmp`.

gidden:
OK @khaeru this should be good to go now. @OFR-IIASA time to move on to our next topic! All, I have updated the Wiki. Perhaps we can sit down briefly on Friday to discuss next steps and how to get others involved.

khaeru:
Thanks @gidden :+1:

PR #223, opened 2019-07-26
Connect MACRO explicitly to MESSAGEix Scenarios

gidden:
This PR brings explicit `MACRO` support into the `message_ix` library. Given a solved `MESSAGE`-stand alone scenario and additional input data ([example xlsx file](https://github.com/iiasa/message_ix/files/3435741/westeros_macro_input.xlsx)), `MACRO` is parameterized and calibrated. There is an additional option to check that calibration against the number of iterations of a `MESSAGE-MACRO` scenario, checking that `n_iter == 1 `.

cc @volker-krey @danielhuppmann @ClaraLuisa (and any others? feel free to suggest)

Depends on iiasa/ixmp#177

## PR checklist

- [x] explicit GAMS code review - I had to touch the GAMS code in a number of places and would like an explicit review there
  - Aside from #319, the only change seems to be the replacement of `total_cost` and replacement with `COST_NODAL_NET` in one location. If problems arise, hopefully this helps isolate them.—@khaeru
- [x] figure out how to exclude certain demand sectors. E.g., in the global model, we currently do not include `nc_biomass` in MACRO. Should this be done implicitly by input data exclusion or explicitly via the API?
- [x] test the code on an existing global model baseline
  - this will require migrating existing R data files to the proposed new data format (see above example)
  - I would very much appreciate assistance here, as I need to start prioritizing my time elsewhere
  - If we can get this working, we can add tests explicitly to our nightly CI that runs over global model instances
- [x] address non-convergence issues
  - in the test example, `MACRO` standalone does not converge before `max_iterations` (100) is reached
  - perhaps as a symptom of this non-convergence, tests are failing when checking that `MESSAGE-MACRO` iterations are `== 1` after calibration
- Checks added by @francescolovat according to @khaeru's comment:
  - [x] Rebase/merge master to remove merge conflicts. (done by @francescolovat @khaeru)
  - [x] Consider how to combine with #249. (done by @khaeru)
  - [x] Address comments from @behnam-zakeri. (No action; see https://github.com/iiasa/message_ix/pull/223/files#r392681157—@khaeru)
- [x] Tests added.
- ~Documentation added. And docstrings for several functions, making it understandable for e.g. the new research assistants.~ See below.
- [x] Release notes updated.

## Follow-up

The following items were **not** completed as part of this PR, and must be addressed before this code can be released with message_ix 3.0:
- ~docs added, with a tutorial explaining how to add MACRO to an example case (the already-existing test should be used here) *and* add to Sphinx docs~ → see #315, #316.
- ~understand unit issues. There are a variety of places in the previous R code as well as existing GAMS code where unit conversions are hardcoded (usually going from $B to $T via either `/ 1000` or `/ 1e3`).~ → see #317.
- ~decide what to do if MACRO-related structure already exists~ → see #320.
- ~add macro calibration to the nightly tests~ → see #318.
- ~Wrap up the discussion on interest rates and potentially open a new issue.~ → see #319.


gidden:
Added lots of potential reviewers. I think since this is so core to our main modeling infrastructure, lots of eyes should go on it, if possible.

gidden:
rebased on current master. still expect 1 test failure from convergence check. @volker-krey is investigating.

gidden:
@volker-krey we now check for 0 prices and I added a list item above for supporting the exclusion of certain sectors.

behnam2015:
> * [ ]  test the code on an existing global model baseline
>   
>   * this will require migrating existing R data files to the proposed new data format (see above example)
>   * I would very much appreciate assistance here, as I need to start prioritizing my time elsewhere
>   * If we can get this working, we can add tests explicitly to our nightly CI that runs over global model instances

Thanks @gidden for this excellent piece of work. I can help with this item. The data of global models (R14 and R11) has been removed previously from the R files and migrated to Excel (please see [this PR in message_data](https://github.com/iiasa/message_data/pull/35)). The data can be saved in the format you proposed too.

gidden:
Hey @behnam2015, this would be great! As a first step, I think we could simply check that calibrated values using this branch match those in the existing calibrated MESSAGE-MACRO scenario. Would you mind helping with that?

I think the steps would be:

1) create an excel file input dataset for the R11 model (based on what you write above, I think that should be pretty straightforward)
2) run the "new" calibration
3) check calibrated values

The code to do so would look something like this:

```
mp = ixmp.Platform()
base = message_ix.Scenario(mp, 'r11 model', 'uncalibrated baseline') # obviously changing model/scenario names
calibrated = message_ix.Scenario(mp, 'r11 model', 'calibrated baseline')

scen= base.clone(scenario='checking calibration')
scen.add_macro('path_to_excel_file') # we might need to start with also `check_convergence=False`

for par in ['aeei', 'grow']:
    obs = scen.par(par)
    exp = calibrated.par(par)
    pd.testing.assert_frame_equal(obs, exp)
```

@OFR-IIASA could you please let us know the correct model/scenario names to test here (uncalibrated/calibrated)?

behnam2015:
Thanks @gidden for elaborating and writing the code. I'll test this and share the results with you.

OFR-IIASA:
@gidden and @behnam2015 I haven't looked at the code, but I assume we only need to use a single scenario.  For this use the CD_Links_SSP2_v2 // baseline scenario.  This already has macro calibrated values, but I assume there is some cleanup function, i.e. some function to remove any existing macro related parameters and sets.  Otherwise, I would suggest to either add this functionality at some point (make new issue) and for the time being just manually remove any macro related parameters.

message_ix/macro.py
```diff
@@ -1,61 +1,464 @@
...
-        s.init_set(*args)
-    for args in MACRO_INIT['pars']:
-        s.init_par(*args)
-    # for args in MACRO_INIT['vars']:
-    #     print(args)
-    #     s.init_var(*args)
-    # for args in MACRO_INIT['equs']:
-    #     s.init_equ(*args)
+    for key, values in MACRO_INIT['sets'].items():
+        s.init_set(key, values)
```
behnam2015: @gidden, if a scenario already has MACRO-related sets, parameters, variables and equations; there will be an error when initializing these items here. 
1) To resolve this issue, we can add a check here for each item, e.g. for sets:
`if not s.has_set(key): s.init_set(key, values)`
2) On the other hand, if a scenario has a MACRO-related data, we may want to remove that data from the scenario before adding new data (in case the existing data is not correct or relevant anymore). So, we may want to remove these items in the first place and initialize them again.

message_ix/macro.py
```diff
@@ -1,61 +1,464 @@
...
+]
+
+
+def _validate_data(name, df, nodes, sectors, years):
+    def validate(kind, values, df):
+        if kind not in df:
+            return
+
+        diff = set(values) - set(df[kind])
+        if diff:
```
behnam2015: @gidden, this check needs to be revised as there may be some nodes or demand sectors in a scenario that are not included in MESSAGE-MACRO linkage. For example, in the global model we have a set 'cat_node', in which we define the nodes included in MACRO with the `node_type` of "economy". One solution as discussed can be to explicitly include/exclude the relevant nodes and sectors for macro-economic analysis.

volker-krey:
The Westeros calibration ends with very small aeei_correction and grow_correction parameters in the order of 1e-7, but at the same time the gdp_scale and demand_scale parameters that measure the change in GDP and demand that MACRO provides to the calibration values are 0.482 and 0.837, respectively (for all years). The initial grow parameter provided to the calibration process via the GDX consists of 3 entries (with four periods incl. base period) while the examples using the R calibration process include 4 entries, one of which is appended in the data file. As a result the growth_correciton parameter calculated in the calibration process only consists of 2 entries which means that the parameter of one period is not calibrated. My suspicion therefore is that this results in an offset as GAMS seems to be able to work with the mismatch in dimensionality somehow. As a quick test of this hypothesis, I would propose to duplicate the last entry of the grow parameter so that four values are provided.

gidden:
Hey @volker-krey thanks for the analysis! I'm digging into this now. One clarifying question to wrap my head around it. The model periods (where there are decision variables) are 700, 710, and 720 here. Should the `grow` parameter (and `gdp_calibrate`?) have additional values in 690 or 730?

If I look at the R file examples, I see that grow and gdp_calibrate both have values from 2010-2110, so I guess it's in the "base year" (i.e., 2010 in the global case, 690 in the westeros case)?

volker-krey:
Yes, it is 690 that needs to be added. In principle the GAMS code in macro_calibration.gms should reveal what is needed as this is where things are going wrong.

gidden:
Ok, great, I can look into it. A first test showed that the calibration is still taking 4 iterations, so I suspect there are a few parameters that need values here. Will let you know when I dig further.

volker-krey:
Looking at the gdp_scale and demand_scale parameters at the end of the calibration is an additional good test. If these are not ~1, there is going to be a demand feedback in an iterated run.

message_ix/macro.py
gidden: Hey @behnam2015, good points here. My initial reaction is that if a scenario already has MACRO-related data, then it *should* be an error to try to "re-add" it. Perhaps we can add an additional function to `remove()` MACRO related parameters as well. 

This is a bit tricky, because I observed that even in the westeros model, some things already existed (e.g., the set `mapping_macro_sector`), which I believe could be added automatically on the java side of things. Can @danielhuppmann chime in with his thoughts here?

gidden: I did a quick test to remove things and then re-add, but got the following error:

```
self = <message_ix.core.Scenario object at 0x7efd66bfd0f0>, name = 'historical_gdp', idx_sets = ['node', 'year'], idx_names = None

    def init_par(self, name, idx_sets, idx_names=None):
        """Initialize a new parameter.
    
        Parameters
        ----------
        name : str
            Name of the parameter.
        idx_sets : list of str
            Names of sets that index this parameter.
        idx_names : list of str, optional
            Names of the dimensions indexed by `idx_sets`.
        """
>       self._jobj.initializePar(name, *make_dims(idx_sets, idx_names))
E       jpype._jclass.at.ac.iiasa.ixmp.exceptions.IxException: Parameter 'historical_gdp' can only be initialized with MESSAGE standard dimensions!
```

message_ix/macro.py
gidden: Hey @behnam2015, thanks again for the insight here. The first thing to note is that (in theory) a "clean" message model (without any prior knowledge of MACRO), should *not* have a `type_node` economy. In the test example in this PR, we add this model structure in `add_model_data()`.

As discussed in person, we agreed to add a sheet/dataframe to the input data called `config` which will have for the moment two possible columns: `ignore_node` and `ignore_sector`. Upon reading in the data, if there are entries here, they will be removed from the expected sets.

gidden: @behnam2015, this feature is now available in the PR added as commit 7994810.

message_ix/macro.py
gidden: @behnam2015, I have made a new branch on my remote called `macro-behnam-test`, which includes a new try/catch for `init_set`. Let's see what @danielhuppmann says before we do anything on this PR branch. But in the meantime, you can use `macro-behnam-test` to start testing the global model (it includes the `ignore_node` and `ignore_sector` feature as well).

behnam2015: Thanks @gidden. By removing data, I meant only removing the content of a parameter and not the parameter itself. Something like:
key = 'historical_gdp'
```
if s.has_par(key):
    s.remove_par(key, s.par(key))
else:
    s.init_par(key, idx_sets)
```

message_ix/macro.py
behnam2015: Thanks @gidden, sounds like a very good and straightforward remedy.

message_ix/core.py
```diff
@@ -366,6 +366,16 @@ def solve(self, model='MESSAGE', solve_options={}, **kwargs):
         os.remove(fname)
         return ret
 
+    def add_macro(self, data, scenario=None, check_convergence=True):
+        scenario = scenario or '_'.join([self.scenario, 'macro'])
```
khaeru: Docstring, please!

message_ix/model/MESSAGE-MACRO_run.gms
```diff
@@ -91,6 +91,8 @@ Scalar
     max_adjustment     maximum adjustment in current iteration
     convergence_status status of convergence (1 if successful) / 0 /
     scaling            scaling factor to adjust step size when iteration oscillates / 1 /
+    max_it /%MAX_ITERATION%/
+    ctr /0/
```
khaeru: These should have description text, like the preceding.

gidden:
@behnam2015 I have now added support for reading in data with multiple regions and sectors in a92369617. It has also been added to the `macro-behnam-test` branch, so you should be good to go with further testing of the global model.

message_ix/macro.py
```diff
@@ -1,61 +1,471 @@
...
+                # TODO: this seems required because for some reason DEMAND (and
+                # perhaps others) seem to already be listed in the java code,
+                # but still needs to be initialized in the python code.
+                # However, you cannot init it with dimensions, only with the
+                # variable name.
+                s.init_var(key, values)
+            except:  # noqa: ignore=E722
+                s.init_var(key)
+    for key, values in MACRO_INIT['equs'].items():
+        s.init_equ(key, values)
```
behnam2015: @gidden, I needed to change this to work, it had some problems with existing equations in the scenario.
```
        try:
            s.init_equ(key, values)
        except:
            pass
```

message_ix/macro.py
```diff
@@ -1,61 +1,471 @@
...
+    # add parameters
+    for name, values in MACRO_INIT['pars'].items():
+        try:
+            key = values.get('data_key', name)
+            data = c.data[key].reset_index()
+            data['unit'] = values['unit']
+            # some data may have information prior to the MACRO initialization
+            # year which we need to remove in order to add it to the scenario
+            if 'year' in data:
+                data = data[data['year'] >= c.init_year]
```
behnam2015: @gidden, I needed to add a check for `sector` here to prevent an error when using `add_par` for _demand_MESSAGE_:
```
            if 'sector' in data:
                data = data[data['sector'].isin(clone.set('sector'))]
```

message_ix/macro.py
```diff
@@ -1,61 +1,471 @@
...
+        except Exception as e:
+            msg = 'Error in adding parameter {}\n'.format(name)
+            raise type(e)(msg + str(e))
+
+
+def calibrate(s, check_convergence=True):
+    # solve MACRO standalone
+    var_list = ['N_ITER', 'MAX_ITER', 'aeei_calibrate', 'grow_calibrate']
+    gams_args = ['LogOption=2']  # pass everything to log file
+    s.solve(model='MACRO', var_list=var_list, gams_args=gams_args)
```
behnam2015: @Gidden, here I receive an error saying that _gams_args_ is not a known method for `Scenario.solve()`. Can this be passed as _solve_options_ instead?

gidden: You may need to get the current version of the `ixmp` master branch, I added that capability in a pr: https://github.com/iiasa/ixmp/pull/177

message_ix/macro.py
gidden: Hmm, yes, this is also related to https://github.com/iiasa/message_ix/pull/223#discussion_r312356449. I'll add it as a top-level item.

message_ix/macro.py
behnam2015: Thanks @gidden! I'll update my ixmp.

message_ix/macro.py
gidden: Added a check when reading in model data (as discussed in person)

behnam2015:
Thanks a lot @gidden, for this nice job. I tested this for R11 and R14 and it works well.

behnam2015:
Based on my observation, choosing "barrier" as the LP method in solving the calibrated scenario is much faster compared to the default method; a few minutes vs. more than an hour for a global model. If you have the same experience, I suggest changing the default method of _solve_ to "barrier" in `add_macro`. 
```
scen.solve(model='MESSAGE-MACRO', gams_args=['LogOption=2'], solve_options={'lpmethod': '4'})
```

message_ix/model/MACRO/macro_data_load.gms
```diff
@@ -231,7 +231,7 @@ LOOP(year_all $( ORD(year_all) > sum(year_all2$( macro_initial_period(year_all2)
 * new labor supply
    newlab(node_macro, year_all) = SUM(year_all2$( seq_period(year_all2,year_all) ), (labor(node_macro, year_all) - labor(node_macro, year_all2)*(1 - depr(node_macro))**duration_period(year_all))$((labor(node_macro, year_all) - labor(node_macro, year_all2)*(1 - depr(node_macro))**duration_period(year_all)) > 0)) + epsilon ;
 * calculation of utility discount factor based on discount rate (drate)
-   udf(node_macro, year_all)    = SUM(year_all2$( seq_period(year_all2,year_all) ), udf(node_macro, year_all2) * (1 - (DRATE(node_macro) - grow(node_macro, year_all)))**duration_period(year_all)) ;
+   udf(node_macro, year_all)    = SUM(year_all2$( seq_period(year_all2,year_all) ), udf(node_macro, year_all2) * (1 - (interestrate(year_all) - grow(node_macro, year_all)))**duration_period(year_all)) ;
```
gidden: nice @volker-krey! I noticed as well that there are a few other places `drate` is used:

```
$ ack -i drate message_ix/model/MACRO/
message_ix/model/MACRO/macro_data_load.gms
63:* VK, 10 April 2008: DRATE (social discount rate, i.e. net rate of return on capital from MESSAGE) introduced as a new parameter as in MERGE 5
64:         drate(node)                 Social discount rate
134:$LOAD kpvs,kgdp,esub,depr,drate,lotol
233:* calculation of utility discount factor based on discount rate (drate)
262:* that are greater than drate = depr. To avoid negative values the absolute value is
266:finite_time_corr(node_macro, year) = abs(DRATE(node_macro) - grow(node_macro, year)) ;

message_ix/model/MACRO/macro_core.gms
142:* The utility discount rate for period :math:`y` is set to :math:`drate_{n} - grow_{n,y}`, where :math:`drate_{n}` is the discount rate used in MESSAGE, typically set to 5%,

message_ix/model/MACRO/macro_calibration.gms
86:* calculation of utility discount factor based on discount rate (drate)
87:   udf(node_macro, year_all)    = SUM(year_all2$( seq_period(year_all2,year_all) ), udf(node_macro, year_all2) * (1 - (DRATE(node_macro) - grow(node_macro, year_all)))**duration_period(year_all)) ;
91:finite_time_corr(node_macro, year) = abs(DRATE(node_macro) - grow(node_macro, year)) ;
```

Specifically lines 266 and 91 above. Should this also be changed?

message_ix/macro.py
```diff
@@ -1,61 +1,494 @@
...
+        model_cost.rename(columns={'lvl': 'value'}, inplace=True)
+        model_cost = model_cost[idx + ['value']]
+        # TODO: in the R code, this value is divided by 1000
+        # do we need to do that here?!!?
+        model_cost['value'] /= 1e3
+        # get data provided in init year from data
+        cost_ref = self.data['cost_ref'].reset_index()
+        cost_ref['year'] = self.init_year
+        # combine into one value
+        total_cost = pd.concat([cost_ref, model_cost]).set_index(idx)['value']
```
behnam2015: A minor suggestion to turn off the warnings:
```
total_cost = pd.concat([cost_ref, model_cost], sort=True).set_index(idx)['value']
```

message_ix/macro.py
```diff
@@ -1,61 +1,494 @@
...
+            raise RuntimeError(msg)
+        model_price.rename(columns={'lvl': 'value', 'commodity': 'sector'},
+                           inplace=True)
+        model_price = model_price[idx + ['value']]
+
+        # get data provided in init year from data
+        price_ref = self.data['price_ref'].reset_index()
+        price_ref['year'] = self.init_year
+        # combine into one value
+        price = pd.concat([price_ref, model_price]).set_index(idx)['value']
```
behnam2015: A minor suggestion to turn off the warnings:
```
price = pd.concat([price_ref, model_price], sort=True).set_index(idx)['value']
```

message_ix/macro.py
```diff
@@ -1,61 +1,494 @@
...
+            self.s.var('DEMAND', filters={'level': 'useful'})
+        )
+        model_demand.rename(columns={'lvl': 'value', 'commodity': 'sector'},
+                            inplace=True)
+        model_demand = model_demand[idx + ['value']]
+        # get data provided in init year from data
+        demand_ref = self.data['demand_ref'].reset_index()
+        demand_ref['year'] = self.init_year
+        # combine into one value
+        demand = pd.concat([demand_ref, model_demand]).set_index(idx)['value']
```
behnam2015: A minor suggestion to turn off the deprecation warnings:
```
 demand = pd.concat([demand_ref, model_demand], sort=True).set_index(idx)['value']

**Comment by volker-krey** [comment]
@gidden @behnam2015 as discussed in today's MESSAGE meeting, I switched drate to interstrate in the finite time horizon factor in macro_data_load.gms where I had missed it previsously. In addition, macro_calibration.gms still used drate which I also switched to interestrate which is part of a PR into gidden:explore-macro (https://github.com/gidden/message_ix/pull/5) .

**Comment by volker-krey** [comment]
We should have a discussion whether using interestrate is really the parameter to use for MACRO discounting in the long-term or, if we instead turn drate into a time-dependent parameter. The experiments with the national model will tell us more, so would suggest to wait with merging this PR until we have those insight.

**Comment by vruijven** [comment]
> 
> 
> @gidden @behnam2015 as discussed in today's MESSAGE meeting, I switched drate to interstrate in the finite time horizon factor in macro_data_load.gms where I had missed it previsously. In addition, macro_calibration.gms still used drate which I also switched to interestrate which is part of a PR into gidden:explore-macro ([gidden#5](https://github.com/gidden/message_ix/pull/5)) .

Could it be that the other changes of DRATE into interestrate are in another branch? In gidden:explore-macro there is still a DRATE in the udf calculation in macro_calibration.gms and what about the use of DRATE in finite_time_corr?

**Comment by OFR-IIASA** [comment]
@volker-krey @gidden @behnam-zakeri @vruijven @khaeru Hi all, I was wondering whether this PR can be merged into the master? it seems that the branch requires a re-base as there are some conflicts and there are some outstanding points. Can these be put off for later? Otherwise, who can finish this PR?

**Comment by khaeru** [comment]
Hi all—per a request from @gidden on his visit to IIASA last week, here is my gloss of the tasks needed to complete this PR:

- Rebase/merge master to remove merge conflicts.
- Address comments from @behnam-zakeri.
- Wrap up the above discussion on interest rates. Contra @volker-krey I think, in the interest of having a not-too-large PR and continuing to move forward, it is fine to simply:
  - *Clearly describe/document* the current representation, and then…
  - *Open a new issue* to state or discuss any planned change/improvement to the representation.
- Consider how to combine with #249; on this point I can guide/advise.
- Update and complete the checklist in the PR description, above.
  - Documentation is essential and I will oppose merging this until the procedure is documented to the point that it can be understood by one of our new research assistants.

We can discuss who can do this work at this Wednesday's meeting.

**Comment by francescolovat** [comment]
Hi @gidden , 

I'm Francesco, one of the newcomers to IIASA who'll be working with @khaeru on MESSAGE-Transport.

To make you aware of the recent changes: 

Under Paul's supervision I've just gone through the conflicts and solved them. Still, the CI tests are failing because Paul has to add a MACRO class (as he did in #249 in the `message_ix/__init__.py` file) to account for the possibility of running MACRO alone. He'll add it soon.

I'll also edit the checklist soon. 

I hope everything is easy readable 😃

**Comment by khaeru** [comment]
@francescolovat in 2f5a28219f4409f763f502bfe050e146f149b3ca I added a `MACRO` class to go with the `MESSAGE` and `MESSAGE_MACRO` classes added in #249. This requires a fix in ixmp: iiasa/ixmp#200—please review and approve that, when you have a moment. Once I merge that, these tests should pass again.

**Comment by volker-krey** [comment]
@khaeru we can clearly remove the change from DRATE to interestrate from this PR (related to your suggestion in https://github.com/iiasa/message_ix/pull/223#issuecomment-549285690) as this was a specific need introduced for the Israel model. As I am not sure that we want to include this change in other instances, I will revert this PR to the previous parametrization. However, this won't help with making the PR simpler as this affect maybe five lines of code.

File: tests/test_macro.py [hunk 12]
```diff
@@ -1,12 +1,238 @@
...
+                self.data[name] = df
+
+    def has_solution(self):
+        return True
+
+    def var(self, name, **kwargs):
+        df = self.data['aeei']
+        # add extra commodity to be removed
+        extra_commod = df[df.sector == 'i_therm']
+        extra_commod['sector'] = self.data['config']['ignore_sectors'][0]
```
ClaraLuisa: A little thing I noticed that doesn't change much but helps with the million Pandas Slice warnings.

> C:\Program Files\JetBrains\PyCharm 2019.1.1\helpers\pydev\pydevconsole.py:1: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

From my experience using assign insted of trying to overwrite a slice works very well. e.g. changing statements such as: 
`extra_commod['sector'] = self.data['config']['ignore_sectors'][0]` to 
`extra_commod = extra_commod.assign(sector=self.data['config']['ignore_sectors'][0])`

message_ix/macro.py
```diff
@@ -1,61 +1,494 @@
...
+
+def _validate_data(name, df, nodes, sectors, years):
+    def validate(kind, values, df):
+        if kind not in df:
+            return
+
+        diff = set(values) - set(df[kind])
+        if diff:
+            raise ValueError(
+                'Not all {}s included in {} data: {}'.format(kind, name, diff)
```
ClaraLuisa: Might there not be cases where not all nodes are calibrated (e.g. in the global model the "virtual trade node" _GLB_  is not calibrated so far. Is this really an issue?

message_ix/macro.py
```diff
@@ -1,61 +1,494 @@
...
-        ['COST_ACCOUNTING_NODAL', ['node', 'year', ], ]
-    ],
+    'sets': {
+        'sector': None,
+        'mapping_macro_sector': ['sector', 'commodity', 'level', ],
+    },
+    'pars': {
+        'demand_MESSAGE': {
+            'idx': ['node', 'sector', 'year', ],
+            'unit': 'GWa',
```
volker-krey: As for any MESSAGEix model, units in any MACRO model are not defined ex-ante for each model instance, but can in principle be freely chosen. While this set of units might be the right ones in the most frequently used models, I don't think they should be hard coded into this script, but rather be read in together with the numerical data.

message_ix/macro.py
```diff
@@ -1,61 +1,494 @@
...
+            'data_key': 'price',
+        },
+        'cost_MESSAGE': {
+            'idx': ['node', 'year', ],
+            'unit': 'G$',
+            'data_key': 'total_cost',
+        },
+        'gdp_calibrate': {
+            'idx': ['node', 'year', ],
+            'unit': 'T$',
```
volker-krey: In addition to the above comment about units in general, specifically for the GDP parameter on the MACRO side, it worthwhile mentioning that there is a (hard-coded) unit conversion factor of 1000 between systems costs in MESSAGE and GDP in MACRO. For example, if MESSAGE system costs are in units of billion $ (G$), then MACRO GDP needs to be in trillion $ (T$).

khaeru:
Postponing this to after 2.0 because of the amount of work remaining to be done.

ClaraLuisa:
Hey @gidden and @volker-krey I am trying to create a MESSAGE-MACRO tutorial for the Westeros example and am having some issues with the calibration process. While I am pretty sure, that I am not yet using it quite correctly, I think that there might also be some little unit / number issues with the PR code as is.

I, for example, do not understand why COST_NODAL_NET are exactly 1000 times larger in the baseline_macro scenario compared to the baseline. Is this supposed to be this way?

gidden:
Hey @ClaraLuisa - indeed this is related to the unresolved issues in the original PR statement, namely:

> - [ ] understand unit issues. There are a variety of places in the previous R code as well as existing GAMS code where unit conversions are hardcoded (usually going from $B to $T via either `/ 1000` or `/ 1e3`).

ClaraLuisa:
> Hey @ClaraLuisa - indeed this is related to the unresolved issues in the original PR statement, namely:
> 
> > * [ ]  understand unit issues. There are a variety of places in the previous R code as well as existing GAMS code where unit conversions are hardcoded (usually going from $B to $T via either `/ 1000` or `/ 1e3`).

Hey @gidden, thanks for your prompt response! Do I understand you correctly, that what you're saying that  the reason why there is no demand response my dummy Westeros_Macro model has nothing to do with those unit issues but is due to a problematic calibration?

gidden:
No I was just saying that random factors of 1000 pop up in different places
=) You should be able to model the tutorial directly after the tests I
think?
https://github.com/iiasa/message_ix/pull/223/files#diff-a8d9d7de9eade2639ca357b7fea64082
I did a westeros macro setup there

ClaraLuisa:
Oh, thanks for the clarification. However, despite setting up and running macro works super well using the workflow you created, the demand does not change between the scenarios _westeros_solved_ and _with_macro_. Is this behavior intended?

gidden:
Yes, this should just be the scenario with calibrated macro values (only `aeei` and `grow` if memory serves). The demand should be the same. The demand should change only when the macro baseline is perturbed by a policy.

ClaraLuisa:
> Yes, this should just be the scenario with calibrated macro values (only `aeei` and `grow` if memory serves). The demand should be the same. The demand should change only when the macro baseline is perturbed by a policy.

Hey @gidden, thanks for the info - I will try this in a new tutorial as soon as I have my system back up and running. Unfortunately my system crashed last week, so I'm now trying to set everything up fresh. However, now I'm running into the issue that if I create an environment and install ixmp from conda-forge (v.2.0.0) and then message-ix from source from your _explore_macro_ branch I run into the GAMS error that the ixmp version is not compatible with the message version. Any idea on which ixmp release might be best to continue here?

gidden:
we might need to rebase this branch on latest master - @khaeru?

khaeru:
> we might need to rebase this branch on latest master - @khaeru?

Yes, a rebase is needed to finish the PR.

I am not assigned here, but the last commits on the branch were by me, ~2 months ago. I am going to remove the assignees to make it clearer that this PR doesn't currently have an owner.

khaeru:
[This text](https://github.com/iiasa/message_data/pull/33/files) from iiasa/message_data#33 may be useful when adding documentation for this PR.

khaeru:
Rebased, tests pass on my machine.

khaeru:
> - [x] Consider how to combine with #249; ask @khaeru for guidance.

I've contributed one commit, 875a1d5, that makes some of these changes and also harmonizes this code with iiasa/ixmp#212. This allows to remove an entire method, `macro.init()`. See the commit description for more.

gidden:
rebased now, :crossed_fingers: tests will pass =)

khaeru:
Per discussion at the 2020-03-11 MESSAGE meeting, we will merge this PR roughly as-is, and convert the remaining 6 or 7 TODO items to individual follow-up issues, to be assigned and addressed before the next release.

message_ix/core.py
khaeru: See #315.

message_ix/model/MESSAGE-MACRO_run.gms
khaeru: See #315.

khaeru:
Opened six issues.

message_ix/macro.py
khaeru: As of pandas 1.0.0, the warning is no longer raised; new users installing from scratch will not see this, so no need to explicitly give the argument to avoid a visible warning.

message_ix/macro.py
khaeru: See https://github.com/iiasa/message_ix/pull/223/files#r392681157

PR #231, opened 2019-08-21
Updating the URL to Anaconda in Installation notes

behnam2015:
The url link to download Anaconda seems broken. This change does not require a new patch release and has no change to the code, only docs.


khaeru:
Good eye @behnam2015, thank you!

Issue #246, opened 2019-10-17
Improving Austrian Energy System tutorials - experiences from new users


jkikstra:
Overall:
- Add order: the python notebook tutorials are to be run in a specific order, but this is not specified.
- 

austria.ipynb:
- when adding spatial detail: add a sentence showing what this command would look like if you are adding multiple countries/adding a full spatial set.
- economic parameters: 
	- if the value is 0.05, the unit should be '-', rather than '%'
	- add some explanation about 'beta' == macro relation between gdp growth and energy demand (demand flexibility)?
- technologies:
	- check units: be more clear about the input data units from IEA statistics: if this is really demand per year, then why are you dividing by 8760?
- Engineering Parameters:
	- for inputs (and outputs): if the values are 1.0 and 0.3, the unit should be '-', rather than '%' ==> explanation of what these values represent (as a comment, e.g. # efficiency) wouldn't hurt 
	- already here add explanation: why are both 'bulb' and 'cfl' here?
	- counterintuitive: why is 'import' modelled as an output, rather than an input => perhaps add explanation on how import is outputting to add to the energy of the system, and can be an input to other technologies after that?
	- again, for capacity factor, change unit from '%' to '-'
- Technoeconomic Parameters:
	- should the unit of investment cost (base_inc_cost) change to 'USD'?
	- base_var_cost and base_fix_cost should be in the same units; its value should probably be muliplied by the same number as well (unless the input units are different - if this is the case, that should be clarified)
- Dynamic Behavior Parameters:
	- base_growth & base_initial: change '%' to '-'
	- what is base_initial? ==> add explanation
	- add notes/explanation that both lower and upper bound activity are first set to be the same in this simplified energy system model

_To be updated with more comments on the other tutorial ipynb files_

GamzeUnlu95:
I also encountered some of the issues above. To complement the comments above: 

- In techno-economic parameters for investment and fixed costs the unit is mentioned as USD/GWa. But for investment and fixed costs the unit is USD/GW in the calculation which is correct. For variable cost, the unit specified looks correct and consistent with the calculation (USD/GWa). 

- Some parameters and how they are used/calculated can be more clear such as: initial_activity_up, bound_new_capacity_up; bound_activity_up and bound_activity_lo in Dynamic Behavior Parameters Section.

francescolovat:
[...]

PR #248, opened 2019-10-21
Updated tutorial file austria.ipynb

francescolovat:
Collection of errors found by @GamzeUnlu95 @jkikstra and myself.

Still, some comments are missing for describing some lines of code and some parameters in order for an external reader to fully understand the code. We @jkikstra  @GamzeUnlu95 don't understand this specific lines/what methodology was used at in those:

- add some explanation about 'beta' == macro relation between gdp growth and energy demand (demand flexibility) (and its source)
-  difference between the parameters `initial_activity_up` and `activity_growth_up`, can't find their description in the documentation.
- what this line do: `capacity = (act / 8760 / cf).dropna().to_dict()`
- Moreover, the values obtained in some of the plots seem a bit little for the Austrian energy system. - The values from the plot of the new capacity they do not correspond to the increases reflected in the cumulative capacity plot attached below. _Is it because of the presence of a linear decommissioning curve proportional to the technical lifetime of each of the technologies?_

image

- The values for the electricity price in $/kWh from the graph below seem extremely low. Is that because the demands and the base_activity have been inserted in GWa instead of kWa? @OFR-IIASA 

image

Please refer to issue #246 for further details of our findings and disccussion on the tutorials.


khaeru:
Hi @francescolovat—after Monday's tutorial, please have a go at updating the tests to match your changes to the tutorials, and ask for help if needed!

volker-krey:
I had a look through the changes in the tutorial and to me they all look correct. As suggested by @khaeru the tests will need to be adjusted to match the correct solution of the updated tutorial.

volker-krey:
@francescolovat thanks for adjusting the tests and getting the unit issues in the tutorial fixed.

PR #253, opened 2019-11-02
Run Travis cron jobs as part of the main test suite

khaeru:
This PR integrates the slow-running tests for Travis nightly ('cron') CI with the overall message_ix test suite. The tools used to prepare, upload, and retrieve the test databases are moved to `message_ix.testing.nightly` and made available through the CLI.

Closes #252.

**PR checklist:**

- [ ] Tests added.
- [ ] Documentation added.
- [ ] Release notes updated.

Issue #244, opened 2019-10-10
MESSAGE-MACRO dependency on GAMS version >24.8.1 not explicit in installation instructions


volker-krey:
Running MESSAGEix with MACRO linkage includes a GAMS version dependency. The “break” statement used in MESSAGE-MACRO_run.gms to exit the iteration loop between MESSAGE and MACRO was added to GAMS version 24.8.1 (https://www.gams.com/latest/docs/RN_248.html#RN_2481). It seems the documentation of this dependency has been accidentally removed in the restructuring of the installation guidelines (at least I cannot find it) and should be re-added more clearly again.

PR #255, opened 2019-11-02
Run Travis cron jobs as part of the main test suite

### LINKED ISSUE #252: Travis cron CI is broken
The nightly/cron Travis builds were broken by #249; see e.g. [Travis job #1173.1](https://travis-ci.org/iiasa/message_ix/jobs/606056647). The culprit seems to the re-org of CI files.

khaeru:
This PR integrates the slow-running tests for Travis nightly ('cron') CI with the overall message_ix test suite. The tools used to prepare, upload, and retrieve the test databases are moved to `message_ix.testing.nightly` and made available through the CLI.

Closes #252.
Partially addresses #244 by adding docs for a new `MESSAGE_MACRO.GAMS_min_version` attribute.

Supersedes #253. The secure environment variables added by this PR are only available in builds from within the main (iiasa/message_ix) repo, [per Travis security restrictions](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions). (The other PR was from my fork.)

**PR checklist:**

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.


khaeru:
@gidden @zikolach I'd value your thoughts at this point. As of now, the long-running tests pass on Travis/Linux; but consistently run into Java heap space issues. See the most recent build logs on this PR. Two questions:

1. Is there an easy way to resolve this? I've tried giving `jvmargs=['-Xmx4G']`, but no change.
2. Is it even important to do so? These nightly tests are not really to test that the global model **can be run** in some/all environments, just that it **gives correct results** when run. So is it sufficient to run them only on Linux?

zikolach:
@khaeru referring to https://docs.travis-ci.com/user/reference/overview/ we should probably not set max heap size to 4g as travis vm (osx) has only 4g ram.
How big is the dataset to be written to gdx? We should probably have some reproducible test case and do some optimizations.

khaeru:
> travis vm (osx) has only 4g ram

I also noticed this. I'd like to test with different values, but this is a clean PR so far and I'd like to avoid larding it up with dozens of CI debugging commits. So maybe that means I should merge with macOS tests disabled and then use a new PR for that debugging.

> How big is the dataset to be written to gdx?

They are the ones defined here: https://github.com/iiasa/message_ix/blob/99e6b6e5527ca076a928a10bc40b42dce5b38199/tests/data/scenarios.yaml#L1-L23

> We should probably have some reproducible test case and do some optimizations.

Agreed, we should.

khaeru:
> 2. Is it even important to do so? These nightly tests are not really to test that the global model **can be run** in some/all environments, just that it **gives correct results** when run. So is it sufficient to run them only on Linux?

Per discussion on Slack with @gidden—yes, it is sufficient. The nightly runs are a “smoke” test, i.e. an indicator that something might be wrong and should be investigated; this indication will be given by a failing test on any one platform.

I have tried one final tactic in d14375f45c45eaf87951a2b469d66f29237e1f46 to avoid the heap space issue on macOS. If that doesn't work, we will merge this with the macOS tests disabled.

khaeru:
> I have tried one final tactic in [d14375f](https://github.com/iiasa/message_ix/commit/d14375f45c45eaf87951a2b469d66f29237e1f46) to avoid the heap space issue on macOS. If that doesn't work, we will merge this with the macOS tests disabled.

It didn't work :upside_down_face:

khaeru:
Okay, [Travis build # 1196](https://travis-ci.org/iiasa/message_ix/builds/607684641) is our confirmation that this works.

I will now disable this temporary setting that was just for this PR: https://github.com/iiasa/message_ix/blob/0b9fb46f3f2afbbe24e28fa14c25955a52cb1070/tests/test_nightly.py#L15-L25

The tests should again pass, but without the nightly tests being run.

PR #258, opened 2019-11-06
added import statement for using add_year function in code (vs. via command line)

volker-krey:
When trying to run the add_year function from inside a Jupyter notebook I had to import the add_year function from the package first. Adding the import statement explicitly to the README has the benefit of making it easier for less experienced Python users to utilize this function.


behnam-zakeri:
@volker-krey, thanks for the good suggestion.

khaeru:
Thanks @volker-krey!

Issue #254, opened 2019-11-02
Disentangle message_ix and ixmp_source


khaeru:
One goal of iiasa/ixmp#182 / #249 was to clarify what core functionality of the message_ix _modeling framework_ currently resides in the Java code of ixmp_source (i.e. behind ixmp's JDBCBackend).

As the name suggests, it is _not_ the task of a storage backend to provide core model/framework logic; this logic should instead live in this repo. (In the current situation, new, non-Java backends would each be forced to re-implement that logic, which would be duplicative and increase points of failure.)

This issue is for:
- @danielhuppmann @gidden @zikolach myself and possibly others to identify these pieces, and
- a broader set of people to divide and coordinate (in further issues/PRs) the work of porting them from ixmp_source to message_ix.

## Pieces
(@khaeru to flesh these out/keep them updated based on discussion below; **others** feel free to edit)

1. **Scenario initialization.** `message_ix.Scenario(..., version='new')` returns a Scenario _pre-populated_ with certain sets and parameters.
   - #223 sets an example here: the MACRO sets and parameters are visible in the Python code. As a further iteration @gidden @khaeru feel they could be moved to a YAML file.
2. **Mapping sets.** `message_ix.Scenario.cat_list(...)`, `.cat(...)` and `.add_cat(...)` operate by creating and manipulating related sets named `type_*` and `map_*`.
3. **GDX output.** The ixmp_source method `Scenario.toGDX()` writes Scenario data to a GDX file.
   - Depends on item 2.
4. **GDX input.** The ixmp_source method `Scenario.readSolutionFromGDX()` reads Scenario data (or a specific subset that is "solution" data) from a GDX file.

## Notes

- @zikolach pointed out that some of the Java code is *used by other builds from ixmp_source*, e.g. behind the Scenario Explorer. So, _re-implementing_ the functionality in Python does not always mean it can be _removed_ from ixmp_source. On a case-by-case basis, we will need to decide this; and if the code is in both places, then there must be clear warnings to future developers that the behaviour must be kept in sync, perhaps by marking one or the other as authoritative.
- Items 3 and 4 are needed to make ixmp's GAMSModel non-backend-specific. Currently it only works with JDBCBackend.

danielhuppmann:
I'm not sure why you think that item 3 depends on item 2?

gidden:
I guess it is item 4 that actually depends on item 2? I.e., you need to correctly populate the mapping sets in order to write a correct gdx file, right?

[...]

PR #269, opened 2019-11-24
 Convert 'year'-indexed items to 'int' columns

### LINKED ISSUE #268: Handle 'year' integer types in message_ix.Scenario
MESSAGEix expects parameter dimensions indexed by the set 'year' (including `year_act`, `year_vtg`, and others) to be handled as `int`, rather than `str` for all other dimensions.

This has been handled in ixmp: https://github.com/iiasa/ixmp/blob/c6ae2f74f8d8e1936203abdd79f1d403bb9477bc/ixmp/core.py#L930-L938

However, an ixmp.Scenario need not have any set named 'year'. 'year' could be an equation, parameter, or variable in a valid Scenario. So, that is the wrong place for such code.

Instead, message_ix.Scenario.par should override ixmp.Scenario.par, and then apply this conversion.

Filed via iiasa/message_ix#213.

khaeru:
Closes #268 and iiasa/ixmp#218.

- Adds test_core.test_year_int for the behaviour described in #268.
  - `pytest -k year_int` with only the first commit (b7a8428) should fail.
- Overrides Scenario.equ, .par, .set, and .var to provide the desired behaviour.
- Adjusts other code in `tools.add_year` and tutorials/utils/run_scenarios.py.

See also #254.

**PR checklist:**

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

### DIFF
--- a/message_ix/core.py
+++ b/message_ix/core.py
@@ -27,25 +27,40 @@ class Scenario(ixmp.Scenario):
         super().__init__(mp, model, scenario, version, scheme, annotation,
                          cache)
 
+    # Utility methods used by .equ(), .par(), .set(), and .var()
+
     @lru_cache()
     def _year_idx(self, name):
-        """Return (idx_set, idx_name) for 'year'-indexed dims of *name*."""
-        # NB Since item dimensionality is fixed, the result can be cached for
-        #    performance
-        return list(filter(
-            lambda e: e[0] == 'year',
-            zip(self.idx_sets(name), self.idx_names(name))
-        ))
+        """Return a sequence of (idx_set, idx_name) for 'year'-indexed dims.
+
+        Since item dimensionality does not change, the the return value is
+        lru_cache()'d for performance.
+        """
+        # filter() returns a 1-time generator, so convert to a fixed tuple()
+        return tuple(
+            # Keep only tuples where the idx_set is 'year'
+            filter(lambda e: e[0] == 'year',
+                   # Generate 2-tuples of (idx_set, idx_name)
+                   zip(self.idx_sets(name), self.idx_names(name))))
 
     def _year_as_int(self, name, df):
-        """Convert 'year'-indexed columns of *df* to int dtypes."""
+        """Convert 'year'-indexed columns of *df* to :obj:`int` dtypes.
+
+        :meth:`_year_idx` is used to retrieve a sequence of (idx_set, idx_name)
+        for *only* the 'year'-indexed dimensions of item *name*.
+
+        If at least one dimension is indexed by 'year', all such dimensions are
+        converted to :obj:`int`. Otherwise, *df* is returned unmodified.
+        """
         year_idx = self._year_idx(name)
+
         if len(year_idx):
-            return df.astype({c: 'int' for _, c in year_idx})
+            return df.astype({col_name: 'int' for _, col_name in year_idx})
         else:
             return df
 
     # Override ixmp methods to convert 'year'-indexed columns to int
+
     def equ(self, name, filters=None):
         """Return equation data.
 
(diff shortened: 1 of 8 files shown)


khaeru:
> Adjusts other code in `tools.add_year` and tutorials/utils/run_scenarios.py.

Just for posterity, a note about this. In the latter case, the old code looked like this:
https://github.com/iiasa/message_ix/blob/19a00c48507f1181dabda0d0b28255e7f8275135/tutorial/utils/run_scenarios.py#L24-L27

This old method (Platform.Scenario) is deprecated anyway, but it returns an ixmp.Scenario object. ixmp doesn't (shouldn't) know anything about MESSAGE; so later on when `scen.par(...)` is called, a `year` dimension comes back as a `str` rather than `int` column, causing other code to fail. This is a cue that one must always use `message_ix.Scenario` for MESSAGE scenarios:
https://github.com/iiasa/message_ix/blob/df604ff13c7c248c68f57c4bb400f7521b8873f6/tutorial/utils/run_scenarios.py#L24-L27

message_ix/core.py
```diff
@@ -26,6 +27,111 @@ def __init__(self, mp, model, scenario=None, version=None, annotation=None,
...
+        #    performance
+        return list(filter(
+            lambda e: e[0] == 'year',
+            zip(self.idx_sets(name), self.idx_names(name))
+        ))
+
+    def _year_as_int(self, name, df):
+        """Convert 'year'-indexed columns of *df* to int dtypes."""
+        year_idx = self._year_idx(name)
+        if len(year_idx):
```
zikolach: Just a question - will it work in case we later add values to dataframe? How will new year values be handled - as integers (expected) or default (category or inherited dtype)?

message_ix/core.py
```diff
@@ -26,6 +27,111 @@ def __init__(self, mp, model, scenario=None, version=None, annotation=None,
...
+            Name of the equation.
+        filters : dict (str -> list of str), optional
+            Filters for the dimensions of the equation.
+
+        Returns
+        -------
+        pd.DataFrame
+            Filtered elements of the equation.
+        """
+        return self._year_as_int(name, super().equ(name, filters))
```
zikolach: How much efficient is such a post-processing in comparison of having years dimensions converted to integers upon df creation in ixmp core?

message_ix/tools/add_year/cli.py
```diff
@@ -84,12 +83,20 @@ def main(context, model_new, scen_new, create_new, years_new, firstyear_new,
 
     try:
         sc_ref = context.get('scen', None)  # AttributeError if context is None
-        if not isinstance(sc_ref, ixmp.Scenario):
+        if not sc_ref:
```
zikolach: what about `issubclass`?

message_ix/core.py
khaeru: You're talking about something like:
```
df = scen.par('parname')
df = df.append(rows)
df.loc[:, 'year'] = 9999
```

When `rows` has 'year' values, they will be handled as `int`. So this means that they (or the literal `9999`) are not restricted to a limited set of categorical values, e.g. the contents of `scen.set('year')`.

However, if one then tries to `scen.add_par('parname', df)`, then it will fail.

I think this is the current behaviour, i.e. one can construct any kind of DataFrame, but ixmp/message_ix does not complain about it until one adds it.

Does this seem adequate?

message_ix/core.py
khaeru: To be frank, I don't know. I hope pandas' dtype conversion is fast…

Having it done in ixmp core would require the extension to the Backend API that we discussed [here](https://github.com/iiasa/message_ix/issues/268#issuecomment-557885384).

message_ix/tools/add_year/cli.py
khaeru: Good call.

khaeru:
> I would like to have more examples in docstrings as some utility methods (for me barely familiar with the code) are really confusing (e.g. `_year_idx`).

Okay, will do.

khaeru:
Thanks for the weekend review, @zikolach. Merging now—please either mark those two comment threads as resolved, or we can open a new issue for each one to be resolved later.

message_ix/core.py
zikolach: My original observation (or surprise) was about different treating empty and non-empry dataframes.

Issue #51, opened 2018-06-19
Mistakes in documentation of the parameters


ClaraLuisa:
Some parameters are described inaccurately on http://messageix.iiasa.ac.at/. eg  "historical_emission" should be something like:
 - node | type_emission | type_tec | type_year 

instead it reads:

- node | emission | type_tec | year

khaeru:
I came across this while looking for documentation-related issues.

The docstring seems to correctly match the definition:
https://github.com/iiasa/message_ix/blob/154c58150755f1445ed48de028656daefccde995/message_ix/model/MESSAGE/parameter_def.gms#L483-L487

The parameter is used here:
https://github.com/iiasa/message_ix/blob/154c58150755f1445ed48de028656daefccde995/message_ix/model/MESSAGE/model_core.gms#L1701-L1708

As far as I can tell:
- The historical emissions are added to projected emissions *for each species*—i.e. each member of the set `emission`, e.g. "CO2", "CH4", etc.—separately.
- *After* this, emissions are summed across sets of species (e.g. "GHG"s) appearing in `type_emission`.
- `year_all` is the set of all historical *and* future years.
- Other parameters like `bound_emission` may be defined as applying not only to particular years, but to "cumulative" emissions; "cumulative" is a member of `type_year` with the special meaning "sum across `year_all`".
- But, historical emissions are specific to particular `year`s.

So the definition and description here seem to be correct.

Can @OFR-IIASA, @danielhuppmann or someone else confirm this is the right interpretation?

francescolovat:
Hi @ClaraLuisa ,

have you found other innaccuracies like the one you presented here?

francescolovat:
@ClaraLuisa I saw there are inconsistencies when defining the sets `node` and `node_loc` to describe dimensions of different parameters... Do you know which is the criteria of them? Please see my comment on this in the #275 for a detailed description of it.

ClaraLuisa:
@khaeru thank you for the explanation. I was confused as both definitions work for adding the parameter.

glatterf42:
I'm closing this issue since @khaeru's understanding seems to be correct. As far as I can tell, @francescolovat tried adapting the parameter definition as suggested here, which didn't work. He then reverted the change because it "ha[d] no sense and [was] wrong", so if it is an actual error or if you find any new errors, please create a new issue for them :)

Issue #241, opened 2019-10-08
Add-on parameter documentation is incomplete


khaeru:
The [documentation for add-on technology parameters](https://message.iiasa.ac.at/en/master/model/MESSAGE/parameter_def.html#parameters-for-the-add-on-technologies) contains the sentence fragment “The upper bound of”.

- This appears to have come in with #91 which incorporated #83; so @danielhuppmann can probably complete the sentence.
- #51 could perhaps be fixed at the same time.

PR #275, opened 2019-11-27
Addition of parameter definitions in MESSAGEix documentation

### LINKED ISSUE #260: documentation of parameters only lists dimensions, but lacks description
https://github.com/iiasa/message_ix/blob/71e52bd1b3635243adbca4a4b5ceddbc487807a8/message_ix/model/MESSAGE/parameter_def.gms#L18

With exception of very few 'general' parameters, a simple of description of parameters is not included in the MESSAGEix documentation and would be very helpful and straight forward to add as this information is included in the parameter definitions in the GAMS code.

GamzeUnlu95:
Closes #260.
Two other issues that are related and might also be addressed in this PR: #241 and #51.

This PR updates the parameter documentation and adds explanations for the parameters that are not self-explanatory. The parameter explanations are based on the definitions provided in `parameter_def.gms` file. 

Even though the initial idea was to add the definitions in the tables for every parameter, after a discussion with @volker-krey , we agreed that it is better to add these definitions only for parameters that are not clear since most of the parameters are already self-explanatory (e.g. bound_activity_up, construction_time etc.) and the tables look exhaustive (especially in the `pdf` version of the documentation) when all the definitions are added. 

The additional information and related links are added to the footnotes and to the explanation paragraphs for the parameters/sections that might require more explanation. (e.g. rating_bin, add-on technologies, cost parameters for ‘soft’ relaxations of dynamic constraints, auxiliary investment cost parameters and multipliers, emission scaling, relations, resources). 

These additions are based on the `parameter_def.gms` file which includes only basic definitions of the parameters. The reviewers could suggest changes to the explanations in case they are not clear enough (Especially for the section of **"Auxiliary investment cost parameters and multipliers"**).   

The documentation page can be built by using the command: `sphinx-build [source-directory] [output-directory]`

source-directory: `C:\....\message_ix\doc\source`  
output directory: `C:\....\message_ix\doc\build`

### DIFF
--- a/message_ix/model/MESSAGE/parameter_def.gms
+++ b/message_ix/model/MESSAGE/parameter_def.gms
@@ -73,6 +73,11 @@ Parameters
 * Parameters of the `Resources` section
 * -------------------------------------
 *
+* The quantity of the resources in MESSAGEix is represented by two parameters: ``resource_volume`` and ``resource_remaining``.
+* The first parameter is the volume of resources at the start of the model horizon while the second is the maximum extraction relative
+* to remaining resources by year. The maximum extraction of the resources (by grade) in a year is constrained by the parameter
+* ``bound_extraction_up``. Extraction costs for resources are represented by ``resource_cost parameter``. 
+*
 * .. list-table::
 *    :widths: 25 75
 *    :header-rows: 1
@@ -92,8 +97,8 @@ Parameters
 *    * - historical_extraction [#hist]_
 *      - ``node`` | ``commodity`` | ``grade`` | ``year``
 *
-* .. [#stock] This parameter allows (exogenous) additions to the commodity stock over the model horizon,
-*    e.g., precipitation that replenishes the water table.
+* .. [#stock] Commodity stock refers to an exogenous (initial) quantity of commodity in stock. This parameter allows
+*    (exogenous) additions to the commodity stock over the model horizon, e.g., precipitation that replenishes the water table.
 *
 * .. [#hist] Historical values of new capacity and activity can be used for parametrising the vintage structure
 *    of existing capacity and implement dynamic constraints in the first model period.
@@ -130,8 +135,9 @@ Parameter
 *    as an auxiliary reporting variable; it equals ``demand_fixed`` in a `MESSAGE`-standalone run and reports
 *    the final demand including the price response in an iterative `MESSAGE-MACRO` solution.
 *
-* .. [#peakload] The parameters ``peak_load_factor`` and ``reliability_factor`` are based on the formulation proposed
-*    by Sullivan et al., 2013 :cite:`sullivan_VRE_2013`. It is used in :ref:`reliability_constraint`.
+* .. [#peakload] The parameters ``peak_load_factor`` (maximum peak load factor for reliability constraint of firm capacity) and
+*    ``reliability_factor`` (reliability of a technology (per rating)) are based on the formulation proposed by Sullivan et al., 2013 :cite:`sullivan_VRE_2013`.
+*    It is used in :ref:`reliability_constraint`.
 *
 ***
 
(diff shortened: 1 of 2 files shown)


khaeru:
@GamzeUnlu95 thanks for this. I have turned on the automatic RTD builds of this branch, and they seem to complete succesfully! Here is the page you modified: https://message.iiasa.ac.at/en/parameter_doc_new/model/MESSAGE/parameter_def.html

Please solicit others' input, in the form of comments on this PR, as to which (Option 1 or Option 2) is the clearest form for the documentation.

khaeru:
Also note that, per our in-person discussion, message_ix tests are currently failing because iiasa/ixmp#225 needs to be completed, reviewed, and merged.

khaeru:
@GamzeUnlu95 —commit 800c35a seems to revert the addition of descriptions and the Option 2 table that were added by bb8a35b. Was that intentional?

GamzeUnlu95:
@khaeru , that was intentional. After converting to pdf and a discussion with Volker, we decided not to include all the definitions in the tables, but include them in the footnotes or in the paragraphs where it is relevant. I also updated the PR definition.

francescolovat:
Hi @GamzeUnlu95,
thanks for you work. 

I've edited the description of the PR to include that closes issue #260. You've used the command _closes_ the other way around: you wrote closes #275 in the issue itself. The command closes issues from PRs. Therefore I also deleted that comment in #260.

Now I'm going through what you've wrote and I'll evidence very minor detail and also push a commit correcting a typo to also see if all the CIs pass after iiasa/ixmp#225 was merged.

message_ix/model/MESSAGE/parameter_def.gms
```diff
@@ -200,9 +206,14 @@ Parameter
...
-*    (:ref:`reliability_constraint`) and flexibility (:ref:`flexibility_constraint`) can depend on the share
-*    of the technology output in the total commodity use at a specific level.
+* .. [#rating] Maximum share of technology in commodity use per rating. The upper bound of a contribution by any technology to the constraints on system reliability
+*    (:ref:`reliability_constraint`) and flexibility (:ref:`flexibility_constraint`) can depend on the share of the technology output in the total commodity use at
+*    a specific level.
+*
+* .. [#flexfactor] Contribution of technologies towards operation flexibility constraint. It is used in :ref:`flexibility_constraint`.
+*
+* .. [#renewables] ``renewable_capacity_factor`` refers to the quality of renewable potential by grade and ``renewable_potential`` refers to the size of the renewable potential per grade. 
+*
```
francescolovat: I don't understand the sentence: _refers to the quality of renewable potential by grade_ here. Do you think it is enough? Probably I missed other previous descriptions that could help me understand it.

message_ix/model/MESSAGE/parameter_def.gms
```diff
@@ -73,6 +73,11 @@ Parameters
 * Parameters of the `Resources` section
 * -------------------------------------
 *
+* The quantity of the resources in |MESSAGEix| is represented by two parameters: ``resource_volume`` and ``resource_remaining``.
+* The first parameter is the volume of resources at the start of the model horizon while the second is the maximum extraction relative
+* to remaining resources by year. The maximum extraction of the resources (by grade) in a year is constrained by the parameter
```
francescolovat: I don't understand the sentence: _while the second is the maximum extraction relative to remaining resources by year_ here. Do you think it is enough? Probably I missed other previous descriptions that could help me understand it.

Shouldn't it be that the second are just the remaining sources per year?

message_ix/model/MESSAGE/parameter_def.gms
```diff
@@ -450,17 +452,17 @@ Parameters
...
 * ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 *
-* Auxilary investment cost parameters include the remaining technical lifetime at the end of model horizon in addition to the
-* different scaling factors and  multipliers as listed below. These factors account for remaining capacity or construction time of new capcaity, 
-* the value of investment at the end of mdoel horizon or the discount factor of remaining lifetime beyond model horizon. 
-* 
+* Auxiliary investment cost parameters include the remaining technical lifetime at the end of model horizon (``beyond_horizon_lifetime``) in addition to the
+* different scaling factors and multipliers as listed below. These factors account for remaining capacity (``remaining_capacity``) or construction time of new capacity (``construction_time_factor``),
+* the value of investment at the end of model horizon (``end_of_horizon_factor``) or the discount factor of remaining lifetime beyond model horizon (``beyond_horizon_factor``).
+*
```
francescolovat: please @GamzeUnlu95 have a look if I inserted the right parameters into the right pieces of the paragraph. It was not clear for me that's why I wanted to explicitly relate the text to the parameters listed below.

francescolovat:
Looking for some other wrong parameter definitions in [Parameter definition page](https://message.iiasa.ac.at/en/master/model/MESSAGE/parameter_def.html), as emphasized in #51, I have a doubt now:

- which is the difference between `node_loc` and `node` used in the tables in the [Parameter definition page](https://message.iiasa.ac.at/en/master/model/MESSAGE/parameter_def.html)? `node_loc` is not even defined in the [Sets definition page](https://message.iiasa.ac.at/en/master/model/MESSAGE/sets_maps_def.html), so I don't really know why it's being used to describe parameters and what differentiates it from `node`. 

I thought probably the aim was to differentiate the _sets_ **location** and **node** from the `model_core.gms` file. But, as you can see in the snapshot below, both of them have indistinctly been used to describe parameters having **node** and **location** as dimensions in `model_core.gms`.

Can anyone help me interpret them so I can properly edit their dimensions and eventually close #51 ? @OFR-IIASA @volker-krey @khaeru  

image

khaeru:
@francescolovat —that's a great question. Let's also CC @danielhuppmann @gidden per #254.

There are at least three things going on here:

1. In ixmp (Python/Java code), every dimension of a parameter has both an index **set** and an optional **index/dimension name**.
   - See e.g. the [init_par() docs](https://message.iiasa.ac.at/projects/ixmp/en/master/api-python.html#ixmp.Scenario.init_par).
   - The *sets* must be actual sets in the Scenario.
   - The *names* can be anything; e.g. you can have a dimension indexed by `year` but named `'foo'`.
2. In the MESSAGE GAMS code, there are a number of **aliases for sets**.
   - See [sets_maps_def.gms](https://github.com/iiasa/message_ix/blob/master/message_ix/model/MESSAGE/sets_maps_def.gms).
   - For instance, the lines: https://github.com/iiasa/message_ix/blob/d7a75e63e7d16632c7bd0ed88b0a220b70a8c42b/message_ix/model/MESSAGE/sets_maps_def.gms#L133-L136 show that e.g. `node_share` is always an alias of `node`.
   - This is needed because a GAMS parameter that has the *same* index set for *two* of its dimensions can't re-use the same name, e.g. you can't define a GAMS `parameter foo(year, year);`.
3. See #254. While the GAMS definition of MESSAGE's structure is in the .gms files, there is (must be) a parallel definition of the same structure for the Python code to use. At the moment, this is (wrongly, IMO) hidden in the non-open ixmp_source Java package.
   - See for instance here: https://github.com/iiasa/ixmp_source/blob/6c2afd932e10e284ff12c4a20a0d62372f609228/src/main/java/at/ac/iiasa/ixmp/modelspecs/MESSAGEspecs.java#L877 and here: https://github.com/iiasa/ixmp_source/blob/6c2afd932e10e284ff12c4a20a0d62372f609228/src/main/java/at/ac/iiasa/ixmp/modelspecs/MESSAGEspecs.java#L76-L82
    - These lines are basically equivalent to: `scen.init_par('historical_new_capacity', idx_sets=['node', 'technology', 'year'], idx_names=['node_loc', 'technology', 'year_vtg'])`

So what you've uncovered (I hadn't noticed myself!) is:
- There is not necessarily a correspondence between a **dimension name** on the Python side and any **alias** on the GAMS side.
  - E.g. `node_loc` is a dimension name, but there is no GAMS alias.
  - E.g. `land_emission` is defined over the GAMS alias `location`, but on the Python side it does not have a dimension name.
- Some index names, like `node_loc`, are **hidden in the Java code** and thus the correspondence between them and particular sets can only be guessed at.

I hope this helps to understand the situation and to propose improvements.

francescolovat:
Yes, thanks a lot @khaeru , I didn't know about that aliases in GAMS were needed. I'll go through the links you suggested to understand the whole think better. Btw, I've just realized that the last commit trying to help solving the issue #51 has no sense and it's wrong. I'll revert it now.

danielhuppmann:
To quote Douglas Adams: "I apologize for the inconvenience".

There are indeed many things going on here, and it is important to keep two things separate:

1) **data handling** - think of a `pandas.DataFrame` representing parameter data
   Here we need to have:
   - **consistency checking**, so that an entry in the node-column (e.g., where a technology is located) exists in the set `node`. This is implemented via `index_set` (and can be overridden by setting `index_set='*'`).
   - **column naming**: each column of a dataframe has to have a unique name so that we easily work with it. The column names should be a trade-off between intelligibility and consistency (for example using `node_loc` for the column that specifies a location for all parameters related to a technology).
2) **mathematical formulation**
    The equations of the optimisation problem are quite elaborate to make life easier for a user. For example, the emissions bound of a node automatically considers the emissions of all sub-nodes - otherwise, a user would have to specify specific bounds of sub-nodes (determining the allocation of the constraint between sub-nodes a priori), or make some other elaborate parameter specifications.
    This is implemented in GAMS using multiple nested loops, and this requires aliases to correctly implement the summations or if-else-statements. The primary goal of the alias naming is to make it easy for users to understand an equation.

These two use cases are, in principle, not related.

PR #276, opened 2019-11-27
Update add_year for turning off some warnings and adding bound extension option

behnam-zakeri:
This PR modifies the tool `add_year` for:
turning off some warning in python and adding an option to extend the bound

**PR checklist:**

- [X] Tests added
- [x] Documentation added.
- [X] Release notes updated.

PR #281, opened 2019-12-11
Correct logic of Scenario.years_active()

### LINKED ISSUE #278: years_active() returns wrong years
The function years_active() was introduced as part of PR: Adjust to ixmp Backend/Model refactor (#249). The function works when period lengths are all the same, but not when they change and a technology lifetime extends over such a switch in period length.

The issue is traced back to the following line.
data.where(data.age <= lt)['year'].dropna().astype(int).tolist()

Based on an example in the global model: A powerplant built in 1990 with a 30 year lifetime will still be active in 2020, but only for 5 of the 10 year period. The cumulative sum for the years in 2020 will be 35 years, meaning that based on the above line, the returned active_years will only include 2010.

### LINKED ISSUE #279: vintage_and_active_years() fails when there are no 'year_pairs'
in the function vintage_and_active_years(), if active years are requested for a vintage which does not extend into or beyond the first model year, an error occurs in the line:
v_years, a_years = zip(*year_pairs)

a possible solution could be to change this line and the return statement as follows:
from
        v_years, a_years = zip(*year_pairs)
        return pd.DataFrame({'year_vtg': v_years, 'year_act': a_years})
to
        if year_pairs:
                v_years, a_years = zip(*year_pairs)
                return pd.DataFrame({'year_vtg': v_years, 'year_act': a_years})
       else:
                return pd.DataFrame(columns=['year_vtg', 'year_act'])

khaeru:
Closes #278.
Closes #279.

- `Scenario.years_active()` and `.vintage_and_active_years()` are improved.
- `test_years_active()` is rewritten to be more explicit.
- `test_years_active()` was **incorrect**. In this PR, the period '2050' is removed from the expected return value. See the inline comments: https://github.com/iiasa/message_ix/blob/3585ab3eafda1be8d2e24e1fe6190695e07f1410/tests/test_core.py#L222-L228

I have followed our documentation, specifically [here](https://message.iiasa.ac.at/en/stable/model/MESSAGE/sets_maps_def.html#id4) where it reads:
> In MESSAGEix, the key of an element in set year identifies the last year of the period, i.e., in a set year = [2000, 2005, 2010, 2015], the period ‘2010’ comprises the years [2006, .., 2010].

IMO this is more confusing than it needs to be, and we should rename 'year' to 'period' to disambiguate.

**PR checklist:**

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,6 +10,7 @@
 
 ## All changes
 
+- [#281](https://github.com/iiasa/message_ix/pull/281): Test and improve logic of `years_active` and `vintage_and_active_years`.
 - [#269](https://github.com/iiasa/message_ix/pull/269): Enforce 'year'-indexed columns as integers.
 - [#256](https://github.com/iiasa/message_ix/pull/256): Update to use :obj:`ixmp.config` and improve CLI.
 - [#255](https://github.com/iiasa/message_ix/pull/249): Add :mod:`message_ix.testing.nightly` and `nightly` CLI command group for slow-running tests.
--- a/message_ix/core.py
+++ b/message_ix/core.py
@@ -56,6 +56,9 @@ class Scenario(ixmp.Scenario):
 
         if len(year_idx):
             return df.astype({col_name: 'int' for _, col_name in year_idx})
+        elif name == 'year':
+            # The 'year' set itself
+            return list(map(int, df))
         else:
             return df
 
--- a/message_ix/core.py
+++ b/message_ix/core.py
@@ -292,15 +292,16 @@ class Scenario(ixmp.Scenario):
 
         Parameters
         ----------
-        ya_args : tuple of (node, technology, year_vtg), optional
-            Arguments to :meth:`ixmp.Scenario.years_active`.
+        ya_args : tuple of (node, tec, yr_vtg), optional
+            Arguments to :meth:`years_active`.
         in_horizon : bool, optional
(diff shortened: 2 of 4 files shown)


OFR-IIASA:
@khaeru I have managed to switch branches and this looks fine to me. Should we also resolve issue #279 with this PR?

khaeru:
> Should we also resolve issue #279 with this PR?

Yes, I already did. See the description at the top, and the commit message for 5f6deae.

khaeru:
Note that the CI checks passed for cca6a25, then failed for 0022513. This seems to be the same symptom as noted [here on #276](https://github.com/iiasa/message_ix/pull/276#issuecomment-561605988).

khaeru:
> Note that the CI checks passed for [cca6a25](https://github.com/iiasa/message_ix/commit/cca6a25b369111558a642524c8c931c959a16e27), then failed for [0022513](https://github.com/iiasa/message_ix/commit/00225133302ee389e9e636823d5699ce6bb01a1f). This seems to be the same symptom as noted [here on #276](https://github.com/iiasa/message_ix/pull/276#issuecomment-561605988).

66d5b4d brings in the fix for this issue from `master`. If the checks pass with this change, I will merge.

khaeru:
Total coding time on this, #278, #279, and #283: 2 hours 20 minutes.
Discussion, filing #282, monitoring CI: about 1:00.

PR #291, opened 2019-12-19
Update install from source docs to include pip --editable option

francescolovat:
PR to include `--edtiable` in _MESSAGEix_ docu to be able to run tests. It is present in _ixmp_ docu but **not** in _MESSAGEix_. I discovered after reinstalling MESSAGEix following the docu with the command `pip install .` and then running the tests: `nightly.py` was not skipped and therefore prompting an interruption error.

Also added in `.gitignore.txt` PyCharm's `/.idea` folder (which is automatically created every time you start a project with it).

Should I add sth in `RELEASE_NOTES.md` ? @khaeru 

- [x] Documentation added.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,7 @@ tests/data/testdb/ixmptest.lck
 .pytest_cache
 coverage.xml
 htmlcov
+
+# JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and WebStorm
+/**/.idea
+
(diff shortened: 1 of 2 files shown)


khaeru:
@francescolovat thanks! Can you please update the PR title to something like "Update install from source docs to include pip --editable option"?

> - [ ] Documentation updated.

That's what this PR does, so you can probably check that box—right?

> - [ ] Release notes updated.

No, I don't think this is necessary.

francescolovat:
@khaeru thanks! Do you apporve now? So then I'll merge :)

PR #265, opened 2019-11-20
Updating read_excel for initializing new sets and parameters

behnam-zakeri:
In response to issue #264, this PR modifies `read_excel` to have the option to initialize any sets and parameters that are not in default lists of a `message_ix` scenario, when reading that from Excel.

- [X] Tests added: existing `test_excel_read_write(test_mp)` modified to test the reading of new sets and parameters
- [X] Documentation added: Not needed
- [X] Release notes updated.

PR #321, opened 2020-03-21
Transfer Excel data I/O to ixmp

khaeru:
This PR removes the `message_ix.Scenario.to_excel()` and `.read_excel()` methods. ~iiasa/ixmp#286~ iiasa/ixmp#289 adds these to `ixmp.Scenario`, so they are available for non-MESSAGE scenarios.

It also incorporates test improvements from #265.

## How to review

- ~Wait for iiasa/ixmp#289 to merge, or check out its branch.~
- Run tests against the updated ixmp code.
- Try with any old Excel exports one has lying around.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -4,6 +4,7 @@ Next release
 All changes
 -----------
 
+- `#321 <https://github.com/iiasa/message_ix/pull/321>`_: Move :meth:`.Scenario.to_excel`, :meth:`.read_excel` to :class:`ixmp.Scenario`; they continue to work with message_ix.Scenario.
 - `#323 <https://github.com/iiasa/message_ix/pull/323>`_: Add `units`, `replace_vars` arguments to :meth:`.Reporter.convert_pyam`.
 - `#308 <https://github.com/iiasa/message_ix/pull/308>`_: Expand automatic reporting of add-on technologies.
 - `#313 <https://github.com/iiasa/ixmp/pull/313>`_: Include all tests in the message_ix package.
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -55,6 +55,10 @@ Support for R usage of the core classes is provided through the `reticulate`_ pa
    included here for convenience. :class:`message_ix.Scenario` defines
    additional methods specific to |MESSAGEix|:
 
+   .. versionchanged:: 3.0
+
+      :meth:`.read_excel` and :meth:`.to_excel` are now methods of :class:`ixmp.Scenario`, but continue to work with message_ix.Scenario.
+
    .. autosummary::
 
       add_cat
@@ -71,7 +75,6 @@ Support for R usage of the core classes is provided through the `reticulate`_ pa
       var
       vintage_and_active_years
       years_active
-
 
 
 Model classes
--- a/message_ix/core.py
+++ b/message_ix/core.py
@@ -4,7 +4,7 @@ from itertools import product
 import logging
 
 import ixmp
-from ixmp.utils import as_str_list, pd_read, pd_write, isscalar, logger
+from ixmp.utils import as_str_list, isscalar
 import pandas as pd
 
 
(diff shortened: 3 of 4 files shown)


behnam-zakeri:
Thanks a lot @khaeru for cleaning up here and moving this to ixmp. 
I tested this with the existing test (`core.test_excel_read_write`) and it passes the test on my machine as is. So, now it does initialize if a set or parameter is in Excel but not in the scenario default items. That's great.
However, I believe the behaviour of the functionality has changed a bit. Because:
1- The method `to_excel()` now  is trying to save output variables of a scenario as well. Previously `to_excel()` only saved sets and parameters. This was enough for a user to read the data back and solve the scenario.
2- Due to this change in behaviour, the functionality fails if a scenario has a solution. Because there is a variable _DEMAND_ in message_ix, which has exact the same name but in upper case as parameter _demand_. This makes pandas xlsxwriter raising the following error:
```
  File "C:\Users\...\anaconda3\lib\site-packages\xlsxwriter\workbook.py", line 805, in _check_sheetname)

DuplicateWorksheetName: Sheetname 'DEMAND', with case ignored, is already in use.
```
3- I personally think having variables written in Excel is not needed while it can be useful sometimes, especially for small scenarios. Because for scenarios with large data, there is already a memory burden in writing all data in Excel. I assume with many variables added to this, it becomes more difficult.
4- Previously, only those sets and parameters were written in Excel, which were not empty. Now it seems all items are written, making many empty sheets, which makes it difficult for the user to browse. 
5- Suggestion: Usually for the user it's difficult to browse the sheets in Excel, even when there is data for the sheets, as there are many items and not alphabetically ordered. I understand this is because this comes directly from `Scenario.set_list()` and `Scenario.par_list()`, which are not ordered alphabetically. If possible, it's good to have an alphabetical order for sets and parameters separately. I mean the Excel sheets start from the left hand side with sets but orderd, and then ordered parameters. 
Todo:
a. I'll add a line to the test to check the functionality when a scenario has a solution (to check No. 2 above).
b. It's good to ignore writing those items that are empty (No. 4 above). Is this easy for you to add or I can do that?
c. Any feedback welcome for No. 3 and 5 above.

khaeru:
Thanks for these detailed comments, and for catching the issue with `demand`/`DEMAND`. I think they help to clarify what our requirements are here. Picking out just a couple points:

- > Previously to_excel() only saved sets and parameters. This was enough for a user to read the data back and solve the scenario.

  This requirement (call it **R1**) is “the Excel file contains data necessary to solve a Scenario.”
- > Now it seems all items are written, making many empty sheets, which makes it difficult for the user to browse.

  This requirement (**R2**) is “the Excel file should be easy for the user to browse,” e.g. without having to skip sheets which are empty, or to hunt through sheets which are out of order.

- Since these weren't written anywhere, I guessed the requirement was (**R3**) “the Excel file contains [all] the data of a Scenario,” i.e. it's sufficient to make `s2.read_excel(...)` *identical* to `s1.to_excel()`. The difference gives rise to some other things you point out, e.g.:
  - R1 was previously met while omitting variables and equations, i.e. not meeting R3.
  - R2 was previously met by omitting empty sets and parameters, which conflicts with R3: without writing info for empty sets/parameters, they can't be initialized, so the Scenario as-read cannot be identical.

As @zikolach pointed out (see https://github.com/iiasa/ixmp/issues/292#issuecomment-604999454), we can imagine multiple ways to change the format. But I don't think that's very urgent and—as I commented on that issue—people have existing Excel files sitting around that they will want to continue to use. So I would like to minimize disruption.

Here's what I'd propose. All these changes will go in ixmp:
1. Change `to_excel()` to write only sets and parameters by default, with an option to write variables and equations. This should address @behnam-zakeri's points (1), (2), and (3).
2. As you suggest, change the order of sheets to the following groups. Sheets within each group will be in alphabetical order.
   a. All sets _with_ data.
   b. All parameters _with data_.
   c. All sets _without_ data.
   d. All parameters _without_ data.
   e. Equations (if the option is selected).
   f. Variables (if the option is selected).

   This helps with R2 because the user can focus on a and b. It shouldn't make the `to_excel()` code much more complicated, and requires no change to the `read_excel()` code.

Some other reactions:
- > Because for scenarios with large data, there is already a memory burden in writing all data in Excel. I assume with many variables added to this, it becomes more difficult.

  Can you say how this memory burden manifests? Is it just the sheet rows limit, or something else? Ideally, we should add a test of a MESSAGEix-GLOBIOM-sized Scenario (i.e. full of random data, but the same number of elements in parameters) and test that the code works; kind of similar to [this code](https://github.com/iiasa/ixmp/blob/3e16ec8d0f2d3888dcf1bd5195cbaa4228ad3949/ixmp/tests/reporting/test_reporting.py#L558-L615).
- >  there is a variable `DEMAND` in message_ix, which has exact the same name but in upper case as parameter `demand`. This makes pandas xlsxwriter raising the following error

  This is interesting: on my system I think openpyxl is used, not xlsxwriter. I wonder if this is a difference in the two packages, or a limitation of the Excel format itself. Will investigate.

khaeru:
>  On my system I think openpyxl is used, not xlsxwriter. I wonder if this is a difference in the two packages, or a limitation of the Excel format itself. Will investigate.
```python
for engine in 'openpyxl', 'xlsxwriter': 
    xw = pd.ExcelWriter(f'{engine}.xlsx', engine=engine) 
    df = pd.DataFrame()
    df.to_excel(xw, sheet_name='foo') 
    df.to_excel(xw, sheet_name='FOO') 
    xw.close() 
```
What I observe:
- With engine='openpyxl', the code completes successfully, but the file has two sheets named 'foo', and 'FOO1', i.e. the '1' is automatically appended.
- With engine='xlsxwriter', the exception reported by @behnam-zakeri is reported.

So this is likely a limitation of the Excel format *per se*, and openpyxl contains a workaround that would, in any case, defeat our code.

khaeru:
In order to keep moving, I went ahead and implemented the suggestion in the above comment as iiasa/ixmp#297. @behnam-zakeri @Jihoon can either of you confirm that the combination of these two branches works as desired?

khaeru:
As with iiasa/ixmp#297, will merge this without review in order to clear the ground for hackathon PRs. If needed, we can make another to adjust, later.

behnam-zakeri:
Thanks @khaeru for taking this forward. The solutions you suggested are excellent, for making sets and parameters default data to be written, and for ordering the sheets. And interesting test for Excel engine you illustrated.
related to your question:

> Can you say how this memory burden manifests?

 I meant memory burden for writing many rather large parameters to multiple Excel sheets, and not particularly exceeding row limits.
To put an end to this, I suggest we also resolve the issue of parameters exceeding the row limits of Excel. My proposal is to write such parameters in two or more sheets, like: land_output(1), land_output(2), etc. And for reading back the data to ixmp, it should be quite straightforward to drop these parentheses and add several dataframes one after each other. I can make an issue and change the code for this proposal is you think it makes sense.

khaeru:
iiasa/ixmp#292 is the follow-up issue for modifying the Excel file format to extend past the ~10⁶ row limit. It would be great if you wanted to tackle this, but no pressure. If so, please create an *ixmp* (here is message_ix) branch named e.g. `issue/292`.

Also note that @volker-krey supplied example code in iiasa/ixmp#300 that addresses the same limitation by establishing a new, zipped-CSV format, which might be rolled into the same PR.

behnam-zakeri:
Thanks @khaeru for the instructions. I'll follow this on ixmp.

PR #328, opened 2020-04-01
Add reporting improvements from the 2020 hackathon

khaeru:
- `Reporter.convert_pyam()`
  - The argument *year_time_dim* is passed as a keyword argument. If passed as a string positional argument ('y'), the Reporter internally interpreted it as a reference to the key, 'y', that yields the list of years. This list is supplied to `as_pyam()`, which is expecting just the string 'y'.
  - Avoid manipulating units if the pd.DataFrame is empty.
- Add auto-reported `price emission` quantity = `PRICE_EMISSION` broadcast across
- `computations.map_as_qty` now automatically adds a map from 'all' to all elements in the target set, e.g. of all technologies, `t`. This exists in GDX and Java but isn't returned by the standard methods, for some reason.

## How to review

- Note that CI checks pass.

## PR checklist

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -4,6 +4,7 @@ Next release
 All changes
 -----------
 
+- `#328 <https://github.com/iiasa/message_ix/pull/328>`_: Expand automatic reporting of emissions prices and mapping sets; improve robustness of :meth:`Reporter.convert_pyam`.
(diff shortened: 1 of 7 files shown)


khaeru:
Thanks @Jihoon for the assist with the test! :muscle:

Issue #251, opened 2019-10-28
Remove install.bat; clarify R usage


khaeru:
I sat with @guofei2016 today who was trying to reinstall both ixmp and message_ix for use from R on Windows. We encountered some difficulties; see also iiasa/ixmp#196.

Per these, the following changes to message_ix are needed:
- Remove install.bat.
- Clarify in documentation that the only steps required to use message_ix from R are:
  1. Install rixmp.
  2. Give ``message_ix <- reticulate::import('message_ix')``.
- Remove rmessage_ix. The only code it contains needlessly duplicates code from Python message_ix.__init__; is broken; and is not tested.
- (Maybe?) Add minimal tests of message_ix from R.

PR #329, opened 2020-04-02
update R tutorial and include testing

adrivinca:
I updated the name and some parts of the Austrian R tutorials and updated the documentation with new links. I also added some code that run the R tutorial while testing if the flag option `--test-r` is used.

## How to review

- Build the documentation and check the installation part and the tutorial page related to R
- run the tests including the R tutorials, as described in the documentation `py.test --test-r`

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -71,6 +71,11 @@ Install |MESSAGEix| from source
     $ pip install --editable .[tests,reporting]
     $ py.test
 
+8. **(Optional for R users)** Run the built-in test suite to check that |MESSAGEix| functions
+   correctly on your system, including use the ``rixmp`` package::
+
+    $ py.test --test-r
+
 Common issues
 -------------
 
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -68,8 +68,8 @@ Install |MESSAGEix| from source
 7. (Optional) Run the built-in test suite to check that |MESSAGEix| functions
    correctly on your system::
 
-    $ pip install --editable .[tests]
-    $ py.test tests
+    $ pip install --editable .[tests,reporting]
+    $ py.test
 
 Common issues
 -------------
--- a/conftest.py
+++ b/conftest.py
@@ -1,4 +1,20 @@
-
+import pytest
 
 # Use the fixtures test_mp, test_mp_props, and tmp_env from ixmp.testing
 pytest_plugins = ['ixmp.testing']
+
+
+# Hooks
+
+def pytest_addoption(parser):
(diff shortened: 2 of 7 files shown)


adrivinca:
Thanks to @guofei2016 I also found that running the tutorials' code on Rstudio fails at solving. It seems a path issue and after some debugging the path seems indeed correct.
Strangely, running on Jupyter or from command line with `Rscript` works

I attached an example, the austrian tutorial written in R.
@guofei2016 or @volker-krey could you please try, after making sure the austrian tutorial works for you, try to run file I send you in Rstudio and then form command line? `Rscript austrian.R`

To check whether these conclusions are correct. Then it would be something with Rstudio that I can not figure out. Thanks!

volker-krey:
@adrivinca related to issues with running message_ix in RStudio, I noticed different setting of the PATH environment variable in Jupyter notebooks (where code works fine) compared to RStudio. The difference that I noticed is that in the Jupyter environment the user-specific part of the PATH environment variable which typically contains the Anaconda related paths, is added before the system part of the PATH variable. If I modify the PATH variable in RStudio manually in this way, then the code also works in RStudio. I have not figured out how to access the user- and system-specific parts of the PATH variable, so just did a quick and dirty manual addition which solves the issue, but is unfortunately not transferable.

khaeru:
@adrivinca @volker-krey something I'm not clear on: have you installed Anaconda and RStudio *in parallel*, or installed the latter *via* the former (e.g. as outlined [here](https://docs.anaconda.com/anaconda/user-guide/tasks/using-r-language/) and [here](https://docs.anaconda.com/anaconda/navigator/tutorials/create-r-environment/))?

There are many ways for a user to set up their system. Our install docs could recommend one, and/or qualify installation advice ("ensure your RStudio installation has the correct environment set up…") so it's specific to the case in which it applies, and/or mention that some combinations are not tested/supported.

Also found this while searching "RStudio anaconda paths": https://community.rstudio.com/t/how-to-add-conda-to-path-in-rstudio-with-new-conda-initialization/12182

adrivinca:
@khaeru I think we both have anaconda and Rstudio installed separately. It is quite unlikely that any R user will have it installed through anaconda, that's why so far the parallel systems is the only case we look at.

@volker-krey I tried the path change you suggested but I still have the same solving issues

khaeru:
Postponing to after v3.0.

khaeru:
To rebase, resolve the conflicts, carry over R CI improvements from ixmp, and address #251, I opened #377 which supersedes this PR. Thanks @adrivinca for getting this started!

PR #354, opened 2020-06-08
Update doctring message-ix-dl

francescolovat:
Added _Docstring_ for the `message-ix dl` command is appropriate, before it was missing.

Notes: **second tentative**, since in https://github.com/iiasa/message_ix/pull/353 I forced-pushed the revert of the only commit in that PR so Git automatically closed it.


francescolovat:
Yes, I agree @khaeru. It was wired since in my local machine that `test_dl` is passing.

PR #387, opened 2020-08-19
WIP: sketch of 2 variants to add commodity flow of capacity variables

volker-krey:
The PR sketches two alternative variants that allow connecting a commodity flow to capacity addition and retirement.
- variant 1: commodity input and output strictly associated with construction and retirement of capacity (commodity flow is strictly triggered in periods when new capacity comes online or is retired, in case of new periods being added, no user action is required)
- variant 2: commodity input and output associated with operation of capacity at any period (more flexible as commodity flows can also be triggered before construction or after retirement, but requires adjustments by user in case new periods are being added to a model)

Comments on the two variants of adding commodity flow of capacity variables are appreciated.


message_ix/model/MESSAGE/parameter_def.gms
```diff
@@ -229,6 +229,16 @@ Parameters
 * technology input-output mapping and costs parameters
     input(node,tec,vintage,year_all,mode,node,commodity,level,time,time)  relative share of input per unit of activity
     output(node,tec,vintage,year_all,mode,node,commodity,level,time,time) relative share of output per unit of activity
+* +++++
+* variant 1: commodity input and output associated with construction and retirement of capacity
+    input_cap_new(node,tec,vintage,node,commodity,level,time)  relative share of input per unit of new capacity built
+    output_cap_new(node,tec,vintage,node,commodity,level,time) relative share of output per unit of new capacity built
```
Jihoon, line 235: I can understand `input_cap_new` will specify commodity input flow for certain capacity construction, but I am not entirely sure how `output_cap_new` is relevant to a commodity flow. I can see we can still keep the `output_cap_new` part for consistency of the formulation, but practically isn't it simpler without the part?

volker-krey: out_cap_new could be used for specifying generation of new scrap upon construction and more broadly I think we want to develop a formulation that allows going beyond a narrow use case that we may have at hand. Based on the comments in the modeling meeting yesterday, I tend towards going with variant 2 which is more flexible as it would also allow adding commodity requirements for maintainig the capacity (e.g., materials for rennovation of a building after a given amount of time).

volker-krey:
@GamzeUnlu95 good point. The implications of premature retirement for continuous commodity flow could be addressed in variant 2 by switching from the CAP_NEW to the CAP variable (equivalent to fix_cost), but to ensure that end-of-life release of commodities is treated correctly, we either have treat commodity flow beyond the lifetime in a special way or switch to variant 1.

volker-krey:
@GamzeUnlu95 please have a look at the revised formulation of variant 2. This now consistently deals with commodity flows during lifetime, but the issue of commodity flows after end-of-life remains unsolved. The question is whether this is an acceptable error.

volker-krey:
One issue that still needs attention are commodity flows upon capacity retirement if the vintage period (when the capacity was built) has a different length than the retirement period (period after last available of capacity). Since the CAP variables are normalized to annual construction, in such a case, the flow triggered upon retirement can be larger or smaller than intended. For the global model this is relevant in the transition from 5- to 10-year steps after 2060. A rescaling will need to be introduced.

GamzeUnlu95 (approved):

khaeru:
In today's MESSAGE meeting, regarding the new items (sets, variables, equations, and parameters) please look at #190, in particular the additions to `message_ix.models.MESSAGE_ITEMS`: https://github.com/iiasa/message_ix/blob/32031901266b0005b8c79a0c4b5b0de59f2c165a/message_ix/models.py#L20-L44

New additions due to this PR should be handled in the same way.

Jihoon:
This one is now superseded by #451.

PR #404, opened 2020-09-23
systematically add tags for documentation sections to enable improved cross-referencing

volker-krey:
For cross-referencing within the MESSAGEix documentation as well as from other documentation (e.g. message_doc), it would be useful to systematically add tags to sections of the auto-documentation in the GAMS code which then allows references of the form :ref:`<message_ix:tag>`  instead of `<https://docs.messageix.org/...>` which is less maintenance friendly. 

## Reviewer instructions

- check whether important documentation section have not been tagged

## PR checklist

- [ ] Update release notes.
- [ ] ?


danielhuppmann:
Thanks @volker-krey, I think this is a useful addition. Other documents can be tagged in a similar manner when it becomes necessary to cross-reference them.

volker-krey:
@danielhuppmann, I would agree that systematically adding tags to sections to all documents is a good idea in general to allow straight forward cross-referencing between different documents. Could you merge this PR then in case you don't have any further comments on this specific PR.

danielhuppmann (approved):
Thank you @volker-krey for this useful addition!

Issue #208, opened 2019-06-19
Discounting is incorrect


gidden:
This issue should be addressed *after* release 1.2. We discovered in the process of that release that the current implementation of discounting is incorrect. The correct implementation is now in an upstream branch named `fix-discounting`, commit 175fb61.

The critical parts of the diff is:

```
diff --git a/message_ix/model/MESSAGE/scaling_investment_costs.gms b/message_ix/model/MESSAGE/scaling_investment_costs.gms
index 22327b2..daea22b 100644
--- a/message_ix/model/MESSAGE/scaling_investment_costs.gms
+++ b/message_ix/model/MESSAGE/scaling_investment_costs.gms
@@ -133,7 +133,7 @@ construction_time_factor(node,inv_tec,year)$( map_tec(node,inv_tec,year) AND con
 beyond_horizon_factor(node,inv_tec,vintage)$( beyond_horizon_lifetime(node,inv_tec,vintage) )
     = sum(last_period,
 * compute the discount factor of the very last year (not period) in the model horizon
-        df_year(last_period) * POWER( 1 / ( 1 + interestrate(last_period) ), duration_period(last_period) ) * (
+        df_year(last_period) * (
 * multiply this by the geometric series of remaining technical lifetime if interestrate of last model period > 0
             (
                 ( 1 - POWER( 1 / ( 1 + interestrate(last_period) ), beyond_horizon_lifetime(node,inv_tec,vintage) ) )
diff --git a/message_ix/model/includes/period_parameter_assignment.gms b/message_ix/model/includes/period_parameter_assignment.gms
index 2293952..7649414 100755
--- a/message_ix/model/includes/period_parameter_assignment.gms
+++ b/message_ix/model/includes/period_parameter_assignment.gms
@@ -94,8 +94,8 @@ loop(year_all$( ORD(year_all) > 1 ),
 df_period(year_all) =
     df_year(year_all) * (
 * multiply the per-year discount factor by the geometric series of over the duration of the period
-        ( ( 1 - POWER( 1 / ( 1 + interestrate(year_all) ), duration_period(year_all) ) )
-       / ( 1 - 1 / ( 1 + interestrate(year_all) ) ) )$( interestrate(year_all) )
+        ( ( POWER( 1 + interestrate(year_all) , duration_period(year_all) ) - 1 )
+                   / interestrate(year_all) )$( interestrate(year_all) )
 * if interest rate = 0, multiply by the number of years in that period
[...]

PR #421, opened 2020-11-10
Fix discounting

MiguelPobleteCazenave:
Fix discounting from forward looking to backward looking.

As described in https://github.com/iiasa/message_ix/issues/208#issue-457950456: This commit reverts correct calculations of discounting to their prior incorrect implementation. The prior values are incorrect because discounting was calculated in forward looking manner, rather than backward looking. That is, the discounting for years in period 1-10 which is defined as year 10, was being calculated for years in 10-19

## How to review

**Required:** describe specific things that reviewer(s) need to do to, in order to ensure that the PR does what it intends.

## PR checklist

- [x] Add or expand tests.
  - [x] Adjust tests of tutorial notebooks.
  - [x] Adjust [message_ix/tests/data/scenarios.yaml](https://github.com/iiasa/message_ix/blob/master/message_ix/tests/data/scenarios.yaml), used for the nightly tests.
  - [x] Ensure there is a complete description of the discounting logic somewhere in the docs.
- [x] Update release notes.


khaeru:
Thanks @MiguelPobleteCazenave. I'll edit the description to add the required points from the PR template, since these cover some things that occur on a first glance at the code.

MiguelPobleteCazenave:
I added a description of the discounting, I would appreciate if someone else looks at it to check whether it is understandable and/or if there are more things that need to be discussed.

volker-krey:
Thanks @MiguelPobleteCazenave. This looks excellent to me. I repeated the calculation of the "within-period" discount factor `df_period(year)` and arrived at the same answer as in the pull request. With the addition of the release notes, I would suggest to merge the pull request.

Issue #418, opened 2020-11-01
Issue with recursive dynamic model


aaronliuqy:
When I set the $SETGLOBAL foresight to any value more than 0 in model_setup.gms, the model fails to solve.
I am using the scenario.solve() command, and even with all constraints relaxed, it still fails to solve.
This is an example error message from the terminal when I set the foresight to 10 using the Austria tutorial model.

Optimal solution found.
Objective :       97345.372405

--- Restarting execution
--- MESSAGE_run.gms(4268) 2 Mb
--- Reading solution for model MESSAGE_LP
--- MESSAGE_run.gms(4158) 3 Mb
    +++ Solve the recursive-dynamic version of MESSAGEix - iteration 2030  +++
    Give it a bit more time, not even two minutes yet...
--- Generating LP model MESSAGE_LP
--- MESSAGE_run.gms(4268) 3 Mb 1 Error
*** SOLVE aborted
--- MESSAGE_run.gms(4268) 3 Mb 1 Error
*** Status: Execution error(s)
--- Job MESSAGE_run.gms Stop 11/01/20 12:59:54 elapsed 0:00:00.271
[I 13:01:44.629 NotebookApp] Saving file at /iiasa-message_ix-ee0287a/tutorial/Austrian_energy_system/austria.ipynb

I have tried it with the Westeros tutorial and a model I built, neither would solve with the recursive dynamic model.

khaeru:
In order to diagnose this issue, please:
- Give the output of `message-ix show-versions`.
- Attach the file MESSAGE_run.lst from the directory that contains the model files you are using.

aaronliuqy:
The output of show versions:

ixmp:        3.1.0
message_ix:  3.1.0
message_data: None

click:       7.1.2
dask:        2.28.0
graphviz:    None
jpype:       1.0.2
… JVM path:  C:\Users\30338\anaconda3\envs\message_env\Library\bin\server\jvm.dll
pandas:      1.1.2
pint:        0.16.1
xarray:      0.16.1
xlrd:        1.2.0
xlsxwriter:  1.3.6
yaml:        5.3.1

iam_units:   installed
jupyter:     installed
matplotlib:  3.3.2
plotnine:    None
pyam:        0.7.0

python:      3.8.5 | packaged by conda-forge | (default, Sep 24 2020, 16:20:24) [MSC v.1916 64 bit (AMD64)]
python-bits: 64
OS:          Windows
OS-release:  10
machine:     AMD64
[...]

PR #430, opened 2020-12-08
MACRO GDP reporting correction and documentation updated 

GamzeUnlu95:
This PR:

- Corrects the reporting of GDP by introducing the` GDP_corr `parameter which includes the deduction of `trade_costs` coming from MESSAGE. 

- Updates MESSAGEix based total_cost parameter in iteration with MACRO. 

- Extends the documentation of the capital constraint section in MACRO to explain the reasoning behind the changes.

### DIFF
--- a/message_ix/model/MACRO/macro_core.gms
+++ b/message_ix/model/MACRO/macro_core.gms
@@ -161,8 +161,17 @@ SUM(node_active,
 * Equation CAPITAL_CONSTRAINT
 * ---------------------------------
 * The following equation specifies the allocation of total production among current consumption :math:`{C}_{n, y}`, investment into building up capital stock excluding
-* the sectors represented in MESSAGE :math:`{I}_{n, y}` and the MESSAGE system costs :math:`{EC}_{n, y}` which are derived from a previous MESSAGE model run. As described in :cite:`manne_buying_1992`, the first-order
-* optimality conditions lead to the Ramsey rule for the optimal allocation of savings, investment and consumption over time.
+* the sectors represented in MESSAGE :math:`{I}_{n, y}` and the MESSAGE system costs :math:`{EC}_{n, y}` which are derived from a previous MESSAGE model run. MESSAGE system
+* costs are all the energy system costs for a certain region including the trade balance from fuel and emissions certificate trade. Therefore, the constraint assures that the net
+* expenditures and savings cannot exceed the total income in a region. As described in :cite:`manne_buying_1992`, the first-order optimality conditions lead to the Ramsey rule for
+* the optimal allocation of savings, investment and consumption over time.
+
+* Since MACRO does not include the trade of normal goods, there is not an explicit component in the equation which takes the trade revenues into account. The trade of energy goods and
+* emission certificates is accounted within :math:`{EC}_{n, y}` and not added as a seperate component as well. The reason for that is, the NEW PRODUCTION equation in MACRO
+* only accounts for non-energy producing part of the economy and the energy system is modeled seperately in MESSAGE. However, the revenues from energy goods and emission allowances
+* would increase the domestic income and therefore should be accounted in the GDP definition. This is achieved via deducting the trade costs of MESSAGE from :math:`{Y}_{n, y}`in GDP
+* reporting. Positive contribution of EC to GDP would be counter balanced by the decrease of consumption :math:`{C}_{n, y}` and investment :math:`{I}_{n, y}` in the certain year as well
+* as by the reduction in capacity to invest and grow the economy for later years. 
 *
 * .. math:: Y_{n, y} = C_{n, y} + I_{r, y} + {EC}_{n, y} \qquad \forall{n, y}
 *
--- a/message_ix/model/MESSAGE-MACRO_run.gms
+++ b/message_ix/model/MESSAGE-MACRO_run.gms
@@ -114,6 +114,7 @@ Parameters
     price_diff_rel(*,node,sector,year_all)
 
     report_iteration(iteration,*)
+    trade_cost_detail(node, commodity, year_all)              'net of commodity import costs and commodity export revenues by commodity, node and year'
 ;
 
 * variables to report back to user if needed
@@ -158,6 +159,11 @@ ctr = ctr + 1 ;
 
 * to keep the model in memory between runs, this will speed up computation
 MESSAGE_LP.solvelink = 1 ;
+
+*----------------------------------------------------------------------------------------------------------------------*
+* update total energy system costs by node and time with information from latest MESSAGE run
+total_cost(node_macro, year) = COST_NODAL.L(node_macro, year) / 1000 ;
+trade_cost_detail(node, commodity, year) = import_cost(node, commodity, year) - export_cost(node, commodity, year) ;
 
 $INCLUDE MESSAGE/model_solve.gms
 
--- a/message_ix/model/MESSAGE-MACRO_run.gms
+++ b/message_ix/model/MESSAGE-MACRO_run.gms
@@ -120,6 +120,7 @@ Parameters
 Variables
     N_ITER
     MAX_ITER
+    GDP_corr(node, year_all)                   corrected gross domestic product (GDP) in market exchange rates for MACRO reporting
 ;
 
 
@@ -235,6 +236,8 @@ COST_NODAL_NET.L(node_macro,year) =
 ;
 
 GDP.L(node_macro,year) = (I.L(node_macro,year) + C.L(node_macro,year) + EC.L(node_macro,year)) * 1000 ;
+GDP_corr.L(node_macro,year) = (I.L(node_macro,year) + C.L(node_macro,year) + EC.L(node_macro,year) - ( trade_cost(node_macro, year) * 1E-6 ))*1000;
+
 
 * calculate convergence level (maximum absolute scaling factor minus 1 across all regions, sectors, and years)
 max_adjustment_pos = smax((node_macro,sector,year)$( NOT macro_base_period(year) AND demand_scale(node_macro,sector,year) > 1),
--- a/message_ix/model/MESSAGE-MACRO_run.gms
+++ b/message_ix/model/MESSAGE-MACRO_run.gms
@@ -162,7 +162,7 @@ MESSAGE_LP.solvelink = 1 ;
 
 *----------------------------------------------------------------------------------------------------------------------*
 * update total energy system costs by node and time with information from latest MESSAGE run
-total_cost(node_macro, year) = COST_NODAL.L(node_macro, year) / 1000 ;
+total_cost(node_macro, year) = COST_NODAL_NET.L(node_macro, year) / 1000 ;
 trade_cost_detail(node, commodity, year) = import_cost(node, commodity, year) - export_cost(node, commodity, year) ;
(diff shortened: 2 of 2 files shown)


message_ix/model/MACRO/macro_core.gms
```diff
@@ -160,11 +160,20 @@ SUM(node_active,
...
+* economy and the energy system is modeled seperately in MESSAGE. In addition, MACRO does not include the trade of normal goods. Therefore, the revenues from fuel or emission certificate
+* trade should be sepertely added to the total regional income. This is assured by including the trade balance in the :math:`{EC}_{n, y}`. As described in :cite:`manne_buying_1992`,
+* the first-order optimality conditions lead to the Ramsey rule for the optimal allocation of investment and consumption over time. 
+
+* The GDP is reported as the sum of investment, consumption and energy costs without the trade balance. Positive contribution of energy costs to GDP would be counter balanced by the
+* decrease of consumption and investment in the certain year as well as by the reduction in capacity to invest and grow the economy for later years. 
+* 
+* .. math:: Y_{n, y} = C_{n, y} + I_{n, y} + {EC}_{n, y} \qquad \forall{n, y}
+* .. math:: EC_{n, y} = COST NODAL_{n, y} + import cost_{r, y} - export cost_{n, y} \qquad \forall{n, y}
```
behnam-zakeri, line 175: Two equations introduced here, i.e., EC and GDP are not followed by their respective mathematical formulation. EC comes a some lines later, while GDP is not calculated here.

message_ix/model/MESSAGE-MACRO_run.gms
```diff
@@ -114,6 +114,7 @@ Parameters
     price_diff_rel(*,node,sector,year_all)
 
     report_iteration(iteration,*)
+    trade_cost_detail(node, commodity, year_all)              'net of commodity import costs and commodity export revenues by commodity, node and year'
```
behnam-zakeri, line 117: Perhaps better to describe with the same order of indexes, i.e., "node, commodity, and year"

behnam-zakeri (commented):
Thanks @GamzeUnlu95 for drafting this correction. In my view, this is an enhancement; just a few points before merging:
- I suggested some minor edits in the GAMS files as inline comments
- More importantly, can you develop a test that show this correction. For example, a test that asserts that the calculated GDP should be equal to a certain value?
- I believe after applying these changes, some overnight tests may not pass because the objective value may change. I can't see that at the moment as the PR needs to be rebased and the tests should be run again.
Overall, nice contribution.

message_ix/model/MESSAGE-MACRO_run.gms
```diff
@@ -199,6 +202,10 @@ DISPLAY enestart, eneprice, total_cost ;
 * solve MACRO model                                                                                                    *
 *----------------------------------------------------------------------------------------------------------------------*
 
+* update total energy system costs by node and time with information from latest MESSAGE run
+total_cost(node_macro, year) = COST_NODAL_NET.L(node_macro, year) / 1000 ;
```
behnam-zakeri, line 206: I see that here we introduce a new MESSAGE parameter (`total_cost`) to report an existing variable (`COST_NODAL_NET`) only with a unit conversion. As the units are not explicit in MESSAGE, it is not clear to me the use case of this new parameter? In general, I believe it's better not to introduce too many parameters in the output file if the same information can be seen in another parameter/variable.

GamzeUnlu95: `total_cost ` parameter is not introduced in this code but it is already introduced in `macro_data_load.gms` and used in macro_core.gms as follows:

```
COST_ENERGY(node_active, year) $ (NOT macro_base_period(year))..
EC(node_active, year) =E=
(total_cost(node_active, year)/1000
+ SUM(sector, eneprice(node_active, sector, year) * 1E-3 * (PHYSENE(node_active, sector, year) - enestart(node_active, sector, year)))
+ SUM(sector, eneprice(node_active, sector, year) * 1E-3 / enestart(node_active, sector, year) * (PHYSENE(node_active, sector, year) - enestart(node_active, sector, year)) * (PHYSENE(node_active, sector, year) - enestart(node_active, sector, year))))
;
```
Here the purpose of adding that line was not unit conversion but to update this parameter based on the COST_NODAL_NET in each iteration with MESSAGE.

message_ix/model/MACRO/macro_core.gms
GamzeUnlu95: Thanks this is a good point. Instead of EC, total_cost was meant actually and EC is approximated based on total_cost as in the below formula in Equation COST_ENERGY section of the documentation. So I changed this part of the formulation. 

GDP calculation is also explained in this section since it used to be the same as Y (income) but now there is a change in the calculation. The trade balance is considered in Y (income) and therefore not again in the GDP.

message_ix/model/MESSAGE-MACRO_run.gms
behnam-zakeri: Thanks for the explanation.

gidden:
Hi @GamzeUnlu95 - I understand there are still some requirements left for this to be merged. Would you say it is available to try to use on a trial basis?

gidden:
Hi folks, I wanted to report back after having tried to use this PR in some of my ongoing work.

Observed issue: Running a mitigation scenario with a calibrated scenario causes MESSAGE LP infeasibility within MESSAGE-MACRO iteration.

Suggestion: Before merging this, run a test case on the current ENGAGE set up where MACRO is recalibrated with these changes and confirm that mitigation cases can still converge.

More specifics:

It's hard for me to pin point exactly why this PR causes my failure, but I am combining a number of updates - I start from the new R15 model, apply 1-year time steps for early periods, and calibrate using this MACRO branch. I did not observe any issues while creating baselines or non-mitigation scenarios.

khaeru:
> MESSAGE LP infeasibility within MESSAGE-MACRO iteration

Can you please upload a snippet from the GAMS .lst file to indicate which variables are infeasible?

gidden:
That file has already been overwritten. I can try later to reproduce - can you tell me what to `grep` for?

gidden:
But this is a good opportunity to plug our need for better tooling when dealing with infeasibilities. It would be great to have some first (or even second) class support to say, dump out *why* a model was infeasible at the end of a run that is persistable.

khaeru:
> can you tell me what to `grep` for?

https://github.com/iiasa/message_ix/issues/418#issuecomment-720388779 for example

OFR-IIASA:
I have pulled these changes onto the latest main branch, updated the macro-calibration xlsx but then encountered errors.

```
**** Exec Error at line 544: log: FUNC SINGULAR: x = 0
**** Exec Error at line 544: overflow in * operation (mulop)
```

Has anyone encountered such issues? Was this PR already tested with the calibration? I have added the macro_run.lst file as a .txt file if this helps with debugging.
[MACRO_run.txt](https://github.com/iiasa/message_ix/files/9553986/MACRO_run.txt)

GamzeUnlu95:
> I have pulled these changes onto the latest main branch, updated the macro-calibration xlsx but then encountered errors.
> 
> ```
> **** Exec Error at line 544: log: FUNC SINGULAR: x = 0
> **** Exec Error at line 544: overflow in * operation (mulop)
> ```
> 
> Has anyone encountered such issues? Was this PR already tested with the calibration? I have added the macro_run.lst file as a .txt file if this helps with debugging. [MACRO_run.txt](https://github.com/iiasa/message_ix/files/9553986/MACRO_run.txt)

Thanks for adding the .lst file. I attempted to test this PR with calibration a couple of months ago and at that time there was an issue with the macro-calibration xlsx in the main branch and I could not complete it. I need to find and allocate time soon to test this PR again and finalize it.

GamzeUnlu95:
I tested this PR with the ENGAGE setup by using the model `ENGAGE_SSP2_v4.1.7_T4.5_r3.1_test`. I did not experience any issues with the calibration of the baseline. I ran the mitigation case with 600 budget and was able to get a solution for `EN_NPi2020_600`. However, in Step 3 of `EN_NPi2020_400`, in the third iteration, there is an execution error division by eps as in this [MESSAGE-MACRO_run.log](https://github.com/iiasa/message_ix/files/10959553/MESSAGE-MACRO_run.log) and [MESSAGE-MACRO_run_lst](https://github.com/iiasa/message_ix/files/10959566/MESSAGE-MACRO_run_lst.txt) file. Not sure if this is directly related to the changes that are made in this PR. The line indicated in the error message is the following: 
```
price_diff_rel(iteration,node_macro,sector,year)$( price_init(node_macro,sector,year) ) =
5455      price_diff_abs(iteration,node_macro,sector,year) / price_init(node_macro,sector,year)
```

glatterf42:
At today's message meeting, this PR was suggested to me as a possible fix for #752. As far as I can tell, this PR does not enable usage of user-defined units. However, I rebased this onto current main so if you want to run any tests again to see if this can be merged, this might be a good time, @GamzeUnlu95, @OFR-IIASA, @gidden.

khaeru:
This PR may even predate the PR template, but it is missing one item that appears there: "Update release notes". I've added this to the PR description above.

These notes are especially important when the behaviour and outputs of the core GAMS code change, as in this PR. Users *must* have an opportunity to notice and understand how their model outputs may be affected by a new release.

glatterf42:
As discussed in yesterday's weekly meeting, we hold off on merging these changes because we don't trust our current test suite to adequately verify that the changes won't break anything. As @gidden suggested (please correct me if I misunderstood), we should set up a testing workflow that runs a full emission baseline and a peak budget scenario. Since these runs might take a while, we would ideally keep them optional and either run them as part of nightly tests or enable them on PRs that affect the GAMS code. This would allow us to trust the changes here.

@Jihoon noted that the line currently causing the test failure reported by @GamzeUnlu95 had caused issues before, so was removed locally in his installation without any noticeable negative consequences. No one else in the room knew what this line might be good for, so in a separate PR, we could try to remove it -- and test that with the new test cases proposed above.

gidden:
Thanks @glatterf42. My suggestion is to follow @khaeru's example [here](https://github.com/iiasa/message_data/actions/runs/8746996958/workflow) and combine this with @OFR-IIASA's ENGAGE process script [here](https://github.com/iiasa/message_data/blob/ssp_dev/message_data/projects/engage/__init__.py). 

I would suggest to run 

1. a baseline scenario
2. a full century budget scenario
3. a peak-budget scenario

And compare a few key metrics on the output excel (e.g., co2 emissions, kyoto emissions, gdp, primary energy, final energy).

khaeru:
> My suggestion is to follow @khaeru's example [here](https://github.com/iiasa/message_data/actions/runs/8746996958/workflow) and combine this with @OFR-IIASA's ENGAGE process script [here](https://github.com/iiasa/message_data/blob/ssp_dev/message_data/projects/engage/__init__.py).

While I generally agree with this idea directionally and the basic observation underlying it, to implement these ideas would involve, conservatively, at least 2 person-weeks of work. Either we accept that this PR sits until someone is assigned/freed-up to spend that time, or we find a simpler, manual check (and someone to do it) that can assure that this PR will not change *[expected behaviour]*.

In order to avoid derailing this particular PR with that broader discussion, I'll open a new issue.

Issue #237, opened 2019-10-02
Installation: No module named 'pyam' (and further connected issues)


jkikstra:
Perhaps connected to #215, when installing `message_ix` following https://message.iiasa.ac.at/en/stable/getting_started.html I get the following error when running the tests:

```
(base) C:\Users\kikstra\Documents\GitHub\message_ix>py.test tests
=============================================== test session starts ================================================
platform win32 -- Python 3.7.3, pytest-5.2.0, py-1.8.0, pluggy-0.13.0
message_ix location: C:\Users\kikstra\AppData\Local\Continuum\anaconda3\lib\site-packages\message_ix
ixmp config: {'DB_CONFIG_PATH': None, 'DEFAULT_DBPROPS_FILE': None, 'DEFAULT_LOCAL_DB_PATH': WindowsPath('C:/Users/kikstra/.local/share/ixmp/localdb/default')}
rootdir: C:\Users\kikstra\Documents\GitHub\message_ix
plugins: arraydiff-0.3, doctestplus-0.4.0, openfiles-0.4.0, remotedata-0.3.1
collected 59 items / 1 errors / 58 selected

====================================================== ERRORS ======================================================
_____________________________________ ERROR collecting tests/test_reporting.py _____________________________________
ImportError while importing test module 'C:\Users\kikstra\Documents\GitHub\message_ix\tests\test_reporting.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests\test_reporting.py:15: in <module>
    import pyam
E   ModuleNotFoundError: No module named 'pyam'
================================================= warnings summary =================================================
C:\Users\kikstra\AppData\Local\Continuum\anaconda3\lib\site-packages\_pytest\config\__init__.py:545
  C:\Users\kikstra\AppData\Local\Continuum\anaconda3\lib\site-packages\_pytest\config\__init__.py:545: PytestAssertRewriteWarning: Module already imported so cannot be rewritten: ixmp.testing
    self.import_plugin(import_spec)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=========================================== 1 warnings, 1 error in 0.37s ===========================================
``` 

[...]

PR #500, opened 2021-08-18
Update documentation "Common issues"

LauWien:
This PR updates the documentation "Common issues" section by the _No module named 'pyam'_ error message and solves issue [#237](https://github.com/iiasa/message_ix/issues/237)

## How to review

- Build the documentation and look at the installation page.

## PR checklist

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -224,6 +224,38 @@ To check which JVM will be used by ixmp, run the following in any prompt or term
     $ python -c "import jpype; print(jpype.getDefaultJVMPath())"
 
 
+“No module named 'pyam'”
+------------------------
+
+When running a test in ``message_ix`` an error like this may occur.
+Please don't ``pip install pyam`` directly, this will lead you to an unrelated
+package and additional error messages. The package you want to
+install is ``pyam-iamc``. ``pyam-iamc`` is a dependency of ``message_ix.reporting`` and is
+not automatically installed via ``setup.py`` since the package is listed under `extra requirement`_.
+When you want to run a test which needs this dependency you need to install this package
+additionally. So to resolve this error message and install the correct package,
+please use one of following commands within your environment::
+
+    # If installed using conda
+    $ conda install message_ix[reporting]
+    # or
+    $ conda install pyam-iamc
+
+    # If installed using pip
+    $ pip install message_ix[reporting]
+    # or
+    $ pip install pyam-iamc
+
+If you already installed the ``pyam`` package, please remove it from your environment
+with the following to avoid further issues::
+
+    # If installed using conda
+    $ conda remove pyam
+
+    # If installed using pip
+    $ pip uninstall pyam
+
+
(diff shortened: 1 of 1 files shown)


LauWien:
After creating this PR I realized regarding to the installation one always should `pip install --editable .[docs,reporting,tests,tutorial]`. So I'm not so sure if this error still occurs and therefore if the issue #237 is still up-to-date. And if this addition to the documentation is more confusing. If so I will close the issue and this PR.

INSTALL.rst
```diff
@@ -224,6 +224,38 @@ To check which JVM will be used by ixmp, run the following in any prompt or term
     $ python -c "import jpype; print(jpype.getDefaultJVMPath())"
 
 
+“No module named 'pyam'”
+------------------------
+
+When running a test in ``message_ix`` an error like this may occur.
```
khaeru, line 230: These are all the correct details. For clarity, I'd suggest something like:
```
The package `pyam-iamc <https://pypi.org/project/pyam-iamc/>`_ is one of the "reporting" extra dependencies of :mod:`message_ix`.
These extra dependencies are not installed automatically, but can be installed using::

    # If message_ix is installed using pip
    $ pip install message_ix[reporting]
    # or
    $ pip install pyam-iamc

    # If message_ix is installed using Anaconda
    $ conda install message_ix[reporting]
    # or (see note below)
    $ conda install pyam

Note that this package has the *different* name on conda-forge versus PyPI: `pyam <https://anaconda.org/conda-forge/pyam>`__.

The package listed as `pyam <https://pypi.org/project/pyam/>`__ on PyPI (and not available via Anaconda) is unrelated to :mod:`message_ix`, not compatible with it, and will produce other error messages.
If you installed this package accidentally, remove it using::

    # If installed using pip
    $ pip uninstall pyam
```
The thoughts behind these edits:
- Start with the correct thing to do to resolve the error.
- Don't mention the context of "want to run a test" since there are other possible contexts in which this error could be encountered.
- Then address a common mistake.
- Describe the idea of dependencies without referring to or linking to specific files; these will be inscrutable for most users who run into this error.
- I also put the pip code before the conda code deliberately, because the name (pyam-iamc) aligns. *After* explaining that, the next/separate point is to explain that the name differs on conda-forge vs. on PyPI.
- Remove `conda uninstall pyam` since this would actually remove the desired package, due to the differing names.

Also:
- Per the style guide, text in .rst files is 1 sentence per line.

khaeru (changes requested):
Excellent, thanks for tackling this. See the suggested text edits inline.

khaeru:
Also, I think `conda` lacks (still 😒) support for extra dependencies of python packages, so I don't think `conda install message_ix[reporting]` will actually work. Can you please try and see?

INSTALL.rst
LauWien: Thank you for your time to review this and your input!

khaeru (approved):

Issue #507, opened 2021-09-14
Flaky test test_tutorials.py::test_tutorial[R_austria] on macOS with Python version 3.8


LauWien:
Merging new PRs into the main branch triggers most of the time the flaky test _message_ix/tests/test_tutorials.py::test_tutorial[R_austria]_ e.g. see [here](https://github.com/iiasa/message_ix/runs/3360866578). Since this test on macOS is also covered by Python version 3.9 it is fair to mark the test as xfail.

PR #508, opened 2021-09-15
Mark R_austria tutorial test as xfail on macos-latest-py3.8

LauWien:
This PR marks test _message_ix/tests/test_tutorials.py::test_tutorial[R_austria]_ as xfail due to flakiness. Merging this PR will close issue #507. 

## How to review

## PR checklist

- [x] ~~Add or~~ expand tests; ~~coverage checks both~~ ✅

### DIFF
--- a/message_ix/tests/test_tutorials.py
+++ b/message_ix/tests/test_tutorials.py
@@ -96,15 +96,6 @@ def test_tutorial(nb_path, cell_values, run_args, tmp_path, tmp_env):
 
     If *cell_values* are given, values in the specified cells are tested.
     """
-    # Add the tutorial directory to PYTHONPATH. The tutorials are executed in
-    # `tmp_path`; but they import from a tools.py file in the same directory as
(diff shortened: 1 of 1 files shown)


khaeru:
I pushed a change here. Briefly:
- The `tutorial` variable in this file is a list which is passed to `pytest.mark.parametrize()`.
- Per the Pytest docs ([1](https://docs.pytest.org/en/stable/parametrize.html), [2](https://docs.pytest.org/en/stable/example/parametrize.html#set-marks-or-test-id-for-individual-parametrized-test)), this list already used `pytest.param(…, marks=…)` for some of the cases, to apply marks to individual tests, e.g. the R tutorials.
- So we can (a) add a new mark object, stored as a variable (`mark1`), and then (b) add that to the existing `marks=` argument for the test(s) to which it applies.

This avoids having the logic about which tests to run/outcomes to expect in the body of the test itself.

khaeru (approved):

PR #513, opened 2021-09-28
Change precision of GAMS check for parameter "duration_time"

behnam-zakeri:
This PR changes the way the GAMS formulation checks the sum of `duration_time` of sub-annual time slices (hereafter "time slices"). At the moment, GAMS checks that the sum of `duration_time` at a temporal level is equal to 1. However, as the precision differs between GAMS and python values, in some cases a very small, negligible difference in the scale of 1e-12 triggers an error.  This PR changes the way this check, is done in GAMS, i.e., from equality to 1 to check if the difference is smaller than 1e-12.

## How to review

For testing this PR a model with time slices should be used, preferably with many time slices with different `duration_time`. The review can use the tests designed in this PR, change the number of time slices per year and see if the check for the sum of `duration_time` in GAMS works.

- [x] Update release notes.

### DIFF
--- a/message_ix/tests/test_feature_duration_time.py
+++ b/message_ix/tests/test_feature_duration_time.py
@@ -1,0 +1,249 @@
+# -*- coding: utf-8 -*-
+"""
+This test ensures that the parameter "duration_time" is specified correctly,
+and that the GAMS formulation checks pass.
+Several possible combinations of temporal levels, e.g., year, seasons,
+months; and different number of time slices at each level are tested.
+
+"""
+
+from itertools import product
+
+from message_ix import Scenario
+
+
+# A function for generating a simple model with sub-annual time slices
+def model_generator(
+    test_mp,
+    comment,
+    tec_time,
+    demand_time,
+    time_steps,
+    com_dict,
+    yr=2020,
+):
+    """
+
+    Generates a simple model with two technologies, and a flexible number of
+    time slices.
+
+    Parameters
+    ----------
+    test_mp : ixmp.Platform()
+    comment : string
+        Annotation for saving different scenarios and comparing their results.
+    tec_time : dict
+        A dictionary for mapping of a technology to its input/output temporal levels.
+        # Format: {<technology>: [<input temporal level>, <output temporal level>]}
+        (e.g., {"gas_ppl": ["year", "season"]})
+    demand_time : dict
+        A dictionary for mapping of total "demand" specified at a temporal level.
+        e.g., demand_time = {"season": 100}, means the sum of demand in all "time"
+        slices at the level of "season" must be 100.
+    time_steps : list of lists
+        Information about each time slice, packed in a tuple with four elements,
+        including: "temporal_lvl", number of time slices, and parent time slice.
+        e.g., time_steps = [["season", 4, "year"], ["day", 24, "season"]]
+    com_dict : dict
+        A dictionary for specifying "input" and "output" commodities.
+        (e.g., com_dict = {"gas_ppl": {"input": "fuel", "output": "electr"}})
+    yr : int, optional
+        Model year. The default is 2020.
+
+
+    """
+
+    # Building an empty scenario
+    scen = Scenario(test_mp, "test_duration_time", comment, version="new")
+
+    # Adding required sets
+    scen.add_set("node", "fairyland")
+    for c in com_dict.values():
+        scen.add_set("commodity", [x for x in list(c.values()) if x])
+
+    scen.add_set("level", "final")
+    scen.add_set("year", yr)
+    scen.add_set("type_year", yr)
+    scen.add_set("technology", list(tec_time.keys()))
+    scen.add_set("mode", "standard")
+
+    # Adding "time" related info to the model: "lvl_temporal", "time",
+    # "map_temporal_hierarchy", and "duration_time"
+    map_time = {}
+    for [tmp_lvl, number, parent] in time_steps:
+        scen.add_set("lvl_temporal", tmp_lvl)
+        if parent == "year":
+            times = [tmp_lvl[0] + "-" + str(x + 1) for x in range(number)]
+        else:
+            times = [
+                p + "_" + tmp_lvl[0] + "-" + str(x + 1)
+                for (p, x) in product(map_time[parent], range(number))
+            ]
+
+        map_time[tmp_lvl] = times
+        scen.add_set("time", times)
+
+        # Adding "map_temporal_hierarchy" and "duration_time"
+        for h in times:
+            if parent == "year":
+                p = "year"
+            else:
+                p = h.split("_" + tmp_lvl[0])[0]
+            # Temporal hierarchy (order: tenporal level, time, parent time)
+            scen.add_set("map_temporal_hierarchy", [tmp_lvl, h, p])
+
+            # Duration time is relative to the duration of the parent temporal level
+            dur_parent = float(scen.par("duration_time", {"time": p})["value"])
+            scen.add_par("duration_time", [h], dur_parent / number, "-")
+
+    # Adding "demand" at a temporal level (total demand divided by the number of
+    # time slices in that temporal level)
+    for tmp_lvl, value in demand_time.items():
+        times = scen.set("map_temporal_hierarchy", {"lvl_temporal": tmp_lvl})["time"]
+        for h in times:
+            scen.add_par(
+                "demand",
+                ["fairyland", "electr", "final", yr, h],
+                value / len(times),
+                "GWa",
+            )
+
+    # Adding "input" and "output" parameters of technologies
+    for tec, [tmp_lvl_in, tmp_lvl_out] in tec_time.items():
+        times_in = scen.set("map_temporal_hierarchy", {"lvl_temporal": tmp_lvl_in})[
+            "time"
+        ]
+        times_out = scen.set("map_temporal_hierarchy", {"lvl_temporal": tmp_lvl_out})[
+            "time"
+        ]
+        # If technology is linking two different temporal levels
+        if tmp_lvl_in != tmp_lvl_out:
+            time_pairs = product(times_in, times_out)
+        else:
+            time_pairs = zip(times_in, times_out)
+
+        # Configuring data for "time_origin" and "time" in "input"
+        for (h_in, h_act) in time_pairs:
+            # "input"
+            inp = com_dict[tec]["input"]
+            if inp:
+                inp_spec = [yr, yr, "standard", "fairyland", inp, "final", h_act, h_in]
+                scen.add_par("input", ["fairyland", tec] + inp_spec, 1, "-")
+        # "output"
+        for h in times_out:
+            out = com_dict[tec]["output"]
+            out_spec = [yr, yr, "standard", "fairyland", out, "final", h, h]
+            scen.add_par("output", ["fairyland", tec] + out_spec, 1, "-")
+
+    # Committing
+    scen.commit("scenario was set up.")
+
+    # Testing if the model solves in GAMS
+    scen.solve(case=comment)
+
+    # Testing if sum of duration_time is almost 1
+    for tmp_lvl in scen.set("lvl_temporal"):
+        times = scen.set("map_temporal_hierarchy", {"lvl_temporal": tmp_lvl})[
+            "time"
+        ].to_list()
+        print(sum(scen.par("duration_time", {"time": times})["value"]))
+        assert (
+            abs(sum(scen.par("duration_time", {"time": times})["value"]) - 1.0) < 1e-9
+        )
+    return scen
+
+
+# Tests for "duration_time" of different number of time slices, at different
+# temporal levels ("lvl_temporal")
+# In these tests "demand" is defined in different time slices and one power plant
+# to meet demand, which receives fuel from a supply technology.
+
+# 1) Testing one temporal level ("season") and different number of time slices
+def test_season(test_mp, n_time=[4, 12, 50, 122, 360, 2240]):
+    comment = "season"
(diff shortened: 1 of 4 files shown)


khaeru:
Per a query from @LauWien in Slack:
- Some of the tests introduced by this PR are larger than the maximum LP size that GAMS permits without a license.
- We do use, in the pytest.yaml workflow, a stored GitHub secret to allow solving larger models. However, these secrets are not available in PRs from a fork, like this one (see [GitHub Help](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow)). That is not avoidable; many PRs will be from forks.
- The general solution here is to **selectively disable** these tests. (It's good that tests were written, and no point to throw them away!)
- To do this, use pytest markers and marked parameters, as shown in the documentation: https://docs.pytest.org/en/stable/example/parametrize.html (Another approach is to comment out the specific tests or parameters, but this is relatively crude.)
- We already do this in some places. An example: https://github.com/iiasa/message_ix/blob/df50ef9dac028777fda8a953d1e505592e6d98fe/message_ix/tests/test_tutorials.py#L12-L15
- I would suggest to check one of [these](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables), e.g. `GITHUB_WORKFLOW`, roughly like:
  ```py
  @pytest.mark.skipif(
      condition="GITHUB_WORKFLOW" in os.environ,
      reason="Exceeds maximum L{ size for GAMS without license."
  )
  def test_year_season_day(test_mp):
      …
  ```
- Please _only_ do this for the specific tests, or cases, that are too large.
- This introduces a restriction: no third-party developer will be able to run the full test suite, unless they themselves have a GAMS license, but (IMO) this is acceptable, since:
  - (a) that doesn't happen often, and
  - (b) we will anyway need to solve a larger problem when GAMS deprecates/removes the older versions and only offers downloads for versions >30.0 that have a maximum LP size of 0 without a license. We may solve it via iiasa/ixmp#119.

behnam-zakeri:
Thanks @LauWien and @khaeru for the quick review and detailed instructions. I applied the suggestion for skipping the tests that are large, and it made most of the checks pass. But now the coverage check is not passing as the skipped lines are not being tested.  If this is a real issue for the coverage rate, we can reduce the size of the tests so that we avoid skipping them. I believe the reduced size tests would still serve the purpose.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+    scen.solve(case=comment)
+
+    # Testing if sum of duration_time is almost 1
+    for tmp_lvl in scen.set("lvl_temporal"):
+        times = scen.set("map_temporal_hierarchy", {"lvl_temporal": tmp_lvl})[
+            "time"
+        ].to_list()
+        print(sum(scen.par("duration_time", {"time": times})["value"]))
+        assert (
+            abs(sum(scen.par("duration_time", {"time": times})["value"]) - 1.0) < 1e-12
```
LauWien, line 154: 👍

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+        )
+    return scen
+
+
+# Tests for "duration_time" of different number of time slices, at different
+# temporal levels ("lvl_temporal")
+# In these tests "demand" is defined in different time slices and one power plant
+# to meet demand, which receives fuel from a supply technology.
+
+# 1) Testing one temporal level ("season") and different number of time slices
```
LauWien, line 164: Maybe we don't need to number the tests.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+def test_season(test_mp, n_time=[4, 12, 50]):
+    comment = "season"
+    com_dict = {"power-plant": {"input": [], "output": "electr"}}
+    # Dictionary of technology and input and output temporal levels
+    tec_time = {"power-plant": ["season", "season"]}
+    # Total demand in a temporal level
+    demand_time = {"season": 100}
+    # List of info for time slices: [temporal level, number, parent level]
+    for t in n_time:
+        time_steps = [["season", t, "year"]]
```
LauWien, line 174: See comments above about the tuple. Also in the following tests.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+        model_generator(test_mp, comment, tec_time, demand_time, time_steps, com_dict)
+
+
+# 3) Testing two temporal levels with one technology
+def test_season_day(test_mp):
+    comment = "4-season_24-days"
+    com_dict = {
+        "power-plant": {"input": [], "output": "electr"},
+    }
+    # Dictionary of technology and temporal level
```
LauWien, line 207: I think we can delete all the comments within the tests since `tec_time`, `demand_time`, `time_steps` are already explained in the docstring in the beginning.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+    tec_time = {"power-plant": ["day", "day"]}
+    # Total demand in a temporal level
+    demand_time = {"day": 100}
+    # List of info for time slices: [temporal level, number, parent level]
+    time_steps = [["season", 4, "year"], ["day", 24, "season"]]
+
+    # Check the model solves without error and sumof duration times = 1
```
LauWien, line 214: See comment above.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+    model_generator(test_mp, comment, tec_time, demand_time, time_steps, com_dict)
+
+
+# 4) Testing 24 days linked to 4 seasons, with two technologies
```
LauWien, line 218:
```suggestion
# 4) Testing 24 days linked to four seasons, with two technologies
```

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+    # Total demand in a temporal level
+    demand_time = {"day": 100}
+    # List of info for time slices: [temporal level, number, parent level]
+    time_steps = [["season", 4, "year"], ["day", 24, "season"]]
+
+    # Check the model solves without error and sumof duration times = 1
+    model_generator(test_mp, comment, tec_time, demand_time, time_steps, com_dict)
+
+
+# 5) Testing 60 days linked to 4 seasons linked to year, with three technology
```
LauWien, line 236:
```suggestion
# 5) Testing 60 days linked to four seasons linked to year, with three technology
```

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+        "power-plant": ["season", "day"],
+        "fuel-transport": ["year", "season"],
+        "fuel-supply": ["year", "year"],
+    }
+    # Total demand in a temporal level
+    demand_time = {"day": 100}
+    # List of info for time slices: [temporal level, number, parent level]
+    time_steps = [["season", 4, "year"], ["day", 30, "season"]]
+
+    # Check the model solves without error and sumof duration times = 1
```
LauWien, line 255: See comment above.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+        "fuel-supply": {"input": [], "output": "fuel"},
+    }
+    # Dictionary of technology and temporal level
+    tec_time = {"power-plant": ["year", "season"], "fuel-supply": ["year", "year"]}
+    # Total demand in a temporal level
+    demand_time = {"season": 100}
+    # List of info for time slices: [temporal level, number, parent level]
+    for t in n_time:
+        time_steps = [["season", t, "year"]]
+        # Check the model solves without error and sumof duration times = 1
```
LauWien, line 197:
```suggestion
    # Check the model solves without error and the sum of duration_time = 1
```
Or
`    # Check the model solves without error and the sum of duration time equals one`

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+        "fuel-supply": {"input": [], "output": "fuel"},
+    }
+    # Dictionary of technology and temporal levels
+    tec_time = {"power-plant": ["season", "day"], "fuel-supply": ["season", "season"]}
+    # Total demand in a temporal level
+    demand_time = {"day": 100}
+    # List of info for time slices: [temporal level, number, parent level]
+    time_steps = [["season", 4, "year"], ["day", 24, "season"]]
+
+    # Check the model solves without error and sumof duration times = 1
```
LauWien, line 232: See comment above.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,288 @@
...
+    # Total demand in a temporal level
+    demand_time = {"hour": 100}
+    # List of info for time slices: [temporal level, number, parent level]
+    time_steps = [
+        ["season", n_season, "year"],
+        ["day", n_day, "season"],
+        ["hour", n_hour, "day"],
+    ]
+
+    # Check the model solves without error and sumof duration times = 1
```
LauWien, line 287: See comment above.

LauWien (changes requested):
Thank you! Just some suggestions :)

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,281 @@
...
+    time slices.
+
+    Parameters
+    ----------
+    test_mp : ixmp.Platform()
+    comment : string
+        Annotation for saving different scenarios and comparing their results.
+    tec_time : dict
+        A dictionary for mapping of a technology to its input/output temporal levels.
+        # Format: {<technology>: [<input temporal level>, <output temporal level>]}
```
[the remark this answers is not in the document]
behnam-zakeri: Thanks for spotting. It was leftover from a copy-pasting.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,281 @@
...
+        (e.g., {"gas_ppl": ["year", "season"]})
```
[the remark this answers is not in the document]
behnam-zakeri: Thanks. I removed them, as we have them later specified in the examples.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,281 @@
...
+    demand_time : dict
+        A dictionary for mapping of total "demand" specified at a temporal level.
+        e.g., demand_time = {"season": 100}, means the sum of demand in all "time"
+        slices at the level of "season" must be 100.
+    time_steps : list of lists
```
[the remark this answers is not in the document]
behnam-zakeri: Thanks, I updated this to tuples for all instances.

khaeru:
> But now the coverage check is not passing as the skipped lines are not being tested. If this is a real issue for the coverage rate, we can reduce the size of the tests so that we avoid skipping them. I believe the reduced size tests would still serve the purpose.

Sometimes this is slow to update, i.e. the comment from Codecov and the check results will only show green once all the different jobs for different OS's and Python versions have completed. At the moment, it looks fine.

In the future/general, you can use [`# pragma: no cover`](https://coverage.readthedocs.io/en/stable/excluding.html) comment (click for docs), e.g. on the line where a function is defined, to exclude it from code coverage calculations. See existing examples in the code.

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,255 @@
...
+    ----------
+    comment : string
+        Annotation for saving different scenarios and comparing their results.
+    tec_time : dict
+        A dictionary for mapping a technology to its input/output temporal levels.
+    demand_time : dict
+        A dictionary for mapping the total "demand" specified at a temporal level.
+        e.g., {"season": 100}, means the sum of demand in all "time"
+        slices at the level of "season" must be 100.
+    time_steps : list of lists
```
LauWien, line 40:
```suggestion
    time_steps : list of tuples
```

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,255 @@
...
+# Testing one temporal level ("season") and different number of time slices
+def test_season(test_mp, n_time=[4, 12, 50]):
+    comment = "season"
+    com_dict = {"power-plant": {"input": [], "output": "electr"}}
+    tec_time = {"power-plant": ["season", "season"]}
+    demand_time = {"season": 100}
+
+    for t in n_time:
+        time_steps = [("season", t, "year")]
+        # Check the model solves without error and sum of duration times = 1
```
LauWien, line 163:
```suggestion
        # Check the model solves without error and sum of "duration_time" = 1
```

LauWien (approved):
Only three suggestions. Otherwise good to go, I would say. And the test ran through :)

message_ix/tests/test_feature_duration_time.py
```diff
@@ -0,0 +1,255 @@
...
+    Parameters
+    ----------
+    comment : string
+        Annotation for saving different scenarios and comparing their results.
+    tec_time : dict
+        A dictionary for mapping a technology to its input/output temporal levels.
+    demand_time : dict
+        A dictionary for mapping the total "demand" specified at a temporal level.
+        e.g., {"season": 100}, means the sum of demand in all "time"
+        slices at the level of "season" must be 100.
```
LauWien, line 39: Should we remove also this example or is it fine to leave it here?

behnam-zakeri: Yes, we can remove this. I resolved the remaining issues. Thanks, @LauWien.

PR #514, opened 2021-09-29
draft of regional emission pool implementation

volker-krey:
- added new  parameter emission_sink_rate
- added new equation EMISSION_POOL with variable EMISS_POOL
- possibility to limit size of pools via constraint currently not included
- note that I did not run the code, so it has not been tested for correct syntax

### DIFF
--- a/message_ix/model/MESSAGE/data_load.gms
+++ b/message_ix/model/MESSAGE/data_load.gms
@@ -51,7 +51,7 @@ addon_conversion, addon_up, addon_lo
 rating_bin, reliability_factor, peak_load_factor, flexibility_factor
 renewable_capacity_factor, renewable_potential
 * emission factors, bounds and taxes on emissions (including mapping sets)
-historical_emission, emission_factor, emission_scaling, is_bound_emission, bound_emission, tax_emission,
+historical_emission, emission_factor, emission_scaling, is_bound_emission, bound_emission, tax_emission, emission_sink_rate,
 * historical values of new capacity investment, activity and extraction
 historical_new_capacity, historical_activity, historical_extraction
 * parameters for land-use model emulator
--- a/message_ix/model/MESSAGE/model_core.gms
+++ b/message_ix/model/MESSAGE/model_core.gms
@@ -118,6 +118,8 @@ Variables
     REL(relation,node,year_all)                  auxiliary variable for left-hand side of user-defined relations
 * change in the content of storage device
     STORAGE_CHARGE(node,tec,level,commodity,year_all,time)    charging of storage in each timestep (negative for discharge)
+* regional emission pool
+    EMISS_POOL(node,type_emission,type_tec,year_all)          nodal-regional-global emission pool size
 ;
 
 ***
@@ -272,6 +274,7 @@ Equations
     ACTIVITY_SOFT_CONSTRAINT_LO     bound on relaxation of the dynamic constraint on market penetration (lower bound)
     EMISSION_EQUIVALENCE            auxiliary equation to simplify the notation of emissions
     EMISSION_CONSTRAINT             nodal-regional-global constraints on emissions (by category)
+    EMISSION_POOL                   nodal-regional-global emission pool with carbon sink rate
     LAND_CONSTRAINT                 constraint on total land use (linear combination of land scenarios adds up to 1)
     DYNAMIC_LAND_SCEN_CONSTRAINT_UP dynamic constraint on land scenario change (upper bound)
     DYNAMIC_LAND_SCEN_CONSTRAINT_LO dynamic constraint on land scenario change (lower bound)
@@ -1579,12 +1582,12 @@ NEW_CAPACITY_CONSTRAINT_UP(node,inv_tec,year)$( map_tec(node,inv_tec,year)
 *   .. math::
 *      CAP\_NEW\_UP_{n,t,y} \leq \sum_{y-1} CAP\_NEW_{n^L,t,y-1} & \text{if } y \neq 'first\_period' \\
 *                                + \sum_{y-1} historical\_new\_capacity_{n^L,t,y-1} & \text{if } y = 'first\_period' \\
-*                           \quad \forall \ t \ \in \ T^{INV}   
+*                           \quad \forall \ t \ \in \ T^{INV}
 *
 ***
 NEW_CAPACITY_SOFT_CONSTRAINT_UP(node,inv_tec,year)$( soft_new_capacity_up(node,inv_tec,year) )..
     CAP_NEW_UP(node,inv_tec,year) =L=
-        SUM(year2$( seq_period(year2,year) ), 
+        SUM(year2$( seq_period(year2,year) ),
             CAP_NEW(node,inv_tec,year2)) $ (NOT first_period(year))
       + SUM(year_all2$( seq_period(year_all2,year) ),
             historical_new_capacity(node,inv_tec,year_all2)) $ first_period(year)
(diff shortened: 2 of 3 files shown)


OFR-IIASA:
@volker-krey - I have added a checklist of tasks to be completed. The initial setup now works and the calculations for the tutorial seem correct (see attached calculation in xlsx: [Westeros_Calculation_check.xlsx](https://github.com/iiasa/message_ix/files/7259857/Westeros_Calculation_check.xlsx)).
@khaeru we need some advice on how to proceed with two specific items.
1. we have a new set "is_emission_sink". These are normally calculated in the backend as far as I know. Currently, we manually create this in the tutorial, but we should resolve this for the final PR. 
2. we also want to avoid having to pass the var "EMISS_POOL" in the solve command to be able to get the results. This is also I believe done in the backend?

khaeru:
>  @khaeru we need some advice on how to proceed with two specific items.

Please read this file https://github.com/iiasa/message_ix/blob/1da31367efc405f00b50190d3338fe9ed01ba124/message_ix/models.py#L75-L77

You are correct that some things like this were originally done in the Java/ixmp_source/JDBCBackend code, but we are trying to move away from that. This is in order to have a cleaner separation between the layers of the stack, which means that `ixmp` (including any Backend, Java or not) should not have any hard-coded names of sets, variables, etc.; it should be agnostic about them. The `message_ix` package is the proper place to name and define particular sets, parameters, etc. that are particular to MESSAGE (the word "scheme" is used in some places); `MESSAGE_ITEMS` is where these are stored.

See for example how this was done in #190: https://github.com/iiasa/message_ix/pull/190/files#diff-62f44bbc1d152875567f99d639addb547727233454fe1958278c45770d6a2912

OFR-IIASA:
> 
> 
> > @khaeru we need some advice on how to proceed with two specific items.
> 
> Please read this file
> 
> https://github.com/iiasa/message_ix/blob/1da31367efc405f00b50190d3338fe9ed01ba124/message_ix/models.py#L75-L77
> 
> You are correct that some things like this were originally done in the Java/ixmp_source/JDBCBackend code, but we are trying to move away from that. This is in order to have a cleaner separation between the layers of the stack, which means that `ixmp` (including any Backend, Java or not) should not have any hard-coded names of sets, variables, etc.; it should be agnostic about them. The `message_ix` package is the proper place to name and define particular sets, parameters, etc. that are particular to MESSAGE (the word "scheme" is used in some places); `MESSAGE_ITEMS` is where these are stored.
> 
> See for example how this was done in #190: https://github.com/iiasa/message_ix/pull/190/files#diff-62f44bbc1d152875567f99d639addb547727233454fe1958278c45770d6a2912

In the PR, we have already updated `MESSAGE_ITEMS`. The issue as far as I understand though is that "is_*" needs to be derived dynamically when solving based on the parameters in the scenario as this is a helper table to avoid 0 values being misinterpreted by GAMS.

khaeru:
> The issue as far as I understand though is that "is_*" needs to be derived dynamically when solving based on the parameters in the scenario as this is a helper table to avoid 0 values being misinterpreted by GAMS.

For these "generated", "derived", or "mapping" sets, there are at least three approaches:
1. In Java/ixmp_source/JDBCBackend, during the step of writing out the Scenario data to a GDX file. This approach is 🚫 because it conflates (`ixmp` backend or I/O operation of writing Scenario content to GDX) with (`message_ix` model formulation and associated logic, i.e. the meaning and significance of that content—something about which `ixmp` must be agnostic, per previous comment). I mention it here because it is still in use at the moment, but we should be aiming to eliminate this code, and should not create more of it.
2. In the GAMS code directly. I understand this is already done in some cases. As @behnam-zakeri notes at #516, however, sometimes the mapping set is required _before_ the corresponding parameter data, in order to ensure it is read correctly (with GAMS `EPS = 0` versus missing values). One advantage here is that people using the GAMS code _without_ the Python API have an easier time, because they don't need to populate the sets manually.
3. In Python, in `message_ix.models`.

IMHO (3) is preferable for a few reasons:
- More developers are able to understand and improve Python than GAMS or Java.
- It unites or consolidates the MESSAGE "scheme" and consistency checks/logic specific to the scheme in the same module/class.
- It can (theoretically) be operated without writing to GDX or attempting to solve a Scenario, which means atomic operations are better separated.

khaeru:
@OFR-IIASA @behnam-zakeri I'm happy to jump in here at some point to demo how the code for the dynamic derivation of the `is_*` sets should look, per my suggestion above. Please let me know when's a good moment.

OFR-IIASA:
> 
> 
> @OFR-IIASA @behnam-zakeri I'm happy to jump in here at some point to demo how the code for the dynamic derivation of the `is_*` sets should look, per my suggestion above. Please let me know when's a good moment.

@khaeru many thanks for offering this. I think we are at a stage now where your suggestion makes sense to implement. We have completed the updates to the core formulation and are now at the stage where we want to cleanup the PR in preparation for merging.

khaeru:
> I think we are at a stage now where your suggestion makes sense to implement.

Thanks for the heads-up! Added to my queue.

khaeru:
> Added to my queue.

Now done.
- One of your added tests (`message_ix/tests/test_emission_pool.py::test_bound_emission_pool`) was failing when I checked out the branch (values not as expected), and it is still failing in the same way after my changes. The other tests pass.
- Please have a look at how I adjusted the tests; I did not touch the notebooks.
- At some point, the branch should be rebased on latest `main`, and then you should each do `git pull --rebase`. I did not do this for you.
- Some of these changes should eventually go upstream to `ixmp`, but we can do that later. What I added should be forward-compatible.

OFR-IIASA:
Thanks @khaeru for adding the code.

> * One of your added tests (`message_ix/tests/test_emission_pool.py::test_bound_emission_pool`) was failing when I checked out the branch (values not as expected), and it is still failing in the same way after my changes. The other tests pass.

I have adjusted the tests accordingly.

> * Please have a look at how I adjusted the tests; I did not touch the notebooks.

Thanks, I have adjusted the tutorial notebook.

> * At some point, the branch should be rebased on latest `main`, and then you should each do `git pull --rebase`. I did not do this for you.

I have rebased the branch.

Lastly, I have also added a test to check the removal of the parameter `bound_emission_pool`. This also checks the removal of the corresponding `is_*` set. The test currently fails on the `is_*`, but we will need to correct this otherwise the problem becomes infeasible.

message_ix/models.py
```diff
@@ -378,21 +419,44 @@ class MESSAGE(GAMSModel):
...
+            # Existing and expected contents
+            existing = scenario.set(set_name)
+            expected = scenario.par(par_name).drop(columns=["value", "unit"])
+
+            if existing.equals(expected):
+                continue  # Contents are as expected; do nothing
+
+            # Not consistent; empty and then re-populate the set
+            with scenario.transact(f"Enforce consistency of {set_name} and {par_name}"):
+                scenario.remove_set(set_name, existing)
```
khaeru, line 443: @OFR-IIASA this line is supposed to entirely clear the set before it is populated by the next line. You could perhaps check whether this is having the intended effect.

OFR-IIASA: @khaeru I have added a two tests to see if this feature works correctly. 
- Test 1. adds the parameter `bound_emission_pool` for 2 periods, solves and checks `is_bound_emission_pool`.
   The bound is then extended across another period, solved and checks `is_bound_emission_pool`
- Test 2. adds the parameter `bound_emission_pool` for 3 periods, solves and checks `is_bound_emission_pool`.
   The `bound_emission_pool` is removed.
   The `bound_emission_pool` is then re-added, but only for two periods, solved and checks `is_bound_emission_pool`.
Both the tests work and confirm the working of the code above.

OFR-IIASA: @khaeru The above code has tests, but for the functionality for which the code, in my understanding was designed. The design was to ensure that the "is_*" is consistent when the corresponding parameters is being _modified_; the code does not cover the _removal_ of the corresponding parameter. Further, this code is only executed for `model="MESSAGE"`, an not when passing the argument `model="MESSAGE-MACRO"` as part of the solve command.

khaeru: > The design was to ensure that the "is_*" is consistent when the corresponding parameters is being modified; the code does not cover the removal of the corresponding parameter. The removal is test in a test test_bound_emission_pool_removal(), which fails.

Thanks for directing to this specific test. To be clear:
- The added `enforce()` method is called _at the moment the model is run/solved_.
- In the final lines of that test, there is a call to `.remove_par("bound_emission_pool", …)`, then a call to `.commit()` but—unlike the other tests—there is *no call to `.solve()`* before `.set("is_bound_emission_pool")` is checked.
- So `enforce()` is never invoked, and the result is as expected.
- I suspect that if you called `.solve()` before checking the set, then the test would pass.

More detail: a possible choices would be to have `enforce()` run (a) on every call to `.commit()`, or even (b) on every call to `.add_par()`. The current choice is deliberate: it mirrors the behaviour of the Java code underlying JDBCBackend, which only creates these mapping/etc. sets at the moment the data is written out to GDX. That Java code must someday be ported into message_ix. `enforce()` is now the destination for that code and runs at the same point.

I thought this was better than introducing 1+ additional points at which data is massaged automatically; it would be complex and confusing additional work to think through and plan for possible interactions of these automatic changes. We unfortunately do not have the bandwidth for that at the moment.

> Further, this code is only executed for model="MESSAGE", not when passing the argument model="MESSAGE-MACRO" as part of the solve command.

This I can check, and will do after you try adjusting the test.

OFR-IIASA: I finally got around to sorting out the tests. All tests pass; now the issue would be to ensure that we have this working when the calling `solve(model="MESSAGE-MACRO")`. Unfortunately I dont think we have a simple test-case which can be adapted and run with macro.

khaeru: I can try to look at this next week, after we've wrapped up the current release.

Issue #515, opened 2021-10-01
Zero values for capacity_factor are changed to 1 in GAMS


behnam-zakeri:
When assigning parameter `capacity_factor` in GAMS, the missing values are by default considered to be equal to 1 (see below equation):
https://github.com/iiasa/message_ix/blob/2cf0df3a3257d74fd423ff50818f33265c6cd831/message_ix/model/MESSAGE/data_load.gms#L141

While this is fine for many cases, it can cause unexpected model behavior when a user defines `capacity_factor` equal to zero for a technology in a certain model period or sub-annual time slice, e.g., for solar PV at night. As zero values for `capacity_factor` in GDX are not loaded in GAMS, they are interpreted as missing values in the formulation, and are set to 1.
A quick workaround can be that the user defines very small values for `capacity_factor` instead of zero, and bypass the issue. However, a longer term solution is needed, or at least this should be warned in one way.

LauWien:
#409 is related.

khaeru:
Per today's MESSAGE meeting:
- @behnam-zakeri will come up with a small/minimal example of how to reproduce this, written as a new test on a new branch/PR.
- I will help investigate fixes to get the expected behaviour.

behnam-zakeri:
Thanks, @khaeru and @LauWien for tackling this issue. I made a test for the parameter "capacity_factor" as attached. The test has three parts and can be used beyond the issue raised here.
The first two tests check if "capacity_factor" behaves as expected in a model with time slices, one with different capacity factor per time slice, and another with different duration of time slices.
The last test (named "test_capacity_factor_zero") checks the issue raised here, i.e., zero capacity factors. At the moment, this test fails, as the model shows a technology active ("ACT" > 0)  in a time slice that the capacity factor of that technology is set to zero. The expected outcome is that "ACT" must be zero when "capacity_factor" is zero.
A side note: this test is written by manipulating the existing test "test_feature_temporal_level". This can be used to define a bare model with time slices, and avoid the part that generates the model in this code.
```"""
This test is designed to check if parameter "capacity_factor" works for models
with sub-annual time slices (index of "time" in MESSAGEix) as expected or not.

"""

from itertools import product

from message_ix import Scenario


def add_cap_par(scen, years, tec, data={"inv_cost": 0.1, "technical_lifetime": 5}):
    """
    Adding required parameters for representing "capacity" in a MESSAGEix model.

    Parameters
    ----------
    scen : Class
        message_ix.Scenario.
[...]

Issue #559, opened 2022-02-09
ValueError in .add_year.interpolate_2d() with pandas 1.4.0


LauWien:
The tests `test_add_year.py::test_add_year_cli` and `test_add_year.py::test_add_year` start to fail seemingly randomly in various PRs on different OS e.g. [here](https://github.com/iiasa/message_ix/runs/5125036229?check_suite_focus=true). Looking deeper, failing test are using `pandas==1.4.0`, passing tests are using e.g. `pandas==1.1.5`. With upgrading pandas to 1.4.0 locally these errors can be replicated, while with version 1.1.5 the test succeed.

khaeru:
The fundamental cause here is code that's not compatible with current pandas; the failing CI is only a symptom. The offending lines are: https://github.com/iiasa/message_ix/blob/1df3619a5223194171cc55197f65858ba62f16f2/message_ix/tools/add_year/__init__.py#L849-L851
```
ValueError: cannot set using a list-like indexer with a different length than the value
```

NB #494 was started to simplify these very long methods that mostly duplicate upstream features and are also quite slow, but I lacked clear tests for the expected behaviour for some combinations of arguments.

LauWien:
> The fundamental cause here is code that's not compatible with current pandas;

Yes, was already checking where and how to adapt the code. Thank you for linking the PR!

khaeru:
https://github.com/iiasa/message_ix/pull/557#issuecomment-1048604890
> It looks like there were several regressions related to `pd.DataFrame.loc` fixed in pandas 1.4.1: https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.4.1.html#fixed-regressions.

i.e. perhaps this is only with pandas 1.4.0 precisely.

PR #561, opened 2022-02-16
Resolve the issue of zero capacity factor

behnam-zakeri:
This PR adds tests for parameter "capacity_factor" to address the requirements mentioned in #515.

## How to review

Please make a copy of this branch, run the test with the existing code, and the tests should pass.

## PR checklist
- [x] Reduce duplication, i.e. don't have identical `model_generator()` methods in test_feature_temporal_level.py and test_feature_capacity_factor.py (@khaeru)
- [x] Release notes added.

### DIFF
--- a/message_ix/models.py
+++ b/message_ix/models.py
@@ -118,6 +118,7 @@ MESSAGE_ITEMS = {
     "cat_relation": item("set", "type_relation relation"),
     "cat_tec": item("set", "type_tec t"),
     "cat_year": item("set", "type_year y"),
+    "is_capacity_factor": item("set", "nl t yv ya h"),
     "level_renewable": dict(ix_type="set", idx_sets=["level"]),
     "level_resource": dict(ix_type="set", idx_sets=["level"]),
     "level_stocks": dict(ix_type="set", idx_sets=["level"]),
@@ -399,7 +400,7 @@ class MESSAGE(GAMSModel):
         # handled in JDBCBackend. For the moment, this code does not backstop that
         # behaviour.
         # TODO Extend to handle all masks, e.g. for new backends.
-        for par_name in ():
+        for par_name in ("capacity_factor",):
             # Name of the corresponding set
             set_name = f"is_{par_name}"
 
--- a/message_ix/models.py
+++ b/message_ix/models.py
@@ -331,6 +331,13 @@ class GAMSModel(ixmp.model.gams.GAMSModel):
         # Initialize the ixmp items
         cls.initialize_items(scenario, MESSAGE_ITEMS)
 
+    @staticmethod
+    def enforce(scenario):
+        """Enforce data consistency in `scenario`."""
+        # Implemented in MESSAGE sub-class, below
+        # TODO make this an optional method of the ixmp.model.base.Model abstract class
+        pass
+
     def __init__(self, name=None, **model_options):
         # Update the default options with any user-provided options
         model_options.setdefault("model_dir", config.get("message model dir"))
@@ -350,6 +357,10 @@ class GAMSModel(ixmp.model.gams.GAMSModel):
            simultaneously; but using *different* CPLEX options in each process may
            produce unexpected results.
         """
+        # Ensure the data in `scenario` is consistent with the MESSAGE formulation
+        # TODO move this call to ixmp.model.base.Model.run(); remove here
+        self.enforce(scenario)
+
         # If two runs are kicked off simulatenously with the same self.model_dir, then
         # they will try to write the same optfile, and may write different contents.
         #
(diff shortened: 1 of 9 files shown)


behnam-zakeri:
@khaeru, I went further and resolved the issue in the GAMS formulation, by adding a new masking set ("is_capacity_factor"). For populating this set on the python side, I copied the code you suggested in #514 (the part related to the `enforce()` method). I wanted to make sure that the tests designed here will work if we correct the formulation. Now, you can decide if we should add `enforce()` workflow here, or in a separate PR.

khaeru:
Will jump in here once #559 is addressed.

khaeru:
> Reduce duplication, i.e. don't have identical model_generator() methods in test_feature_temporal_level.py and test_feature_capacity_factor.py (@khaeru)

@behnam-zakeri I've now done this (about 2 hours), saving us about 400 lines of code in the test files while still doing all the same tests. To me, at least, they are now easier to read & distinguish from one another. However, one question came up in the process, so I hope you can help:

Two of the tests (`test_two_seasons_to_year_relative` and `test_linked_three_temporal_levels_relative`) had the following pattern of code:
```py
def test_func():
    try:
        do_stuff()
    except AssertionError:
        pass
```
This test will pass if do_stuff() raises an AssertionError. But it will *also* pass if do_stuff() doesn't raise any error at all. So the expected behaviour is ambiguous.

I changed these to the standard way of using pytest:
```py
def test_func():
    with pytest.raises(AssertionError):
        do_stuff()
```
This is more explicit: it will pass if do_stuff() raises an AssertionError as expected, but the test will fail if the error is *not* raised.

I find these tests now fail. From the docstrings/comments:
- `test_two_seasons_to_year_relative` says "Model should not solve," meaning we expect some kind of failure?
- `test_linked_three_temporal_levels_relative` says "Model solves" in the docstring but, in contrast, "Shouldn't pass with a relative time duration" in the comments.

Can you identify what's going on in each case? What do we actually expect—maybe that the model actually *should* solve in each case, but the assertions in check_solution(scen) should fail?

behnam-zakeri:
Thanks a lot @khaeru for reformatting and improvements. Related to your question
 “ … that the model actually should solve in each case, but the assertions in check_solution(scen) should fail? that the model actually should solve in each case, but the assertions in check_solution(scen) should fail?”
As you identified in these cases the model should solve but Assersion checks must not pass. These checks ensure commodity balances wok as expected across different temporal levels.

message_ix/tests/test_feature_capacity_factor.py
```diff
@@ -0,0 +1,156 @@
...
+        act.loc[i, "cf-corrected"] = act.loc[i, "duration-corrected"] / float(
+            cf.loc[i, "value"]
+        )
+    act = act.fillna(0).reset_index().set_index(["technology"])
+    # CAP = max("ACT" / "duration_time" / "capcity_factor")
+    for i in cap.index:
+        assert max(act.loc[i, "cf-corrected"]) == float(cap.loc[i, "lvl"])
+
+
+TD_0 = {
```
LauWien, line 35:
```suggestion
# Dictionary of technology input/output
TD_0 = {
```

message_ix/tests/test_feature_temporal_level.py
```diff
@@ -165,173 +37,123 @@ def model_generator(
...
-    model_generator(
-        test_mp,
-        comment,
+    tec_dict = TD_0.copy()
+    tec_dict["gas_ppl"]["time_origin"] = ["year"]
+
+    scen = make_subannual(
+        request,
         tec_dict,
         time_steps=[("summer", 1, "season", "year")],
```
LauWien, line 94: We could also think about moving this up and create e.g. TS_0 to reduce duplication (this line and line 75).

message_ix/tests/test_feature_capacity_factor.py
```diff
@@ -0,0 +1,156 @@
...
+            ("winter", 0.7, "season", "year"),
+        ],
+        demand={"summer": 2, "winter": 1},
+        capacity_factor={"gas_ppl": {"summer": 0.8, "winter": 0.8}},
+        var_cost={"gas_ppl": {"summer": 0.2, "winter": 0.2}},
+    )
+    check_solution(scen)
+
+
+TS_0 = [
```
LauWien, line 78:
```suggestion
# List of time step tuples
TS_0 = [
```

message_ix/tests/test_feature_capacity_factor.py
```diff
@@ -0,0 +1,156 @@
...
+]
+
+
+def test_capacity_factor_zero(request):
+    """Test zero capacity factor (CF) in a time slice.
+
+    "solar_pv_ppl" is active in "day" and NOT at "night" (CF = 0). It is expected that
+    the model will be infeasible, because "demand" at night cannot be met.
+    """
+    # Technology input/output
```
LauWien, line 90:
```suggestion
    # Dictionary of technology input/output
```

message_ix/tests/test_feature_capacity_factor.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def test_capacity_factor_zero_two(request):
+    """Test zero capacity factor (CF) in a time slice.
+
+    "solar_pv_ppl" is active in "day" and NOT at "night" (CF = 0). The model output
+    should show no activity of "solar_pv_ppl" at "night". So, "gas_ppl" is active at
+    "night", even though a more expensive technology.
+    """
+    # Technology input/output
```
LauWien, line 119:
```suggestion
    # Dictionary of technology input/output
```

LauWien (approved):
Looks good, thank you! Just some, suggestions in line.

message_ix/tests/test_feature_temporal_level.py
khaeru: Ah, true. I skipped this one because it doesn't save on # of lines.

message_ix/models.py
```diff
@@ -380,6 +392,32 @@ class MESSAGE(GAMSModel):
...
+            set_name = f"is_{par_name}"
+
+            # Existing and expected contents
+            existing = scenario.set(set_name)
+            expected = scenario.par(par_name).drop(columns=["value", "unit"])
+
+            if existing.equals(expected):
+                continue  # Contents are as expected; do nothing
+            # else:
+            #     scenario.add_set(set_name, expected)
```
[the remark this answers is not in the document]
khaeru: Yes, I'm not sure why; they're not on #514 from which this was cherry-picked. That PR will need to be deconflicted when it is rebased, so we can clean up/remove then.

khaeru:
Thanks for the review! Will merge once checks again pass.

khaeru:
Thanks @behnam-zakeri for the contributions! 🎉

PR #568, opened 2022-02-24
Address failing copying of GAMS files

### LINKED ISSUE #560: Copying and using MESSAGEix GAMS files as part of Anaconda installation on Windows 10 does not work
Following the [MESSAGEix installation instructions](https://docs.messageix.org/en/stable/install.html#copy-gams-model-files-for-editing) I wanted to copy (and use) the GAMS code to a separate folder for editing `message-ix copy-model --set-default /path/for/model/files`. The copying did not work with exception of the folder structure. Error output said _file exists, will not overwrite_ for all files. In addition, the `ixmp.config/config.json` file was successfully modified to include the specified path under the `message model dir` argument. However, even after copying the GAMS code manually to the specified folder, MESSAGEix continues to be run in the Anaconda message_ix package folder `anaconda3/envs/message_env/Lib/site-packages/message_ix/model`.

The environment I am using is Windows 10, Anaconda 2021.11/Python 3.10.2, message_ix 3.4.0. Note that I have looked #416 which describes the same issue, but without suggesting a solution to the problem.

LauWien:
This PR adjusts the CLI `message-ix copy-model` function. 
Closes #560. 

## How to review

- Run `message-ix copy-model --set-default "PATH"` to check if CLI works.

Estimated time: < 3 minutes

## PR checklist

### DIFF
--- a/message_ix/cli.py
+++ b/message_ix/cli.py
@@ -42,7 +42,7 @@ def copy_model(path, overwrite, set_default):
     src_dir = Path(__file__).parent / "model"
     for src in src_dir.rglob("*.*"):
         # Skip certain files
-        if src.suffix in (".gdx", ".log", ".lst") or re.search("225", str(src)):
+        if src.suffix in (".gdx", ".log", ".lst") or re.search("225[a-z]+", str(src)):
             continue
 
         # Destination path
--- a/message_ix/cli.py
+++ b/message_ix/cli.py
@@ -1,3 +1,4 @@
+import re
 from pathlib import Path
 
 import click
@@ -41,7 +42,7 @@ def copy_model(path, overwrite, set_default):
     src_dir = Path(__file__).parent / "model"
     for src in src_dir.rglob("*"):
         # Skip certain files
-        if src.suffix in (".gdx", ".log", ".lst"):
+        if src.suffix in (".gdx", ".log", ".lst") or re.search("(225)", src.stem):
             continue
 
         # Destination path
--- a/message_ix/cli.py
+++ b/message_ix/cli.py
@@ -55,7 +55,7 @@ def copy_model(path, overwrite, set_default):
             continue
 
         # Display output
-        if dst.exists:
+        if dst.exists():
             if not overwrite:
                 print("{} exists, will not overwrite".format(dst))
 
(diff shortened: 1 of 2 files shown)


khaeru:
A hint on improving the tests a little here to help debug:

https://github.com/iiasa/message_ix/blob/aae6e31c386fef96f96fb5415d468ebf703e1d85/message_ix/tests/test_cli.py#L10-L12

The variable `r` is a [`click.testing.Result`](https://click.palletsprojects.com/en/8.0.x/api/#click.testing.Result) instance. So we can use the [`assert` statement](https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-assert_stmt) to show more information, e.g.

```py
assert 0 == r.exit_code, (r.exception, r.output)
```
This way, if the assertion/test fails, the stuff after the comma is printed out as part of the test failure message.

LauWien:
> This way, if the assertion/test fails, the stuff after the comma is printed out as part of the test failure message.

Many thanks!

message_ix/cli.py
```diff
@@ -39,9 +40,9 @@ def copy_model(path, overwrite, set_default):
     path = Path(path).resolve()
 
     src_dir = Path(__file__).parent / "model"
-    for src in src_dir.rglob("*"):
+    for src in src_dir.rglob("*.*"):
         # Skip certain files
-        if src.suffix in (".gdx", ".log", ".lst"):
+        if src.suffix in (".gdx", ".log", ".lst") or re.search("(225)", src.stem):
```
khaeru, line 45: I suspect this won't work, since the "stem" is only the part of the filename before the suffix. E.g. in /path/to/files/225a/etc/foo.gdx, the stem is "foo". We want to check if any entire part is `225[a-z]{1,2}`.

Can you experiment locally? Later (not today) I can show how to extend the tests to check this.

LauWien: Thank you for the review!
Ugh, yes, I checked now - it's not working. Apparently I checked something else before pushing which worked.. Sorry for that.

> Can you experiment locally?

Yes!

> Later (not today) I can show how to extend the tests to check this.

Thank you! Will also look into that before.

LauWien: Checking locally with the current adjustment, a folder called "225" gets skipped, but maybe there is a better way do do that. 
I've tried to implemented a test to check if this behavior is working, I guess this could also be optimized.

LauWien:
Should we squash and merge, or should I `git rebase -i HEAD~..` to clean up the commits so we can use a regular merge commit? Or do you think the commits, even though they some of them were corrected, should be kept?

Issue #571, opened 2022-03-21
Return vintage_and_active_years() for a specific technology


khaeru:
(Text by @OFR-IIASA—thanks!)

This issue aims to summarize an ongoing discussion on ECE Slack about a new, desirable feature.

The aim of the feature is to ease the creation of technology specific parameters which correlate vintage and active years, but limited to those that are feasible given a specific technology's `technical_lifetime`.

Possibilities for implementation that were discussed:

## Option 1

Combine/modify two existing functionalities as suggested by @khaeru:

1.  `scen.vintage_and_active_years()` returns a dataframe with the index `year_vtg` and `year_act` and correlates all the possible activity years within the optimization time horizon (>= `firstmodelyear`) with all possible vintage years, including also vintage years prior to the `firstmodelyear`.
2. `scen.years_active()` returns a list of activity years for any given node/technology/vintage_year.

The results of `scen.vintage_and_active_years()` would then be filtered based on `scen.years_active()`.

Note that it would be necessary to also allow for the possibility to retrieve activity years prior to `firstmodelyear`, meaning that `scen.vintage_and_active_years()` would need to be modified so that activity years < `firstmodelyear` are also returned. This is necessary so that historical reporting can be carried out. Possibly a filter option is required e.g. as for the global model we report historic values from 1990 onwards.

## Option 2

As suggested by @behnam-zakeri  and extended by @OFR-IIASA:

```python
from itertools import product
tec = "coal_ppl"
node = "R11_AFR"

def new_function_name(node, tec, vtg_lower=0, act_lower=0):
    """Returns vintage and corresponding activity years.
    
    Parameter
    ---------
    node : str
        Node name.
    tec : str
        Technology name.
    vtg_lower : int
        Vintage year as of which data should be returned.
    act_lower : int
        Activity year as of which data should be returned.
        
    Returns
    -------
    pandas.DataFrame
        with columns 'year_vtg' and 'year_act', in which each row is a
        valid pair.
    """
    
    # Retrieve model years for vintages should be added 
    yr_vtg = [x for x in scen.set("year") if x >= vtg_lower]
    
    # Retrieve model years for which activity should be added
    yr_act = [x for x in yr_vtg if x >= act_lower]
    
    # Retrieve technical lifetime
    tec_lifetime = scen.par("technical_lifetime",
                            filters={"node_loc": node,
                                     "technology": tec}
                            ).year_vtg.tolist()
    
[...]

PR #286, opened 2019-12-17
Set duration_period in add_horizon() 

francescolovat:
This PR closes #282.

This PR edits the `add_horizon` method for `Scenario` class found in `/message_ix/core.py`

Change of `data` argument to `**kwargs`, and added automatic generation of *duration_period* parameter. Added description in the docstring of `add_horizon`.

Still, from #282 the following bullet needs to be addressed:

> - Clarify and test whether it should be possible to call add_horizon more than once. If so perhaps warn or return a value if the user adds overlapping sets of years; if not, raise an exception.

Please help, @khaeru I don't know why the changes in the `core.py` file are not reflected in the jupyter notebook files for the tutorials (I was testing my changes in the Westeros tutorial file but nothing was changing when I was editing or even commenting some methods in `core.py`).

**PR checklist:**

- [x] Tests added.
- [x] Documentation added.
- [x] Release notes updated.

PR #572, opened 2022-03-24
Extend feature vintage_and_active_years()

OFR-IIASA:
This PR extends the features of the function `vintage_and_active_years()` based on issue #571 

## How to review

- Build the documentation and look at a description of `vintage_and_active_years()`
  does in the future.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/message_ix/tests/test_feature_vintage_and_active_years.py
+++ b/message_ix/tests/test_feature_vintage_and_active_years.py
@@ -68,11 +68,15 @@ def test_vintage_and_active_years(test_mp):
     obs = scen.vintage_and_active_years(
         ya_args=("foo", "bar", years[0]), in_horizon=True
     )
-    pdt.assert_frame_equal(pd.DataFrame(columns=["year_vtg", "year_act"]), obs)
+    pdt.assert_frame_equal(
+        pd.DataFrame(columns=["year_vtg", "year_act"]), obs, check_dtype=False
+    )
 
     # Exception is raised for incorrect arguments
-    with pytest.raises(ValueError, match="3 arguments are required if using `ya_args`"):
-        scen.vintage_and_active_years(ya_args=("foo", "bar"))
+    with pytest.raises(
+        ValueError, match="At least 2 arguments are required if using `ya_args`"
+    ):
+        scen.vintage_and_active_years(ya_args=("foo"))
 
 
 def test_vintage_and_active_years_upd(test_mp):
--- a/message_ix/tests/test_feature_vintage_and_active_years.py
+++ b/message_ix/tests/test_feature_vintage_and_active_years.py
@@ -1,9 +1,9 @@
 import pandas as pd
 import pandas.testing as pdt
+import pytest
 
 from message_ix import Scenario
 from message_ix.testing import SCENARIO
-import pytest
 
 
 def test_vintage_and_active_years(test_mp):
--- a/message_ix/tests/test_feature_vintage_and_active_years.py
+++ b/message_ix/tests/test_feature_vintage_and_active_years.py
@@ -1,0 +1,272 @@
+import pandas as pd
+import pandas.testing as pdt
+
+from message_ix import Scenario
+from message_ix.testing import SCENARIO
+import pytest
+
+
+def test_vintage_and_active_years(test_mp):
+    scen = Scenario(test_mp, **SCENARIO["dantzig"], version="new")
+
+    years = [2000, 2010, 2020]
+    scen.add_horizon(year=years, firstmodelyear=2010)
+    obs = scen.vintage_and_active_years()
+    exp = pd.DataFrame(
+        {
+            "year_vtg": (2000, 2000, 2010, 2010, 2020),
+            "year_act": (2010, 2020, 2010, 2020, 2020),
+        }
+    )
+    pdt.assert_frame_equal(exp, obs, check_like=True)  # ignore col order
+
+    # Add a technology, its lifetime, and period durations
+    scen.add_set("node", "foo")
+    scen.add_set("technology", "bar")
+    scen.add_par(
+        "duration_period", pd.DataFrame({"unit": "???", "value": 10, "year": years})
+    )
+    scen.add_par(
+        "technical_lifetime",
+        pd.DataFrame(
+            {
+                "node_loc": "foo",
+                "technology": "bar",
+                "unit": "???",
+                "value": 20,
+                "year_vtg": years,
+            }
+        ),
+    )
+
+    # part is before horizon
+    obs = scen.vintage_and_active_years(ya_args=("foo", "bar", "2000"))
+    exp = pd.DataFrame({"year_vtg": (2000,), "year_act": (2010,)})
+    pdt.assert_frame_equal(exp, obs, check_like=True)  # ignore col order
+
+    obs = scen.vintage_and_active_years(
+        ya_args=("foo", "bar", "2000"), in_horizon=False
+    )
+    exp = pd.DataFrame({"year_vtg": (2000, 2000), "year_act": (2000, 2010)})
+    pdt.assert_frame_equal(exp, obs, check_like=True)  # ignore col order
+
+    # fully in horizon
+    obs = scen.vintage_and_active_years(ya_args=("foo", "bar", "2010"))
+    exp = pd.DataFrame({"year_vtg": (2010, 2010), "year_act": (2010, 2020)})
+    pdt.assert_frame_equal(exp, obs, check_like=True)  # ignore col order
+
+    # part after horizon
+    obs = scen.vintage_and_active_years(ya_args=("foo", "bar", "2020"))
+    exp = pd.DataFrame({"year_vtg": (2020,), "year_act": (2020,)})
+    pdt.assert_frame_equal(exp, obs, check_like=True)  # ignore col order
+
+    # Advance the first model year
+    scen.add_cat("year", "firstmodelyear", years[-1], is_unique=True)
+
+    # Empty data frame: only 2000 and 2010 valid year_act for this node/tec;
+    # but both are before the first model year
+    obs = scen.vintage_and_active_years(
+        ya_args=("foo", "bar", years[0]), in_horizon=True
+    )
+    pdt.assert_frame_equal(pd.DataFrame(columns=["year_vtg", "year_act"]), obs)
+
+    # Exception is raised for incorrect arguments
+    with pytest.raises(ValueError, match="3 arguments are required if using `ya_args`"):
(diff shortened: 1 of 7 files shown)


OFR-IIASA:
@khaeru I have revised the function `vintage_and_active_years()` to meet the requirements of existing and new tests. Note that the existing tests for this function have been (temporarily?) carried over to `test_feature_vintage_and_active_years.py`. Hence it is not unexpected for the core tests to fail until the revisions are complete. I would wait for your feedback on the function revision itself before finalizing the clean-up of tests and adding the remaining documentation to the the function.

OFR-IIASA:
@adrivinca yes, the tests if have this use-case as a test.

adrivinca:
Thanks! I cloned the branch and tried it, it seems ok and the tests as well.
is there any particular review process needed? or could you @OFR-IIASA please rebase the branch and see if tests on Github pass?

adrivinca:
It seems the tests fail on mac, do you understand why? I don't really

OFR-IIASA:
@adrivinca the reason for these test failing was because this was a test carried out in `test_core` and based on the original function. The tests have been moved, adjusted and are included in a separate file. I have now removed the obsolete tests from `test_core`.

OFR-IIASA:
@adrivinca and @khaeru I have made one last adjustment to the PR and added a test for it. Ideally, this function will replicate the behvior of map_tec_lifetime in that the returned activity years for any given vintage wont exceed the maximum technical_lifetime defined for a technology. The documentation has also been updated as have the release notes.

khaeru:
@adrivinca asked me to take a look at this. I can check more extensively perhaps tomorrow or Friday, but at a skim of the thread I'm concerned about this statement:

> the reason for these test failing was because this was a test carried out in `test_core` and based on the original function. The tests have been moved, adjusted and are included in a separate file. I have now removed the obsolete tests from `test_core`.

Please read:
- [pandas' deprecation policy](https://pandas.pydata.org/docs/development/policies.html#version-policy), which we essentially follow.
- Item 7 in the [semantic versioning spec](https://semver.org/#spec-item-7): “[Minor version Y (x.Y.z)] MUST be incremented if any public API functionality is marked as deprecated.”
- …which is linked from our own [versioning policy](https://docs.messageix.org/en/stable/contrib/version.html), in which we again restate “A minor version increment may fix bugs or add new features, but does not change existing functionality.”

To expand:
- A user with code that calls any message_ix function (e.g. `vintage_and_active_years()`) must experience **no surprises** when upgrading from a ‘current’ version (e.g. 3.5.0) to the ‘next’ (e.g. 3.6.0 or 4.0.0). That is, their code must not start to produce different results, without prior warning, simply because of an upgrade.
- If we plan that the (default) behaviour of a function will change, then the next version **must** **warn** about the upcoming change, but **not change** the behaviour.
  - This gives the user a clear signal and time (at least one release cycle) to adjust their code. The release notes must describe what the upcoming change will be, and should give tips on how to handle it.
- Then, in a subsequent major version (e.g. 4.0.0), the change can become effective/the default, and the warning is removed. ([SemVer spec item 8](https://semver.org/#spec-item-8): “Major version X (X.y.z) MUST be incremented if any backwards incompatible changes are introduced to the public API.”)

See some examples in Pandas: https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.4.0.html#deprecations

Sometimes it takes a little thinking, but it is always possible to introduce new functionality in a non-breaking way; often with an added optional argument. If existing tests fail, that means we are introducing breaking changes, and we can't respond by simply removing the tests.

OFR-IIASA:
@khaeru thanks for the comments. The old tests have been moved from the core to a separate file. If you look at the differences, then the `check_dtype=False` was added to one check and the tests which looks at "# Exception is raised for incorrect arguments" naturally needs to be changed because the message is changed. All the old tests still exist/pass.

message_ix/tests/test_feature_vintage_and_active_years.py
```diff
@@ -0,0 +1,344 @@
...
+
+    # Add years of differing time length
+    years = [2000, 2005, 2010, 2015, 2020, 2030]
+    scen.add_horizon(year=years, firstmodelyear=2010)
+
+    # Check if default function call is valid
+    obs = scen.vintage_and_active_years()
+    exp = pd.DataFrame(
+        {
+            "year_vtg": (
```
khaeru, line 93: Because these take up many lines, and similar values occur repeatedly (perhaps with small differences that are hard to spot), it makes sense to e.g.:
- Use a global variable. This way it is clear when the same values are used in multiple tests.
- Write a short utility function to generate the values.

This can make the tests easier to read.

I will do these things as part of checking/tidying the modifications to the core code.

message_ix/tests/test_feature_vintage_and_active_years.py
```diff
@@ -0,0 +1,344 @@
...
+    obs = scen.vintage_and_active_years(ya_args=("foo", "bar", "2010"))
+    exp = pd.DataFrame(
+        {"year_vtg": (2010, 2010, 2010, 2010), "year_act": (2010, 2015, 2020, 2030)}
+    )
+    pdt.assert_frame_equal(exp, obs, check_like=True)  # ignore col order
+
+    # Check if no vintge-year is passed, that all values corresponding
+    # to technical lifetime are passed if the active years >= 2010
+    obs = scen.vintage_and_active_years(ya_args=("foo", "bar"))
+    exp = pd.DataFrame(
```
khaeru, line 167: @OFR-IIASA can you please help me understand this expected value?
- `technical_lifetime` is 20 for all periods.
- The pair yv=2000, ya=2020 and yv=2000, ya=2030 are excluded. I guess this is because 2020 - 2000 ≥ 20.
- However, the pair yv=2010, ya=2030 are *in*cluded, even though 2030 - 2010 ≥ 20.

What is the difference?

khaeru: Thanks, understood. This accords with our documentation.

khaeru:
Okay, I've streamlined the tests here; about 2:45 of work. Some points plus comments inline.
- When using `add_horizon()`, it's not necessary to manually set `duration_period` since #286 (August 2020).
- The general principle is “DRY” (don't repeat yourself):
  A. Some code differs from one test to the next. This represents crucial differences in the invocation and expected outcomes.
  B. A lot of the code does not.
  The more (B) there is, the harder it is to see (A). So it really helps to hide all the repeated bits away in one or a few utility functions—this makes the tests more legible to oneself and other developers.
- In this case the length of the new file is cut by about half, plus see the comments.

I'll next start to look at the actual implementation and the documentation of the method.

message_ix/tests/test_feature_vintage_and_active_years.py
```diff
@@ -0,0 +1,203 @@
...
+    obs = scen.vintage_and_active_years()
+    exp = _q(yvya_all, f"year_act >= {fmy}")
+    assert_frame_equal(exp, obs)
+
+    # Check standard functionality with different period duration lengths
+    obs = scen.vintage_and_active_years(ya_args=("foo", "bar", "2010"))
+    exp = _q(yvya_all, f"year_vtg == 2010 and year_act >= {fmy}")
+    assert_frame_equal(exp, obs)
+
+    # Check if no vintge-year is passed, that all values corresponding to technical
```
khaeru, line 159: With the preparation of `exp` now slimmed down a lot, we can see that this and the next three assertions each have an identical comment above them. This doesn't help understand what is *different* between them.

khaeru:
@OFR-IIASA in [292edf4](https://github.com/iiasa/message_ix/pull/572/commits/292edf4e29c400177d04dae7e2b96199dffe0ab0) I add some tests/examples showing how the functionality of the added `act_lower` and `vtg_lower` arguments can be handled using pandas built-ins.

Kind of like `make_df()` in #415 (“helpful messages pointing users to existing Python/pandas functions that perform identical functions”), I think it is better to teach users how to use pandas more fully than to give ourselves the burden of maintaining/supporting new functionality. Another advantage is that the user doesn't need to memorize a new argument name; they can just `.query()` using the usual column names.

Is this acceptable? If so, I'll remove the added arguments.

message_ix/tests/test_feature_vintage_and_active_years.py
```diff
@@ -0,0 +1,242 @@
...
+    fmy = years[2]
+
+    # Last year for which the technical_lifetime of "bar" will be defined
+    y_max = years[-2]
+
+    scen, yvya_all = _setup(test_mp, years, fmy, filter(lambda y: y <= y_max, years))
+
+    # Check standard functionality with different period duration lengths
+    obs = scen.vintage_and_active_years(ya_args=("foo", "bar", "2010"))
+    exp = pd.DataFrame({"year_vtg": (2010, 2010, 2010), "year_act": (2010, 2015, 2020)})
```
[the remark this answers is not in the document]
OFR-IIASA: So this test does back to pick up the definition of "map_tec_lifetime" (see also issue #583). That means if there is no technical lifetime defined for a period, despite it being in the lifetime horizon of the technology, then no parameter will be returned. If this isnt done, then we will have an issue when generating the matrix as no technological lifetime is defined for an activity year and thus results in a compilation error.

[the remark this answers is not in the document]
OFR-IIASA: At the end of the day, this functionality should replicate the `map_tec_lifetime` set entries.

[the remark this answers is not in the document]
khaeru: Given that there is still a lack of clarity or documentation around the actual meaning of that parameter and whether the current code correctly implements it (https://github.com/iiasa/message_ix/issues/583#issuecomment-1129658961), I think it is bad for us to bake a counter-intuitive, implementation-specific meaning into this function.

I would rather we preserve a straightforward behaviour that is easy to understand, and advertise a workaround. Can we say something along the lines of: “When using `.vintage_and_active_years()` to prepare data for [which specific MESSAGEix parameters?], you may need to discard rows where `year_act` is greater than [specific quirk related to `map_tec_lifetime`], as follows: [code example]"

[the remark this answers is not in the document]
khaeru: Added the *tl_only* argument per this discussion.

message_ix/tests/test_feature_vintage_and_active_years.py
khaeru: I rewrote these comments.

khaeru:
@OFR-IIASA aside from the persistent dtype issues on Windows, which I will fix, this is complete per our discussion. Please have a look and comment if you agree.

OFR-IIASA:
@khaeru thanks for the formidable rework. Everything is as we agreed.

Issue #521, opened 2021-10-12
Mismatch of "time" and "time_act" in RENEWABLE_EQUIVALENCE


behnam-zakeri:
In equation `RENEWABLE_EQUIVALENCE`, the index of sub-annual time slices for `ACT` of the technology linked to the `renewable_potential` is specified as `time`:
https://github.com/iiasa/message_ix/blob/2cf0df3a3257d74fd423ff50818f33265c6cd831/message_ix/model/MESSAGE/model_core.gms#L816
But parameter `input` of the technology in the same equation is specified with the index of `time_act` for activity. This creates issues in models with sub-annual time slices, when the technology linked to a renewable potential bin has two different entries for sub-annual time slices, like "year" and "summer", in parameter `input`.

PR #581, opened 2022-04-13
Correct index of "time" in equation RENEWABLES_EQUIVALENCE

behnam-zakeri:
There is a typo in GAMS equations specified in issue #521, which is corrected through this PR. The inline documentation is already correct and does not need any change. This PR does not need any additional test.

## How to review

**Required:** 
Please look at equation `RENEWABLES_EQUIVALENCE` in `model_core.gms`. Parameter `input` is indexed with `time_act` and `time`, multiplied by variable `ACT` which is indexed by `time`. Here, the correct index of time for ` ACT` must be `time_act`. To check this, one needs to know if the order of indexes of `time_act` and `time` in parameter `input` is correct or not. This can be checked, for example, by running a command `scenario.idx_names("input")`.

## PR checklist

- [x] Add or expand tests; coverage checks both ✅ ==> Not needed
- [x] Add, expand, or update documentation ==> Not needed
- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -10,7 +10,7 @@ Files compatible with v3.4.0 and earlier will not work with this version and sho
 All changes
 -----------
 
-- Correct typo in GAMS formulation, eq. `RENEWABLE_EQUIVALENCE` (:pull:`581`).
+- Correct typo in GAMS formulation, :ref:`equation_renewables_equivalence` (:pull:`581`).
 - Improve configurability of :mod:`.macro`; see the :doc:`documentation <macro>` (:pull:`327`).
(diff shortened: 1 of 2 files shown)


LauWien (approved):
Looks good :)
Just to be sure it is fine, that in the following `time` is used:
https://github.com/iiasa/message_ix/blob/6bf19c37c6a83be0126ec1c7d918040550dbe24b/message_ix/model/MESSAGE/model_core.gms#L813

behnam-zakeri:
> Looks good :) Just to be sure it is fine, that in the following `time` is used:
> 
> https://github.com/iiasa/message_ix/blob/6bf19c37c6a83be0126ec1c7d918040550dbe24b/message_ix/model/MESSAGE/model_core.gms#L813

Thanks for spotting that. Yes, that must be `time_act` too.

PR #645, opened 2022-09-06
Add additional oscillation indicator

OFR-IIASA:
This PR adds an additional check to detect the oscillation between the various iterations when running MESSAGE-MACRO.
The additional check looks to see if the objective function bounces between two points.

Some of the findings and solutions to handling unscaled infeasibilities discussed in issue #337 have also been updated.

## How to review

Review code updates to "MESSAGE-MACRO_run.gms".
Review updated documentation to the section "calibrate-and-tune-message-macro" found in "marco.rst".

## PR checklist

- [x] Continuous integration checks all ✅
~- [ ] Add or expand tests; coverage checks both ✅~ N/A as this cannot be tested
- [x] Add, expand, or update documentation.
- [x] Update release notes.

PR #648, opened 2022-09-14
Correct calculation of COST_NODAL_NET for standalone MESSAGE

OFR-IIASA:
This PR corrects the calculation of `COST_NODAL_NET` in standalone MESSAGE.

The fact that `COST_NODAL_NET` was divided by 1000 in `model_solve.gms` and then again in `macro.py`, meant that the PARAMETER total_cost during the calibration process received values which were inconsistent over time. The values for the model periods were a factor 1000 smaller than those provided in Excel configuration sheet.

`COST_NODAL_NET` as calculated by `model_solve.gms` and `MESSAGE-MACRO_run.gms` are now consistent.

The PR additionally adds documentation to expand on the existing calibration workflow. 

## How to review

## PR checklist

- [x] Update release notes.


message_ix/tests/test_macro.py
```diff
@@ -272,18 +272,18 @@ def test_calibrate_roundtrip(westeros_solved):
     aeei = with_macro.par("aeei")["value"].values
     npt.assert_allclose(
         aeei,
-        1e-3 * np.array([20, -7.5142661, 43.6256127, 21.14346]),
+        1e-3 * np.array([20, -7.53474955303437, 43.6388021078658, 21.1470758210961]),
```
gidden, line 275: If these numbers are sensitive to the second decimals place, maybe it is better to use a [looser tolerance](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html)

message_ix/tests/test_macro.py
```diff
@@ -272,18 +272,18 @@ def test_calibrate_roundtrip(westeros_solved):
...
     npt.assert_allclose(
         grow,
         1e-3
         * np.array(
             [
-                26.5836313,
-                69.1417537,
-                79.1435885,
-                24.5225555,
+                26.583631304232,
```
gidden, line 283: same comment here

message_ix/tests/test_macro.py
gidden: Or maybe not - these changes are a result of the factor 1e3 update, right? So perhaps it is necessary to test at this precision if large changes to certain variables affect these values by such a small amount..

gidden:
@OFR-IIASA - is there any additional 'spot' testing we can do with this change (e.g., on a country model or even the global model) to understand its implications?

FYI @volker-krey / @behnam-zakeri

message_ix/tests/test_macro.py
OFR-IIASA: I would also say that we should keep this for the moment. The change of the aeei values is a result of change above (factor 1e3 update) as you indicated.

OFR-IIASA:
> 

I have done a comparison of two scenario sets which are identical apart from this specific change. The changes are not very big and overall the effects barely noticeable. The biggest changes of some of my spot checks is the fact that trade changes; but even here the changes are minimal and do not impact overall results. What does change though is the fact that without this change I was getting "division by EPS" after a few macro iterations for prices higher than 190$/tCO2 in 2020 (rising with the discount rate). With the changes I encountered no issues up until 250$ - the highest price I tested until now and I will see if I will push the model further, but 250$ in 2025 results in 10.7k$/tCO2 in 2100. A side note on the oscillation checks and iteration logs; these seem very similar so i do not think that it resolves the issue that we dont get "natural convergence" at very low prices (but I havent tested this either with this change).

OFR-IIASA:
An update on some tests:
- Higher carbon prices of up to 340$/tCO2 in 2025 (~13k in 2100) had no issues.
- Running a C-Price scenario initially calibrated with the current `main` branch shows identical results when run with this branch.
- The changes to the calibration as a result of this PR do not have any impact on a 1$/tCO2 C-Price scenario converging naturally. Oscillation check mechanisms are still required.

OFR-IIASA:
@khaeru - As discussed in the MESSAGEix meeting, the current PRs are failing due to an issue related to upgrading R as part of "Run ts-graphviz/setup-graphviz@v1" [see link](https://github.com/iiasa/message_ix/actions/runs/3103048722/jobs/5027235992#step:12:125)

khaeru:
> As discussed in the MESSAGEix meeting, the current PRs are failing due to an issue related to upgrading R as part of "Run ts-graphviz/setup-graphviz@v1" [see link](https://github.com/iiasa/message_ix/actions/runs/3103048722/jobs/5027235992#step:12:125)

The fix for this is to "cherry-pick" (not really, since it's a different repo, but essentially the same change) https://github.com/iiasa/ixmp/commit/6e0f74d446e0915374932a78dd0393a081d4882c (from iiasa/ixmp#457) with the same commit message, i.e. referencing the upstream issue.

> It also seems that there is another error, possibly resulting from the recent pandas update (?) [see link](https://github.com/iiasa/message_ix/actions/runs/3103048722/jobs/5027236238#step:16:218)

I [mentioned this in Slack](https://iiasa-ece.slack.com/archives/CD0GBHHA4/p1663667213431939) the first day it appeared. Someone will need to identify and implement a fix.

behnam-zakeri (commented):
Thanks @OFR-IIASA for this. It will make the values of `COST_NODAL_NET` consistent across the two models. I reviewed and edited the text a bit to be clearer. Content wise this is ready to be merged.

khaeru:
@OFR-IIASA @gidden @behnam-zakeri—thanks for the work on this! Note that:

- This PR resulted in a line (the one for #645) appearing twice in the release notes: see https://github.com/iiasa/message_ix/blob/main/RELEASE_NOTES.rst or [here](https://docs.messageix.org/en/latest/whatsnew.html#next-release) in the built docs for `main`.
- Some of the edits to doc/macro.rst result in a sentence appearing in all bold [here](https://docs.messageix.org/en/latest/macro.html#general-configuration-sheet), which I assume was not intended.

Please try to look at these things when preparing or reviewing a PR.

behnam-zakeri:
Thanks, @khaeru, for noticing this. Should I open a PR and fix this, or do we do this as part of another PR?

Issue #652, opened 2022-09-20
Levelized costs not calculated for technologies with only `var_cost`


OFR-IIASA:
The `levelized_cost` are used to determine the penalty associated with the additional relaxation of dynamic constraints via the `soft_activity_lo/_up` in combination with `level_cost_activity_soft_lo/_up`. 

Currently, the calculation of [levelized_costs](https://github.com/iiasa/message_ix/blob/1680452d53ed646eca1b84e7446da0f255d8351c/message_ix/model/MESSAGE/scaling_investment_costs.gms#L35) excludes all technologies which do not have investment-costs parametrized. This means that for technologies where only variable costs are specified, the "soft" relaxation is without cost.

#### Expected result
If for a technology only the parameter `var_cost` has been defined, then `levelized_cost` are calculated.

#### Problem description
If for a technology only the parameter `var_cost` has been defined, then `levelized_cost` are NOT calculated.

#### Versions

v3.6

behnam-zakeri:
@OFR-IIASA thanks for opening this. I believe `levelized_cost` is defined as the cost of one unit of _capacity_ of a given technology levelized over the lifetime. So, indeed it's a capacity-related parameter, and should be read as "Levelized cost of one unit of capacity". This is not equal to what is more commonly known as "Levelized cost of electricity" or LCOE. I have two suggestions for the way forward:
- making the definition clear in the documentation
- including `relation_cost` in the formulation
- defining another auxiliary parameter for "levelized cost of one unit of output commodity" for technologies. This has been a desirable feature, and I believe you and @Jihoon have worked previously to calculate.

PR #653, opened 2022-09-21
Correct levelized_cost calculation

OFR-IIASA:
As explained in issue #652, this PR aims to adjust the calculation of levelized_costs, so that these are not only calculated when investment costs are parametrized for a technology, but also when variable costs *only* have been defined.

## How to review

A Specific test cannot be written because it is not possible to retrieve the parameter "levelized_cost" from the scenario object.

## PR checklist

~- [ ] Add or expand tests; coverage checks both ✅~
- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -7,6 +7,7 @@ All changes
 - Adjust :meth:`.Scenario.add_macro` calculations for pandas 1.5.0 (:pull:`656`).
 - Add additional oscillation detection mechanism for MACRO iterations (:pull:`645`).
 - Correct calculation of `COST_NODAL_NET` for standalone MESSAGE (:pull:`648`)
+- Ensure `levelized_cost` are also calculated for technologies with only variable costs (:pull:`653`)
 - Add additional oscillation detection mechanism for macro iterations (:pull:`645`)
 
 .. _v3.6.0:
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -4,10 +4,9 @@ Next release
 All changes
 -----------
 
-- Adjust :meth:`.Scenario.add_macro` calculations for pandas 1.5.0 (:pull:`656`).
-- Add additional oscillation detection mechanism for MACRO iterations (:pull:`645`).
-- Correct calculation of `COST_NODAL_NET` for standalone MESSAGE (:pull:`648`)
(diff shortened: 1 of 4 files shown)


OFR-IIASA:
@behnam-zakeri as discussed yesterday, let us extend the calculation of the levelized costs to include the relation-related costs as well. Also, the issues raised on @khaeru can also be addressed in this PR:
- This PR resulted in a line (the one for https://github.com/iiasa/message_ix/pull/645) appearing twice in the release notes.
- Some of the edits to doc/macro.rst result in a sentence appearing in all bold [here](https://docs.messageix.org/en/latest/macro.html#general-configuration-sheet).

gidden:
This seems like a key update that should be tested if at all possible. I wonder if it's possible to use the existing dantzig problem to test levelized cost values with
- `inv_cost` only
- `fix_cost` only
- both `inv_cost` and `fix_cost`

OFR-IIASA:
@gidden I agree, and also to test the 'var_cost' as well as the additions to the formulation related to relations. I believe by solving using the command solve(var_list=["levelized_cost"]) we will hopefully be able to retrieve the levelized_cost from the scenario object.

OFR-IIASA:
In order to address the issues associated with creating tests for this PR, an issue has been started, see #661

behnam-zakeri (changes requested):
Thanks @OFR-IIASA . Can you please merge my edits/corrections suggested through #658 so that we see all proposed changes together here?

OFR-IIASA:
@behnam-zakeri , my proposal would be to merge this PR first and then change the PR for #658 to main. The reason is that while here we are only just changing "when" the `levelized_cost` are being calculated, the changes in #658 change how these are calculated. I would therefore propose to add tests for the other PR, but for some changes on the java end are required to allow doing this. I therefore propose to merge this PR as is and continue work on #658 thereafter.

behnam-zakeri (approved):

gidden (approved):
LGTM, thanks @OFR-IIASA ! I will update to latest main and merge.

gidden:
The failed test here is from a test timeout, e.g., 

```
FAILED message_ix/tests/test_tutorials.py::test_tutorial[westeros_addon_technologies] - nbclient.exceptions.CellTimeoutError: A cell timed out while it was being executed, after 10 seconds.
```

Will merge now.

gidden:
Noting previous tests passed except the noted failures, merging.

PR #654, opened 2022-09-21
Correct calculation of NEW_CAPACITY_CONSTRAINTs

OFR-IIASA:
This PR addresses an issue which occurs when calculating the upper and lower constraints for new capacity (`NEW_CAPACITY_CONSTRAINT_LO` and `NEW_CAPACITY_CONSTRAINT_UP`).
The issue resulting from the current formulation will vary the magnitude of the bound when the `duration_period` of the time-period for the bound is being calculated, differs to the previous time-period.
An adjustment of the bound is made to take into consideration the change in the `duration_period`.

## How to review

## PR checklist

- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -7,7 +7,7 @@ All changes
 - Adjust :meth:`.Scenario.add_macro` calculations for pandas 1.5.0 (:pull:`656`).
 - Add additional oscillation detection mechanism for MACRO iterations (:pull:`645`).
 - Correct calculation of `COST_NODAL_NET` for standalone MESSAGE (:pull:`648`)
-- Account for difference in period-length when calculating `NEW_CAPACITY_CONSTRAINT_LO` and `NEW_CAPACITY_CONSTRAINT_UP` (:pull:`654`)
+- Account for difference in period-length in equations `NEW_CAPACITY_CONSTRAINT_LO` and `NEW_CAPACITY_CONSTRAINT_UP` (:pull:`654`)
 - Add additional oscillation detection mechanism for macro iterations (:pull:`645`)
 
 .. _v3.6.0:
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -7,6 +7,7 @@ All changes
 - Adjust :meth:`.Scenario.add_macro` calculations for pandas 1.5.0 (:pull:`656`).
 - Add additional oscillation detection mechanism for MACRO iterations (:pull:`645`).
 - Correct calculation of `COST_NODAL_NET` for standalone MESSAGE (:pull:`648`)
+- Account for difference in period-length when calculating `NEW_CAPACITY_CONSTRAINT_LO` and `NEW_CAPACITY_CONSTRAINT_UP` (:pull:`654`)
 - Add additional oscillation detection mechanism for macro iterations (:pull:`645`)
 
 .. _v3.6.0:
--- a/message_ix/testing/__init__.py
+++ b/message_ix/testing/__init__.py
@@ -424,7 +424,9 @@ def make_dantzig(mp, solve=False, multi_year=False, **solve_opts):
     return scen
 
 
-def make_westeros(mp, emissions=False, solve=False, quiet=True):
+def make_westeros(
+    mp, emissions=False, solve=False, quiet=True, model_horizon=[700, 710, 720]
+):
     """Return an :class:`message_ix.Scenario` for the Westeros model.
 
     This is the same model used in the ``westeros_baseline.ipynb`` tutorial.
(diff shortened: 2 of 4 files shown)


RELEASE_NOTES.rst
```diff
@@ -7,6 +7,7 @@ All changes
 - Adjust :meth:`.Scenario.add_macro` calculations for pandas 1.5.0 (:pull:`656`).
 - Add additional oscillation detection mechanism for MACRO iterations (:pull:`645`).
 - Correct calculation of `COST_NODAL_NET` for standalone MESSAGE (:pull:`648`)
+- Account for difference in period-length when calculating `NEW_CAPACITY_CONSTRAINT_LO` and `NEW_CAPACITY_CONSTRAINT_UP` (:pull:`654`)
```
behnam-zakeri, line 10: maybe a more accurate description: "Account for difference in period-length in equations `NEW_CAPACITY_CONSTRAINT_LO` and `NEW_CAPACITY_CONSTRAINT_UP` (:pull:`654`)"

OFR-IIASA:
@behnam-zakeri thanks for the comments; these have been addressed. I am not planning to add tests. As we generally dont test the GAMS-core formulation, doing so for this feature would imply adding tests not only to test this specific change, but then go on to check for example the soft-constraints and further the therewith associated costs for example - just to be complete. This means the tests would become very extensive and this is something that would need to be addressed in a more structured manner.

gidden:
Hi both - I wonder if it is possible to think of an 'easy' test. For example, we regularly test things like [emissions bounds](https://github.com/iiasa/message_ix/blob/main/message_ix/tests/test_feature_bound_emission.py) and other aspects. 

Is it possible to think of a minimal test where adding a capacity constraint would change the solution?

RELEASE_NOTES.rst
OFR-IIASA: Done

OFR-IIASA:
@gidden there is unfortunately no off the shelf example to which to test this change i.e. we have nothing of the shelf to test change in period duration. This would be something interesting to do in one of the tutorials which in turn can be used for tests. The closest we have at the moment is [westeros_historical_new_capacity.ipynb](https://github.com/iiasa/message_ix/blob/main/tutorial/westeros/westeros_historical_new_capacity.ipynb). This introduces historical new capacity, but our westeros tutorial doesnt have varying timesteps. For now I have added this as a desired feature to add to the basic westeros tutorial suite, which would serve as a starting point to facilitate these type of tests (see issue #662 ).

OFR-IIASA:
@gidden and @behnam-zakeri I have managed to extend the `make_westeros()` testing utility to accept variants of the `model_horizon`. This seems to suite our requirements in terms of testing the changes we have made. I also ran these new tests on the current main branch, where we can see that the results are very different and match the manual calculations I did in excel for both test cases. Hope this is sufficient now. Have a look please and see if anything else is missing. Thanks.

message_ix/testing/__init__.py
```diff
@@ -476,7 +477,25 @@ def make_westeros(mp, emissions=False, solve=False, quiet=True):
...
+                    {"years": missing_years, "values": [np.nan] * len(missing_years)}
+                ),
+            ]
+        )
+        # Interpolate missing years
+        gdp_profile = (
+            gdp_profile.sort_values(by="years")
+            .set_index("years")
+            .interpolate(how="index", limit_direction="both")
+            .reset_index()
```
behnam-zakeri, line 496: It seems there is no need to "reset_index()" here, as in the following line "years" will be set as index again.

message_ix/tests/test_equation_NEW_CAPACITY_CONTRAINT.py
```diff
@@ -0,0 +1,81 @@
...
+import pytest
+from pandas.testing import assert_frame_equal
+
+from message_ix.testing import make_westeros
+
+
+@pytest.mark.parametrize(
+    "model_horizon",
+    [
+        [[700, 710, 715, 720], [3.746712, 4.149163, 8.731826, 9.182337]],
```
behnam-zakeri, line 11: It will be useful to mention what these values are corresponding to each year, and perhaps how they are derived (e.g., calculation steps or showing the equation used for deriving them). I assume these are maximum `CAP_NEW` that can be installed for `wind_ppl` depending on certain input values for "initial-" and "growth_new_capacity_up".

behnam-zakeri (commented):
@OFR-IIASA thanks for this. I had two comments, and I believe this will be ready to merge.

message_ix/tests/test_equation_NEW_CAPACITY_CONTRAINT.py
OFR-IIASA: Thanks Behnam. The formula used to derive these is documented as part of MESSAGEix code so I dont see why we should repeat this here; it is a simple matter of using the "CAP_NEW" equation and the parameters of the technology to derive the values. I think this is relatively straight-forward. I have also stated in the documentation of the test itself, that we are checking the calculation of "CAP_NEW".

message_ix/testing/__init__.py
OFR-IIASA: It is, as this is required to get the dataframe in the same format if no missing years are added i.e. without the "if" statement.

behnam-zakeri (approved):

gidden (approved):
Looks great, @OFR-IIASA - thank you! Awesome that a test could be added as well =)

gidden:
Noting now that all tests pass. I will resolve conflicts and merge.

Issue #575, opened 2022-03-31
Cleaning of `cplex.opt` causes asyncronous issues with multiple runs


gidden:
I have experienced silent errors when performing multiple `message_ix.Scenario.solve()`s simultaneously on the same computer. As best I can tell, these occur when I employ the `solve_options` kwarg, which forces a writing to disk of the `cplex.opt` file. 

Upon completion, I believe this file is then garbage collected. However, other solves occuring simultaneously expect for this file to be present. 

As a short solution, I would suggest we simply do not erase the file. A better solution would employ asyncronous file manipulation.

gidden:
As I look through the code base, it seems like this was a known issue at the time, specifically citing [this](https://github.com/iiasa/message_ix/blob/main/message_ix/models.py#L368) comment.

khaeru:
This may also be a duplicate of #250.

gidden:
closed via #657

Issue #250, opened 2019-10-27
Parallel scenario run raise during remove cplex.opt a FileNotFoundError


tzipperle:
I have figured out a new problem for me. I get the following error some times when I run my scenario parallel runner:

```
 File "D:\Modelle\MESSAGEix-GHD\ghdmod\core.py", line 83, in make_parallel_scenario
    scenario.solve('MESSAGE')
  File "C:\anaconda3\envs\message-ghd-new\lib\site-packages\message_ix\core.py", line 366, in solve
    os.remove(fname)
FileNotFoundError: [WinError 2] Das System kann die angegebene Datei nicht finden: 'C:\\anaconda3\\envs\\message-ghd-new\\lib\\site-packages\\message_ix\\model\\cplex.opt'
````

khaeru:
I moved this issue from ixmp to message_ix, as the code still lives in message_ix.

However, as mentioned earlier, I think the fix here should be both here and in ixmp:

> One thought is that this is related to #141; if `ixmp` gave every `solve()` its own temporary directory, then cplex.opt could be created in that directory and be guaranteed protection from parallel execution.

This would now be handled within the ixmp GAMSModel class. The features for writing the optfile could be moved upstream at the same time.

khaeru:
Marked as a duplicate of #222, which was reported a day earlier.

gidden:
closed via https://github.com/iiasa/message_ix/pull/657

PR #657, opened 2022-10-07
Update default `lpmethod` and remove cplex.opt deletion

OFR-IIASA:
This PR makes two changes:
- The default `lpmethod` is set to "4", therefore changing the default solving option for CPLEX from "DUAL Simplex" to "Barrier".
  For applications based on the MESSAGEix-GLOBIOM model, this solving option should always be used as it is much faster, solving in seconds as opposed to minutes.
- The `cplex.opt` file generated with each solving workflow has traditionally always been removed upon completion. This causes issues for both debugging as well as when running scenarios in parallel (see #575 )

## How to review

## PR checklist

- [x] Update release notes.

### DIFF
--- a/.github/workflows/conda.yaml
+++ b/.github/workflows/conda.yaml
@@ -43,6 +43,7 @@ jobs:
         version: ${{ matrix.conda.version }}
 
     # TODO move the following 2 steps into the above action
+    # TODO avoid using the deprecated ::set-output:: command
     - name: Determine shell on current OS
       id: shell
       run: |
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -32,23 +32,15 @@ jobs:
         path: message_ix
 
     - uses: actions/setup-python@v4
-      # This should match the "Latest version testable on GitHub Actions"
-      # in pytest.yaml
-      # with:
-      #   python-version: "3.10"
+      with:
+        # If the "Latest version testable on GitHub Actions" in pytest.yaml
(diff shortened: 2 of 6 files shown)


message_ix/models.py
```diff
@@ -378,17 +378,7 @@ def run(self, scenario):
...
-        finally:
-            # Remove the optfile regardless of whether the run completed without error.
-            # The file may have been removed already by another run (in a separate
-            # process) that completed before this one.
-            # py37 compat: check for existence instead of using unlink(missing_ok=True)
-            if optfile.exists():
-                optfile.unlink()
-            if optfile2.exists():
-                optfile2.unlink()
+        result = super().run(scenario)
```
gidden, line 381: Can you add a comment here noting that we previously cleaned up the optfile here but consciously choose not to now because it can cause issues when running scenarios asynchronously

behnam-zakeri (commented):
@OFR-IIASA thanks for this. It may be useful to add to the migration notes that choosing different CPLEX method may impact the results, as we observe here for one of our tests. Just to give a heads-up to users.

behnam-zakeri (commented):
This can be merged from my side.

gidden (approved):
LGTM! Thanks @OFR-IIASA . Do we know why the tests are failing? If that is resolved, please go ahead with merge.

khaeru:
Since this branch was updated by merging `main` rather than by rebase, please remember to use the “Rebase and merge” or “Squash and merge” option.

gidden:
closes #575, closes #250

PR #676, opened 2022-11-22
Correct check for unscaled-infeasibilities

OFR-IIASA:
The current check for unscaled-infeasibilities was incorrect.

The status of the GAMS model and CPLEX differ. The check for unscaled-infeasibilities has been updated from `modelstat = 5` to `solvestat = 4`. Recent debugging showed, that when CPLEX solves with unscaled infeasibilities, "LP Status (5)" is displayed in the log of CPLEX, but GAMS translates this to `modelstat=1` and `solvestat=4` (Terminated by Solver). If CPLEX solves normally, then `solvestat=1`.

## How to review

- Build the documentation and check section "macro.html#issue-2-message-solves-optimally-with-unscaled-infeasibilities".

## PR checklist

~- [ ] Add or expand tests; coverage checks both ✅~
- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -13,13 +13,12 @@ Migration notes
 All changes
 -----------
 
-- Correct check for unscaled-infeasibilities to check for `solvestat=4` (:pull:`676`).
+- Add additional oscillation detection mechanism for macro iterations (:pull:`645`, :pull:`676`)
 - Adjust default `lpmethod` from "Dual Simplex" (2) to "Barrier" (4); do NOT remove `cplex.opt` file(s) after solving workflow completes (:pull:`657`).
 - Adjust :meth:`.Scenario.add_macro` calculations for pandas 1.5.0 (:pull:`656`).
 - Ensure `levelized_cost` are also calculated for technologies with only variable costs (:pull:`653`).
 - Correct calculation of `COST_NODAL_NET` for standalone MESSAGE (:pull:`648`)
 - Account for difference in period-length in equations `NEW_CAPACITY_CONSTRAINT_LO` and `NEW_CAPACITY_CONSTRAINT_UP` (:pull:`654`)
-- Add additional oscillation detection mechanism for macro iterations (:pull:`645`)
 
 .. _v3.6.0:
 


gidden:
LGTM, thanks @OFR-IIASA! If there are no objections, I will merge later today.

RELEASE_NOTES.rst
```diff
@@ -13,6 +13,7 @@ Migration notes
 All changes
 -----------
 
+- Correct check for unscaled-infeasibilities to check for `solvestat=4` (:pull:`676`).
```
khaeru, line 16: Because of point (a) in my review, I would rather we remove this line and just update the one existing line below to read like this:
```
- Add additional oscillation detection mechanism for MESSAGE-MACRO iterations; attempt to use `barcrossalg=2` (dual crossover) when MESSAGE solves with unscaled infeasibilities (:pull:`645`, :pull:`676`).
```
The reason: from the point of view of a user using only the stable releases, there was never a release where there was a check for unscaled infeasibilities using something other than `solvestat=4`. So, there is nothing to be "Correct"ed (per the added line). Compared to v3.6.0, this is a new feature that will appear in v3.7.0 as the result of these two PRs.

khaeru (approved):
Some context was missing from the PR description:

- I went to look at the blame: https://github.com/iiasa/message_ix/blame/main/message_ix/model/MESSAGE-MACRO_run.gms#L315
- It appears this line was added in **#645** also by @OFR-IIASA. ←
- Prior to that, there was no such check: https://github.com/iiasa/message_ix/blame/0c6f0759d86de002b65739e476c3e319840b5c20/message_ix/model/MESSAGE-MACRO_run.gms (Ctrl+F "unscaled")
- Thus:
  - (a) Because #645 is within the same release, we don't have to consider whether this regresses somehow on behaviour that was provided in a past release.
  - (b) Because this is the same author, we don't have to worry whether this was put there ~5 years ago by someone else for some purpose we have since forgotten.

Supplying this information (the point with ←) would have made this quicker to review.

See also one suggestion on the release notes.

RELEASE_NOTES.rst
gidden: done!

PR #677, opened 2022-11-23
Work around conda-forge/click-feedstock#45

khaeru:
Experimental. See conda-forge/click-feedstock#45.


khaeru:
NB this problem resolved itself somewhere upstream over the holiday break. Will close this PR without merging, except that the first commit to satisfy mypy should be preserved somewhere.

khaeru:
> Will close this PR without merging, except that the first commit to satisfy mypy should be preserved somewhere.

Mypy seems okay with the code currently on `main`, so I conclude that first commit is no longer needed.

Issue #691, opened 2023-02-28
Add a citation file


danielhuppmann:
Please add a citation.cff, see https://citation-file-format.github.io/cff-initializer-javascript

I suggest to use all (non-marginal) contributors as authors, and add the MESSAGEix manuscript (Huppmann et al., 2019, as a "related resource".

This is related to a request from the Horizon Europe Project PRISMA by @tscheypidi

fyi @khaeru @volker-krey

glatterf42:
Where do we draw the line for contributors being 'marginal'? At least ten commits? This would exclude Gamze, but only for now. If she makes one more commit, we'd have to adapt this file. We could simply include everyone from https://github.com/iiasa/message_ix/graphs/contributors . I think it's only fair and doesn't hurt anyone.

glatterf42:
Generally, I have not included any email addresses. I have tried to look people up on Orcid and added IDs whenever I found them. Same names still made trouble, though: with Miguel Poblete Cazenave, I have found Poblete as a surname, Cazenave as a surname, and Poblete-Cazenave as a surname, so currently, I set up Poblete as a second given name. 
For the contributor lumbric, I couldn't find any name, so excluded them from the file.
Other notes I took in the process:
- Zenodo doesn't update itself automatically, so while e.g. I am a contributor now on Github, I'm not listed as such on Zenodo 
- We might want to update version-specific stuff whenever that changes/becomes important
- With the doi as I have extracted it, one finds a 2018 version of the paper under some semi-obscure link. Which doi should be used?

PR #695, opened 2023-03-02
Add citation file

glatterf42:
Add a citation file of format CITATION.cff to address issue #691.

## How to review

To review, please
- Check that the information in the file is corrected
- Check that it is formatted correctly (i.e. behaves as is intended)

## PR checklist

- ~Add or expand tests; coverage checks both~ No change in behaviour.
- [x] Update release notes.


CITATION.cff
```diff
@@ -0,0 +1,181 @@
+# This CITATION.cff file was generated with cffinit.
+# Visit https://bit.ly/cffinit to generate yours today!
+
+cff-version: 1.2.0
+title: MESSAGEix
+message: >-
+  Thank you for using the MESSAGEix framework! Please cite
+  it using the metadata from this file :)
```
phackstock, line 8: I like it but maybe leave out the smiley face ;)
```suggestion
  it using the metadata from this file
```

CITATION.cff
```diff
@@ -0,0 +1,181 @@
...
+keywords:
+  - integrated assessment
+  - energy systems
+  - macro-energy
+  - modelling
+  - scenario analysis
+  - data visualisation
+  - Python package
+license: Apache-2.0
+commit: 139c832a9fc195de97e2b4b911461fb49b6fa455
```
phackstock, line 127: What does the commit attribute do? Does this get updated with every release then?

CITATION.cff
phackstock, line 8: Since you put the original 2019 MESSAGEix article in the references, should that one be cited as well or only the DOI of this repo?

CITATION.cff
```diff
@@ -0,0 +1,181 @@
...
+    family-names: Steinhauser
+    affiliation: IIASA
+    orcid: 'https://orcid.org/0000-0002-5989-6855'
+  - given-names: Clara
+    family-names: Orthofer
+    orcid: 'https://orcid.org/0000-0002-7049-7791'
+  - given-names: Thomas
+    family-names: Zipperle
+    orcid: 'https://orcid.org/0000-0003-2512-9508'
+  - given-names: 'Nikolay '
```
phackstock, line 71:
```suggestion
  - given-names: Nikolay
```

CITATION.cff
```diff
@@ -0,0 +1,181 @@
...
+  - given-names: Volker
+    family-names: Krey
+    affiliation: IIASA
+    orcid: 'https://orcid.org/0000-0003-0307-3515'
+  - given-names: Adriano
+    family-names: Vinca
+    affiliation: IIASA
+    orcid: 'https://orcid.org/0000-0002-3051-178X'
+  - given-names: Gamze
+    family-names: ' Ünlü'
```
phackstock, line 44:
```suggestion
    family-names: Ünlü
```

phackstock (commented):
Looks pretty good to me, a few small comments below in line.

CITATION.cff
```diff
@@ -0,0 +1,181 @@
...
+  - given-names: Michael
+    family-names: Pimmer
+  - given-names: Peter
+    family-names: Kolp
+    affiliation: IIASA
+    orcid: 'https://orcid.org/0000-0003-0122-2839'
+identifiers:
+  - type: url
+    value: 'https://github.com/iiasa/message_ix'
+    description: Main repository of hte MESSAGEix framework on Github
```
danielhuppmann, line 86:
```suggestion
    description: Main repository of the MESSAGEix framework on Github
```

CITATION.cff
```diff
@@ -0,0 +1,181 @@
...
+    affiliation: IIASA
+    orcid: 'https://orcid.org/0000-0003-0080-7122'
+  - given-names: Miguel Poblete
+    family-names: Cazenave
+    affiliation: IIASA
+  - given-names: Jarmo S.
+    family-names: Kikstra
+    affiliation: IIASA
+    orcid: 'https://orcid.org/0000-0001-9405-1228'
+  - given-names: Fridolin
```
danielhuppmann, line 54: Please create an ORCID for yourself, @glatterf42

CITATION.cff
```diff
@@ -0,0 +1,181 @@
...
+      family-names: Riahi
+      affiliation: IIASA
+      orcid: 'https://orcid.org/0000-0001-7193-3498'
+    - given-names: Volker
+      family-names: Krey
+      affiliation: IIASA
+      orcid: 'https://orcid.org/0000-0003-0307-3515'
+    title: "The MESSAGEix Integrated Assessment Model and the ix modeling platform (ixmp): An open framework for 
+integrated and cross-cutting analysis of energy, climate, the environment, and sustainable development"
+    doi: "10.1016/j.envsoft.2018.11.012"
```
danielhuppmann, line 173: Are you sure that this shouldn't include the https://doi.org/ part?

CITATION.cff
```diff
@@ -0,0 +1,181 @@
...
+    - given-names: Keywan
+      family-names: Riahi
+      affiliation: IIASA
+      orcid: 'https://orcid.org/0000-0001-7193-3498'
+    - given-names: Volker
+      family-names: Krey
+      affiliation: IIASA
+      orcid: 'https://orcid.org/0000-0003-0307-3515'
+    title: "The MESSAGEix Integrated Assessment Model and the ix modeling platform (ixmp): An open framework for 
+integrated and cross-cutting analysis of energy, climate, the environment, and sustainable development"
```
danielhuppmann, line 172:
```suggestion
    title: >-
      The MESSAGEix Integrated Assessment Model and the ix modeling platform (ixmp):
      An open framework for integrated and cross-cutting analysis of energy, climate,
      the environment, and sustainable development
```

CITATION.cff
danielhuppmann: Yes, please use a statement like "If you use the MESSAGEix framework, please refer to the GitHub repository and cite the MESSAGEix framework manuscript (https://doi.org/10.1016/j.envsoft.2018.11.012)"

danielhuppmann (commented):
Thanks, going in the right direction!

CITATION.cff
glatterf42: Yes: https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsdoi

CITATION.cff
glatterf42: Not automatically, no. So I'm not sure it's worth the effort. On the other hand, I thought the Horizon people might find it neat and maybe it won't hurt if it's a bit outdated occasionally afterwards.

CITATION.cff
glatterf42: Aw :(

CITATION.cff
danielhuppmann: I'd rather remove this. We already have a clear version-and-release-management, this doesn't need to be duplicated (and maybe create inconsistencies) in the citation.cff

glatterf42: Okay, will do :)

glatterf42:
Do we need to update https://docs.messageix.org/en/stable/notice.html or the corresponding NOTICE.rst in the main repo? The information there is still correct, we might still want to add a section about using our new file for citation. I don't yet know how to do that, but I could read up on it. However, if you want it done asap and have worked with cff files before, you might want to take on that part.

danielhuppmann:
Good idea @glatterf42! I suggest to add a reference/link to the citation file as an item in the list of section 2.

phackstock (approved):
Looks all good to me.
One small final side note, maybe you could write somewhere that this file needs to be updated with every release as it contains a direct reference to the version number.

CITATION.cff
```diff
@@ -0,0 +1,181 @@
...
+      orcid: 'https://orcid.org/0000-0002-5611-7780'
+    - given-names: Keywan
+      family-names: Riahi
+      affiliation: IIASA
+      orcid: 'https://orcid.org/0000-0001-7193-3498'
+    - given-names: Volker
+      family-names: Krey
+      affiliation: IIASA
+      orcid: 'https://orcid.org/0000-0003-0307-3515'
+    title: "The MESSAGEix Integrated Assessment Model and the ix modeling platform (ixmp): An open framework for integrated and cross-cutting analysis of energy, climate, the environment, and sustainable development"
```
danielhuppmann, line 172: I guess that the order doesn't really matter for the off format - so I'd first have the title, journal, year, then the authors, then the other attributes...

CITATION.cff
```diff
@@ -0,0 +1,181 @@
...
+  - given-names: Peter
+    family-names: Kolp
+    affiliation: IIASA
+    orcid: 'https://orcid.org/0000-0003-0122-2839'
+identifiers:
+  - type: url
+    value: 'https://github.com/iiasa/message_ix'
+    description: Main repository of the MESSAGEix framework on Github
+  - type: doi
+    value: 10.5281/zenodo.7006303
```
danielhuppmann, line 90: You used the version-specific DOI - however, that DOI is only created AFTER a release, so it would not be possible to have the version-number and version-specific DOI correct at the time of the release.

Better to use the version-independent DOI (always refers to the latest version).
```suggestion
    value: 10.5281/zenodo.4005684
```

danielhuppmann (changes requested):
Having the version hard-coded in the citation file means that this has to be changed explicitly as part of the release-PR. In pyam, there is a specific [Release Procedure](https://github.com/IAMconsortium/pyam/blob/main/RELEASE_PROCEDURE.md) - does this exist for message_ix as well?

Please add an reminder to change the citation.cff to the release-procedure document - or better to not state the version explicitly in this file at all.

cc @khaeru

glatterf42:
I think it might be useful to include some version information. Future versions might not be able to reproduce results, so project scientists might want to pinpoint specific versions as the ones that worked for them. On that regard, I've found a [Github Action to automatically update the citation file](https://github.com/ropensci/cffr/blob/main/.github/workflows/update-citation-cff.yaml), so we should be able to include this information without needing to worry about manual updates. The only downside is that I haven't really worked with Github Actions yet, so I'll have to spend some time understanding how they work, but then I should be able to adapt this workflow to our needs. 
The question would then be: do we want a whole version only or also single commits as they are more accurate?

danielhuppmann:
But the version-information is already included in your workflow/script simply by doing `message_ix.version` (including the commit hash if you're in a non-release state) - I'm very skeptical of duplicating this information.

glatterf42:
On that note, the [pyam citation file](https://github.com/IAMconsortium/pyam/blob/4aa433a54c8e42bec9389a9a67f24e811039fe31/CITATION.cff#L5) also contains a version keyword, referencing version 1.0.
I think this is again a matter of what the use case is: I thought we wanted to include this citation file so that people can simply cite MESSAGEix correctly from the main repo e.g. by extracting a Bibtex entry. This entry will only include version information if we put it in the citation file, if I understand correctly.

danielhuppmann:
> On that note, the [pyam citation file](https://github.com/IAMconsortium/pyam/blob/4aa433a54c8e42bec9389a9a67f24e811039fe31/CITATION.cff#L5) also contains a version keyword, referencing version 1.0.

True, but the pyam-citation refers to the manuscript in **Open Research Europe**, which is specifically linked to v1.0 of the package. But this could obviously lead to some confusion there, too...

glatterf42:
So: in part, I just wanted to see if I could do it. Even if we don't end up using what I cooked up here, I have learned a lot about Github Actions in the process :)
Despite your explanation above, I still think we should include the version information in the citation file. Since we are going to provide one either way, I think this will make citing `MESSAGEix` much more comfortable. However, in order to be accurate, we then also need to include the commit that the citation is referring to. A current example is that `pandas 2.0.0` will abandon some features that are still in use in version 3.6.0 of `MESSAGEix`, but have been fixes by a more recent commit. Thus, citing version 3.6.0 of `MESSAGEix` will lead to people being unable to reproduce results because they won't have the fixes included that every recent clone of the main repo has.
Updating the version and commit manually would be tedious, which is why I have written a Github workflow to automatically update them. With this workflow, every new release would trigger another commit that updates the citation file to include the newest version. Each push to the main branch would update the latest commit. 
As @phackstock pointed out, this would lead to people not having the latest version of the citation file locally available even if they just pushed to the main branch without running `git pull`. So, if they were to use the local contents of the file, they might produce an incorrect citation. However, I don't know how frequently this would actually happen. If I understand correctly, the main purpose of CITATION.cff is to provide a citation option directly in the main Github repo. Anyone wanting to cite `MESSAGEix` might therefore visit our main repo and create e.g. a Bibtex entry with the up-to-date information from there. Furthermore, I think `git pull` is a quicker alternative. If people want to cite the code locally (and don't already have a script using `message_ix.version`), they'd either have to write such a script or simply run `git pull`. And if they already have such a script, the information in there will still be correct.

All in all, I think providing the version and commit information is comfortable, accurate, and not much work with the addition of the workflow.

cc @khaeru

glatterf42:
Together with @khaeru, I just noticed that the automatically created Bibtex entry does not contain a lot of information. For starters, it's just one entry, so we can choose for it to be the software entry _or_ the paper entry. Even if we choose the software entry, the only versioning information would be month and year of the "date-released" entry. So I also don't think there's a point to keep them automatically up-to-date for this feature. Maybe [some of the other tools](https://github.com/citation-file-format/citation-file-format#tools-to-work-with-citationcff-files-wrench) could make use of the data, but for now, I removed them from the citation file. 
Since I still think the version and commit should be mentioned when citing software, I will adapt the corresponding docs section with how to do that.

PR #700, opened 2023-03-27
Improve depiction of Latex text in docs

glatterf42:
This PR simply writes `\text{xyz}` in Latex math mode in the docs where before it used to be `xyz`. This should improve [the time page](https://docs.messageix.org/en/stable/time.html) and [the efficiency page](https://docs.messageix.org/en/stable/efficiency.html).

## How to review

- Build the documentation and look at the changed pages

## PR checklist

### DIFF
--- a/doc/README.rst
+++ b/doc/README.rst
@@ -21,16 +21,16 @@ There are a number of guides out there, e.g. on docutils_.
 Building the docs on your local machine
 ---------------------------------------
 
-From the command line, run::
+From the command line, run in `doc/`::
 
     make html
 
 You can then view the site by::
 
     cd build
-    python -m SimpleHTTPServer
+    python -m http.server
 
-and pointing your browser at http://localhost:8000/html/
+and pointing your browser at http://0.0.0.0:8000/_build/html/
 
 
 Read the Docs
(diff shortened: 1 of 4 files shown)


glatterf42:
On the time page, please check that the equations in [example 5](https://docs.messageix.org/en/stable/time.html#discounting) are correct. What I'm wondering about in particular is why the discount factor for 1020 should not contain any information of the years 1000 to 1010, when in the first part it is explained that the discount should be cumulative. There are two options I can see here: either `df_1020` should be based on `df_1010`, in which case the years should be calculated by (1010-1011) etc and `df_1010` itself should appear in the equation. Or it should be cumulative from 1000 to 1020, in which case the equation should be twice as long (i.e., start with (1000-1001)). 
I briefly looked into the logic behind these equations in general, the present value formula is new to me, but I didn't have the time yet to check if the calculation is done correctly (i.e., is the discount factor defined as PV/C following the definiton [on wikipedia](https://en.wikipedia.org/wiki/Present_value)?). However, based on the success of MESSAGEix in general, I assume that is the case.

Also, SimpleHTTPServer seems to have moved to [http.server](https://docs.python.org/3/library/http.server.html), so I adapted the local-build instructions for the docs.

PR #494, opened 2021-07-18
Test & improve perfomance in .tools.add_year

khaeru:
Most (137 of 187) untested lines in `message_ix` are in `.tools.add_year`: https://codecov.io/gh/iiasa/message_ix/tree/master/message_ix/tools/add_year

This PR adds tests for this code, but also reimplements some internal methods for better performance.

## How to review

**To be added.**

## PR checklist

- [ ] Continuous integration checks all ✅
- [ ] Add or expand tests; coverage checks both ✅
- [ ] Add, expand, or update documentation.
- [ ] Update release notes.

PR #702, opened 2023-04-06
Fix inconsistencies in docs of special time slice value Year

### LINKED ISSUE #701: Documentation of subannual timeslices erroneous
Normally, all values intended for use in the `subannual` category need to be registered on a platform first with `ixmp.Platform.add_timeslice()`. However, this is not the case for the special value `Year`, which indicates that the surrounding timeseries is actually valid for the whole year. This parameter is case sensitive, though, which is not always clear from the documentation. This issue, then, is supposed to fix all occurrences of `year` and `YEAR` as the special values present without prior registration.

glatterf42:
Closes #701 by including minor changes in the documentation.

## How to review

- Read the diff and note that the CI checks all pass
- Build the documentation and look at the page(s) that changed

## PR checklist


glatterf42:
The tests on python 3.8 failing here is a known issue and #494 will resolve them once it is merged.

Issue #409, opened 2020-09-30
Assignment of `capacity_factor` is not properly working for subannual time slices


behnam-zakeri:
In `message_ix\model\MESSAGE\data_load.gms`, there are two equations (see the links below) related to the assignment of `capacity_factor`:
1. The first equation assigns `capacity_factor = 1` to the missing entries for `capacity_factor` in any time slice. 
https://github.com/iiasa/message_ix/blob/6f3d0b2daa6ad8421fd045fd51962016e17ec3b4/message_ix/model/MESSAGE/data_load.gms#L141

2. The second equation calculates an average yearly value for `capacity_factor`:
https://github.com/iiasa/message_ix/blob/6f3d0b2daa6ad8421fd045fd51962016e17ec3b4/message_ix/model/MESSAGE/data_load.gms#L145

The problem is that, in a scenario with sub-annual time slices, e.g., two seasons of _winter_ and _summer_; when a technology is active only in sub-annual time slices, it has no `capacity_factor` defined for time `year`. As such, Eq. No. 1 above adds a `capacity_factor` with the value of 1 for `year`. Then, Eq. No 2 above makes an average yearly capacity factor by considering all time slices, including winter, summer, and `year`. This results in an average `capacity_factor` higher than 1, which is not correct.

To tackle this issue, it seems the solution is to reverse the order that these two equations appear in the formulation, i.e.,:
- first caculating the average yearly `capacity_factor`, and then,
- if there is no capacity_factor for `year`, assiging the value of 1 to that.

khaeru:
@behnam-zakeri could you help us interpret here:
- Is this the same issue as #515, or distinct?
- Will it be fixed by #561, or should remain open?

behnam-zakeri:
@khaeru thanks for picking this. No, this is a separate issue. Indeed, this will remain an issue even after the improvements in #561 . So, it should remain open.

behnam-zakeri:
However, now that we have tests for `capacity_factor`, it should be possible to resolve this and test it. I will give it a try, but perhaps after the release.

khaeru:
Thanks! Adjusted the milestone accordingly.

glatterf42:
[...]

PR #705, opened 2023-04-17
Correct GAMS for the assignment of "capacity_factor" at "year"

behnam-zakeri:
This PR resolves issue #409. The GAMS formulation is corrected not to assign a capacity factor of 1 at "year" if the value is missing in a scenario with sub-annual time slices. The weighted average of capacity factors defined for sub-annual time slices can be used to calculate the capacity factor for "year". This is tested by extending the existing tests in `test_capacity_factor_feature`, i.e., adding data for parameter "operation_factor", which uses capacity factor at "year".

## How to review
The test `test_capacity_factor_feature` should pass with the GAMS formulation in this PR. However, this test must fail if using GAMS formulation of the main branch of `message_ix`. So, first, copy this test to your main branch and ensure it fails. Then, switch to this branch and run the tests again and make sure they pass.

## PR checklist
- [x] Update release notes.


glatterf42:
Can confirm: on my main branch, the new test fails with the following error message: `FAILED message_ix/tests/test_feature_capacity_factor.py::test_capacity_factor_average - assert 3.0 == 2.8` 
At the same time, the test succeeds on the new branch. 

While this looks good, I'm also receiving a new RuntimeWarning:
```
message_ix/tests/test_feature_capacity_factor.py::test_capacity_factor_zero_two
home/glatterf42/message_ix/message_ix/tests/test_feature_capacity_factor.py:27: RuntimeWarning: invalid value encountered in double_scalars
    act.loc[i, "cf-corrected"] = act.loc[i, "lvl"] / float(
```
I will try to understand where that comes from.

message_ix/tests/test_feature_capacity_factor.py
```diff
@@ -7,29 +7,39 @@
...
+        # Correct CF based on duration of each time slice
         duration = float(scen.par("duration_time", {"time": i[1]})["value"])
-        act.loc[i, "duration-corrected"] = act.loc[i, "lvl"] / duration
-        # Divide by (non-zero) capacity factor
-        act.loc[i, "cf-corrected"] = act.loc[i, "duration-corrected"] / float(
-            cf.loc[i, "value"]
-        )
+        if i[1] != "year":
+            cf.loc[i, "duration-corrected"] = cf.loc[i, "value"] * duration
+            act.loc[i, "cf-corrected"] = act.loc[i, "lvl"] / float(
```
glatterf42, line 27: This seems to cause the `RuntimeWarning`. I'm wondering about `i` and the hardcoded index `1`: is that always correct or does it assume a certain structure of `i`? If so, can we change to using no assumption?

glatterf42 (commented):
Looks good to me, but I'd like to understand this `RuntimeWarning` before approving.

message_ix/tests/test_feature_capacity_factor.py
glatterf42: The problem arises in the `test_capacity_factor_zero_two`, where the `solar_pv_ppl` has a `capacity_factor` of `0` in the time slice `night`. In
```
if i[1] != "year":
            cf.loc[i, "duration-corrected"] = cf.loc[i, "value"] * duration
            act.loc[i, "cf-corrected"] = act.loc[i, "lvl"] / float(
                cf.loc[i, "duration-corrected"]
            )
```
this gives us a division by 0 since `cf.loc[i, "value"]` is 0. So we have to provide some condition/prescription on how to deal with cases like this.

message_ix/tests/test_feature_capacity_factor.py
```diff
@@ -7,29 +7,39 @@
...
+                cf.loc[i, "duration-corrected"]
+            )
     act = act.fillna(0).reset_index().set_index(["technology"])
-    # CAP = max("ACT" / "duration_time" / "capcity_factor")
+    # CAP = max("ACT" / "duration_time" / "capcity_factor") / "operation_factor"
```
glatterf42, line 31:
```suggestion
    # CAP = max("ACT" / "duration_time" / "capacity_factor") / "operation_factor"
```

message_ix/tests/test_feature_capacity_factor.py
glatterf42: I've now added a condition that checks if either `duration` or `cf.loc[i, "value"]` are `0`. In that case, we now set `act.loc[i, "cf-corrected"]` to `act.loc[i, "lvl"]`, so we don't change it. This relies on the assumption that in this case, `act.loc[i, "lvl"]` will also be 0, which is implied (and asserted) by the test in line 19. 
All the tests are still passing for me, but there's no `RuntimeWarning` anymore.

behnam-zakeri: Thanks @glatterf42 for reviewing this and suggesting improvements. Indeed, as you noted, zero capacity factors are tested and they can be excluded from the follow-up calculations. The index `i` in that loop is a tuple with two members, the first one being the technology name and the second time slices. So, the index `1` is always related to time.

PR #707, opened 2023-04-27
Update "conda" CI workflow

khaeru:
TBA.

## How to review

Note the CI checks all pass.

## PR checklist

- [ ] Continuous integration checks all ✅
  
- [ ] Add or expand tests; coverage checks both ✅
- [ ] Add, expand, or update documentation.
- [ ] Update release notes.

PR #710, opened 2023-05-12
Update versions of conda and miniconda in workflow conda.yaml

glatterf42:
The conda-forge workflow is currently failing on some windows systems, which might simply be due to using 2021 versions of anaconda and miniconda.

## How to review

- After the workflow file is updated, trigger the workflow and check that it works for all systems

## PR checklist


glatterf42:
This is a duplicate of 
- #707 
It can thus be closed.

PR #715, opened 2023-05-30
Use mamba solver instead of conda solver

glatterf42:
~Since we pinned the message-ix version to help the conda workflow pass, we need to keep this version number up-to-date.~

The above was true in principle, but did not solve our CI conda-forge test problems. As suggested in conda/conda/issues/12741, replacing the conda solver with the mamba solver did. This even allows us to remove the message_ix version pin entirely.

## How to review

## PR checklist

### DIFF
--- a/.github/workflows/conda.yaml
+++ b/.github/workflows/conda.yaml
@@ -62,13 +62,15 @@ jobs:
 
     - name: Configure conda channels, update, create environment, and install message-ix
       run: |
+        conda install -n base conda-libmamba-solver
+        conda config --set solver libmamba
         conda config --prepend channels conda-forge
         conda config --set channel_priority strict
         conda update -n base -c defaults conda
 
         # Also install pytest and packages required for testing. pip/PyPI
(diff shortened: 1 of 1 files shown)


khaeru (commented):
Please:
1. Uncomment these lines: https://github.com/iiasa/message_ix/blob/4663f1d6dfe352a4638624047bac0687b7b371d9/.github/workflows/conda.yaml#L9-L11
2. Commit with a message like "TEMPORARY …"
3. Push.
4. Confirm that the workflow passes with this change, e.g. by leaving a comment or updating the PR description like "Note here that the jobs all pass with this change."
5. Drop the temporary commit from the branch, e.g. `git reset --hard HEAD~1 && git push --force`.

((5) isn't necessary if the PR will be squash-merged.)

glatterf42:
The tests all pass using the mamba solver. Thus, I will 

- Change the description of this PR
- Reset all commits and introduce just the changes we want to keep
- Open a new issue to deal with the cache saving problem

PR #721, opened 2023-06-20
Improve depiction of Latex in GAMS-created docs

glatterf42:
This PR puts all  strings with more than two characters in `\text{}`, which should help the depiction of long formulae in particular. Note that some single characters have seen the same fate, especially in `macro_core.gms`, because it would have appeared inconsistent to have some parameters in a table be in `\text{}` and not others if they represent the same thing thematically.

## How to review

- Build the documentation and look at the 'Mathematical Specification' pages.

## PR checklist

- [x] ~Continuous integration checks all ✅~
  Except for `macos-latest-py3.7` per iiasa/ixmp#484.
- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -1,8 +1,10 @@
-.. Next release
-.. ============
+Next release
+============
 
-.. All changes
-.. -----------
+All changes
+-----------
+
+- Improve readability of LaTeX equations in docs (:pull:`721`).
 
 .. _v3.7.0:
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -138,6 +138,8 @@ intersphinx_mapping = {
 
 # -- Options for sphinx.ext.mathjax ----------------------------------------------------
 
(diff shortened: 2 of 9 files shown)


khaeru (approved):
Thanks for thoroughly going through this, @glatterf42.

Approved, with one minor request:

> - [ ] Update release notes.

I think it's worth advertising this, as IMHO it's a significant improvement in readability. It could be like:
```
- Improve readability of LaTeX equations in docs (:pull:`721`).
```

Some other notes for posterity:
- We had a discussion in Slack about the merits of using `\text{…}` explicitly (e.g. `\text{beyond_horizon_factor}` versus expanding this code in the Sphinx config https://github.com/iiasa/message_ix/blob/3cd004a2bee0d2684111e6a94809091b63358476/doc/conf.py#L141-L157 which would define a TeX commands (e.g. `\beyondhorizonfactor`). We decided for the former for now a variety of reasons.
- #722 is related but (IMHO properly) kept separate and not scheduled for now.

PR #743, opened 2023-09-18
IGNORE: Carbon pool [for building docs]

gidden:
This is a clone of #514 added in order to render docs on RTD


gidden:
And RTD fails to build: 
image

khaeru:
@glatterf42 added the "build.os" key to .readthedocs.yaml in #733. I would guess the `carbon_pool` branch is not up to date.

Issue #731, opened 2023-08-01
Handling flaky tests


glatterf42:
For roughly the past month, I have collected data on flaky CI tests. The initial idea was to mark them as [flaky](https://github.com/box/flaky), but as per [pytest's docs on flaky tests](https://docs.pytest.org/en/stable/explanation/flaky.html), that should never be a long-term solution. Instead, tests should be (randomly) re-ordered, re-written for more atomic assertions, or split up into different groups to find the root cause of the flaky behavior and eliminate it. We will have to see when time permits this. For now, we could mark them as flaky to save us from re-running them manually. 
Here are the flaky tests of this repository I gathered so far:

# Flaky tests

## Auxiliary

Some error message are common and so long that they would render the table even more complex than it already is. They are included here and referenced by their names in the table below.

### Notebook cell timeout prepare reporter {#notebook-cell-timeout-prepare-reporter}
```python
nbclient.exceptions.CellTimeoutError: A cell timed out while it was being executed, after 10 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
['# Prepare a Reporter based on the solved scenario', 'ctax_rep = Reporter.from_scenario(ctax_scen)', 'prepare_plots(ctax_rep)', '', '# Run the same reporting on both the base and modified scenarios']
...
['    r.get("plot new capacity")', '    r.get("plot capacity")', '    r.get("plot activity")', '    r.set_filters(t=None, c=["light", "other_electricity"])', '    r.get("plot prices")']
-------------------
```

### Notebook cell timeout import packages {#notebook-cell-timeout-import-packages}
```python
nbclient.exceptions.CellTimeoutError: A cell timed out while it was being executed, after 10 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
# Importing required software packages
import pandas as pd
import ixmp
import message_ix

from message_ix.utils import make_df

%matplotlib inline
-------------------
```

Names are shortened to `message_ix/tests` as the starting directory. Grouping the tests by names shows commonalities between them. Numerical errors seem to appear exclusively for macos, while ubuntu only sports errors in testing the tutorials, almost all of which are similar to one another. Windows appears only once when it was unable to connect to the database (these two errors are related, see also ). 

| Test name    | Error message | Runners (# of occurrences if > 1) |
|--------------|:-------------:|---------------------------:|
[...]

PR #765, opened 2023-12-11
Record message_ix version in scenario data & GDX

### LINKED ISSUE #739: Split tutorial tests to separate CI job
The "pytest" CI workflow takes up to 20 minutes (Windows) to run, which makes it less useful for its intended purpose of quick feedback on PRs. Most of the 20 longest-duration tests are tutorial tests.

One way to address this:
- Add a second job named "tutorials" within the pytest.yaml workflow file.
- In this job, run only the tests of the tutorials. This might be done for only one Python version.
- In the "pytest" job, omit the test of the tutorials.

Other benefits: we could upload Codecov reports for only the "pytest" workflow. This would help us see if there is any code that is currently tested *only* through its use in a tutorial, and add more direct (faster, simpler) tests.

### LINKED ISSUE #747: Record message_ix version in scenario data & GDX I/O
The GDX file generated by ixmp.JDBCBackend has a parameter named `MESSAGE_ix_version` with values like {"major": 2, "minor": 0}. The original intent of this was to aid with debugging when inspecting/comparing GDX files.

This has not been updated for recent MESSAGE versions, for multiple reasons:
- It is a "stack level violation", in other words in the wrong place: ixmp (or its backends) should be agnostic to whether they are being used by message_ix or not; similarly, logic relevant to message_ix should be actually _in_ message_ix.
- The team lacks a Java developer capable of updating the code and the JAR file in ixmp.

Raised by @volker-krey in the [2023-10-19 MESSAGEix meeting](https://iiasahub.sharepoint.com/sites/eceprog/_layouts/15/Doc.aspx?sourcedoc={2a354107-351e-43ab-8824-542f007aa2ff}&action=edit&wd=target%282023.one%7C010d73b1-6bc1-492d-b32a-c6143fe2b0f2%2F10-19%20Monthly%7Cba547214-3f13-463d-8a64-2a2472bb19da%2F%29&wdorigin=703).

## Fix

- We can use the code underlying `show_versions()` to generate a string with more informative version information (like " 3.7.1.dev55+g2e0a9c49"), and store this in a parameter.
- This can be updated in `message_ix.Scenario.solve()`.
- The name should be something other than "MESSAGE_ix_version" (perhaps just "version"?) in order to avoid a collision with the (fixed) behaviour of the Java code.

khaeru:
Closes #747.

Housekeeping:
- Close #739. This appears to bring the job time down to:
  - 4:40–10 minutes for the tests —so the first indication of success/failure comes in under 5 minutes, which is more fit for purpose.
  - 8–12 minutes for the tutorials.
- [x] Update older tests. The above change caused some of these to begin to fail, which makes me suspect they rely on state leaking from the tutorial tests or are otherwise fragile. The updates should ensure the tests are complete.
  - `test_equation_NEW_CAPACITY_CONSTRAINT.py::test_new_capacity_up`: fix typo in the file name. Clone the subject scenario to a unique URL, so that GDX I/O files do not collide even if the two test cases are run in parallel.
  - `test_integration.py`, `test_macro.py`: similarly clone to unique URLs.
  - Expand some flaky marks from #731 that were previously scoped to macOS only, to now apply to all platforms:
    - `test_feature_bound_activity_shares.py::test_add_bound_activity_up_all_modes`
    - `test_feature_price_emission.py::test_custom_type_variable_periodlength`
    - `test_integration::test_multi_db_run`.
- Advertise and test Python 3.12 support.
- Update copyright year to 2024.

## How to review

## PR checklist

- [x] Merge required changes upstream: iiasa/ixmp#502.
  
- [x] Update release notes.


khaeru:
@glatterf42 FYI, looking at the recent commits on the branch, I am finding that for some apparently flaky tests, it helps to `.clone(scenario=request.node.name)` first.

I think this is because we are using pytest-xdist: where this results in 2+ tests with identical model name and scenario names being run in parallel, then strange effects can occur; for instance, one test reads and sees the result from the GDX file of the other test. (This occurs because message_ix tries to heed users' desire to keep GDX I/O files all in the same directory, rather than separate, temporary directories per-run as the base ixmp GAMSModel class does.)

[`request.node.name`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.FixtureRequest.node) is from the Pytest `request` fixture, and gives a string that is unique to each test—including individual cases of parametrized tests. It's the easiest way to get such a unique label.

We could later (no urgency) try this approach to address other apparently flaky tests.

glatterf42:
Very useful catch, would be great to remove some flaky markers :)

glatterf42:
Just rebasing onto current main.

glatterf42 (approved):
LGTM, thanks :)

glatterf42:
Rebased after dependabot update to main: are we just seeing flakiness in these tests?

glatterf42:
Looks like flakiness; are we merging regardless and clean up/apply markers later if needed?

khaeru:
> are we merging regardless and clean up/apply markers later if needed?

I would say the latter.

I have adapted several tests in this PR to ensure they use unique model/scenario names; this means there is no chance that another scenario being run in another pytest-xdist worker thread will try to access the same scenario in the database or in a GDX file. However, this is different from doing a full sweep through the test suite or adapting fixtures to be sure that *no* test has this fragility. We can do that in a subsequent PR.

PR #766, opened 2023-12-13
Add warning in docs about #723

glatterf42:
Since we won't be able to merge #723 in time, this PR adds a warning in the docs to where `PRICE_EMISSION` [is defined](https://docs.messageix.org/en/latest/model/MESSAGE/model_core.html#auxiliary-variables). This will allow us to postpone the fix for 3.8.1 in January.

I'm actually not sure: should we update the release notes?

## How to review

- Check the documentation built by the RTD check below and look at MESSAGE core formulation -> Variable Definitions -> Auxiliary Variables.

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ Only docs
- [x] Update release notes.


khaeru (approved):
Looks good! I don't think we need to add it to the release notes.

PR #770, opened 2024-01-07
Add `.tools.sankey` and tutorial

mabudz:
Create sankey diagram and add tutorial.

The approach here creates the sankey diagram by using the sankey plot function of the `pyam` package. This function requires a mapping dictionary. The mapping dict is created by using the newly created automatic report `message::sankey` and a utility function ~`sankey_mapper()`~`map_for_sankey`.

This pull request is related to the draft pull request in [message_data](https://github.com/iiasa/message_data/pull/164) , which can be removed, since the sankey diagram has been implemented in the `pyam` package.

> [!NOTE]
> The new key is now only added when calling `Reporter.add_sankey()`.

## How to review

General approach should be reviewed. Also, I am not sure if the util function ~`sankey_mapper()`~ `map_for_sankey()` is at the right place.

## PR checklist

- [x] Update release notes.


glatterf42:
Thanks for this PR :)
Getting the tests to pass is a bit tricky since the API for pyam changed when they went from version 1.9.0 to 2.0.0 (as you might expect) and our code needs to satisfy both versions since we test on Python <= 3.9, which is not supported for pyam 2.0.0. So let me take a closer look, maybe we'd want to use the function that pyam's sankey function wraps around directly, after all.

glatterf42:
Thanks @daymontas1 for stepping up here, much appreciated :)

You should be able to test the PR as is and see how you like it's usefulness, but there are a few things that I would like to see happen before we merge it:

- Please rebase on main to keep it up to date with current developments.
- Please migrate the main functionality from `util/__init__.py` to somewhere more appropriate. E.g. `report/sankey.py` if it's a reporting tool or `util/sankey.py` if there are more general use cases.
- Please check if we can use plotly directly instead of pyam's wrapper functionality. This should allow us to utilize more design options and avoids some dependency constraints (see above). See [some examples](https://plotly.com/python/sankey-diagram/) and [their docs](https://plotly.com/python-api-reference/generated/plotly.graph_objects.sankey.html).
- Please confirm that we want to integrate the sankey functionality in the reporting as it currently stands. I'd probably ask at least @khaeru for his opinion. 
- Please write a test if possible to confirm that `sankey_mapper` maps the correct values to the correct keys. We might also need to adapt one or two reporting tests that check the number of reporting steps if this stays in the default reporting workflow.

Please feel free to ask about any of these steps if you need help :)

gidden:
Hi all, to simplify here, I would simply limit this functionality to pyam > 2 - we can do a version check for this on the fly and raise an error.

message_ix/report/__init__.py
```diff
@@ -141,6 +141,7 @@
         "message::costs",
         "message::emissions",
     ),
+    ("message::sankey", "concat", "out::pyam", "in::pyam"),
```
gidden, line 144: I would get rid of this, and simply do the concatenation outside of reporter (e.g., in the jupyter notebook here)

message_ix/util/__init__.py
```diff
@@ -232,3 +233,23 @@ def expand_dims(scenario: Scenario, name, **data):
 
         # Add the expanded data
         scenario.add_par(name, new_data)
+
+
+def sankey_mapper(df, year, region, flows_not_needed=[], variables_not_needed=[]):
+    df_filtered = df.filter(region=region + "*", year=year)
+
+    mapping = {}
+    for var in df_filtered["variable"]:
```
gidden, line 242: this block could be cleaned up as follows:

```
is_input = gvc(var, 0) == "in"
start_idx, end_idx = [1, 2], [3, 4] if is_input else [3, 4], [1, 2]
mapping[var] = (gvc(var, start_idx, join=True), gvc(var, end_idx, join=True))
```

message_ix/util/__init__.py
```diff
@@ -232,3 +233,23 @@ def expand_dims(scenario: Scenario, name, **data):
...
+        if gvc(var, 0) == "in":
+            mapping[var] = (gvc(var, [1, 2], join=True), gvc(var, [3, 4], join=True))
+        if gvc(var, 0) == "out":
+            mapping[var] = (gvc(var, [3, 4], join=True), gvc(var, [1, 2], join=True))
+
+    for k in mapping.keys():
```
gidden, line 248: If above works, then this full block can be removed

message_ix/util/__init__.py
```diff
@@ -232,3 +233,23 @@ def expand_dims(scenario: Scenario, name, **data):
 
         # Add the expanded data
         scenario.add_par(name, new_data)
+
+
+def sankey_mapper(df, year, region, flows_not_needed=[], variables_not_needed=[]):
```
gidden, line 238: at the top of this function we can check that pyam is > 2

glatterf42:
Rebased onto current main and expanded the PR a bit :)
Thanks for jumping in here, @gidden :)
@daymontas1, please pull these latest changes before making further edits. If you already have some edits locally, you can `git stash` them, `git pull` these changes and `git stash pop` your work on top of this.

The missing type hints (from the failing code quality check) probably indicate that we need to add `pyam.*` to our [mypy config exclude list](https://github.com/mabudz/message_ix/blob/bf156ee9285c8014ca9eb77fe7cd71b859bef6e4/pyproject.toml#L94-L103).

message_ix/util/__init__.py
glatterf42: Not quite sure my edits support the same range of exclusion as the original did, but I'm not quite sure what the difference between "flow" and "variable" is and you can now exclude things like "final|electricity", which seems to be what we want?

message_ix/util/__init__.py
glatterf42: Thanks for this suggestion, it almost worked out of the box :)
Maybe most interesting to you: you need
```python
(start_idx, end_idx) = ([1, 2], [3, 4]) if is_input else ([3, 4], [1, 2])
```
to do the combined check you had in mind.

message_ix/report/__init__.py
glatterf42: I kept this for now as I don't know what the Reporter actually does and it seems useful to just get the dataframe format we need from the Reporter. But please elaborate :)

glatterf42:
The other main error I'm seeing just indicates that by adding this reporter step, we need to adapt the expected value of reporter steps in another test, which is no problem :)

message_ix/util/__init__.py
mabudz: I think it makes fully sense to merge both as you did. I used an older piece of code for that. And the reason for distinguishing "flow" and "variable" was, if I remember correctly, that some reports could have something like emissions or costs included. But even in this case one "exclude" list would have been sufficient ;-)

glatterf42: Thanks for confirming :)

message_ix/report/__init__.py
daymontas1, line 144: import logging
from functools import partial
from typing import Tuple, Mapping, List

from genno.operator import broadcast_map
from ixmp.report import Key
from .pyam import collapse_message_cols

# Assuming TASKS1 was where Sankey tasks were defined:
TASKS1 = (
    ("message::sankey", "concat", "out::pyam", "in::pyam"),
)

def get_sankey_tasks() -> List[Tuple[Tuple, Mapping]]:
    """Return a list of tasks for Sankey diagram reporting."""
    to_add: List[Tuple[Tuple, Mapping]] = []
    strict = dict(strict=True)

    # This might include specific Sankey diagram configuration or additional tasks.
    for t in TASKS1:
        to_add.append((t, strict))

    return to_add

class SankeyReporter:
    """A specialized reporter for generating Sankey diagrams."""

    @staticmethod
    def add_tasks(reporter, fail_action: str = "raise") -> None:
        """Add Sankey-related tasks to a given reporter."""
        reporter.add_queue(get_sankey_tasks(), fail=fail_action)

```suggestion
    ("message::sankey", "concat", "out::pyam", "in::pyam"),
```

daymontas1: I have created a new file, sankey.py, within the report directory, which extracts Sankey diagram functionality from __init__.py. Fridolin (@glatterf42), could you please confirm if this is what we want? Additionally, I have updated the westeros_sankey.ipynb file to ensure compatibility with the new sankey.py file in the report directory. I have annotated these changes in the westeros_sankey.ipynb below. Also, in this case, the following line inside the __init__.py file must be either removed or commented out:
# ("message::sankey", "concat", "out::pyam", "in::pyam")

glatterf42: Thanks for this suggestion, @daymontas1 :)
To clarify here: are these the contents of the new `sankey.py` file you mentioned above? They probably don't need to include the `suggestion` part, then. 
If I understand correctly, what you want to suggest is deleting this line that is no longer needed here:
```suggestion

```

glatterf42: Thanks, @daymontas1, these changes sound good to me. For the line that has become redundant, please remove it rather than commenting it out, this keeps the code clean.
Could you please push your changes to the branch so that we can take a look and see how the tests are doing?
If you don't have write access to @mabudz's fork, please let us know so that we can figure out a solution :)

glatterf42:
Is there any update here, @daymontas1? 
Since we want to publish a new release still this week (most likely), I'll postpone this to the 3.10 milestone.

daymontas1:
Hi @glatterf42, I tried to push the changes, but I didn't have permission to do so. So, you need to grant me access. This was my latest comment. Thanks.

glatterf42:
Alright, that's [good to know](https://github.com/iiasa/message_ix/pull/770#discussion_r1590798689). @mabudz, could you please give @daymontas1 write access to your fork? You can do that under Settings -> Collaborators and Teams -> Add people. 
If you do that, please confirm here so we know we can proceed.

If you don't want to wait, @daymontas1, please consider pushing your changes to a branch on your fork or on the main repository: `git push -u <name of the repo> <name of the branch>`. So you need to pick a repo name from the list produced by `git remote -v` and can choose any branch name.

mabudz:
Hi @glatterf42 and @daymontas1 ,

i have added @daymontas1 to the repo ☺️

glatterf42:
Thanks for the quick reaction, much appreciated :)

glatterf42:
Thanks, I've also added @daymontas1 to @iiasa/messageix-devs :)

glatterf42:
In today's team meeting, @SiddharthJoshi-Git mentioned that he is fine with merging this PR as is and providing feedback later on. So I'm happy to merge this once the tests pass, which ... seem kind of confused about which versions of Python they should be using. For message_ix, that is still 3.8 and not 3.13, so maybe we'll have to merge #881 first.

khaeru (changes requested):
Overall this is a welcome contribution. Several things that I will “request” as changes but actually myself implement on the branch:

1. The additional steps between `rep.get("message::sankey")` and `fig.show()` that appear in the tutorial can be added to the `Reporter.add_sankey()` method and controlled with keyword arguments. This saves the user having to manually re-write these steps for every such diagram they want to create.
2. For (1), but also generally, `plotly` as an added dependency appears required to use the feature(s) as intended/recommended. So I would make a new group of optional dependencies `message_ix[sankey] = [plotly]`, and then make `message_ix[sankey]` a recursive dependency in the `tutorials` group.
3. The tests should be expanded to cover the whole range of usage.
4. The new methods and functions must appear in the documentation; currently from the [preview build](https://iiasa-energy-program-message-ix--770.com.readthedocs.build/en/770/) they do not. (A tutorial does not substitute for API documentation.)
5. The release note line should link directly to the tutorial and to the documentation of the new function(s).

Other comments:
- The history of the branch appears to have some duplicate commits (probably due to failure to rebase/inconsistent merging with and rebasing on `main`) and commit names. We should thus squash-and-merge once this is ready to go.
- In order to take advantage of the sankey features via pyam, the data appear to be used via the IAMC structure in which "variable" names like "in|secondary|electricity|grid|standard" are constructed by mashing together labels for distinct dimensions, and then these are again deconstructed using `.util.sankey.set_start_and_end_index()` / `get_variable_components()`. This is both complicated and fragile: if the user chooses to construct "variable" names differently, it will not work.

  For other purposes we've discussed, like drawing RES diagrams using NetworkX/graphviz/others (FYI @macflo8), it will be simpler to use the data at their full/original resolution and construct node IDs using unique combinations of labels for certain dimensions, like (level, commodity), (technology, mode), or others. This may also end up providing a more direct way to construct Sankey diagrams, but we will see once we get there.

khaeru (approved):

PR #789, opened 2024-01-30
Update tests

glatterf42:
The current CI failures have multiple causes. One is pytest 8.0.0, for which this PR adjusts. The affected test was working properly, as far as I can tell, by raising a ValueError before the DeprecationWarning. In the past, since a ValueError was given as the expected result of the test, this was fine; but now, pytest wants to see the DeprecationWarning.
On other tests, one can run `with pytest.raises(ValueError, match="something")` to specify which ValueError is raised -- which would have made us aware of this issue long ago since `Scenario.add_horizon()` raises ValueErrors on multiple occasions. From [pytest's docs](https://docs.pytest.org/en/stable/reference/reference.html#pytest-mark-xfail-ref), it doesn't look like we can pass a `match=` string to the xfail mark, so I'm wondering how much refactoring if would take to implement this. Another option would be to subclass ValueError and raise custom unique errors for the various fail cases, which we can then tell pytest to expect.

This PR also temporarily pins genno < 1.23 and can be used to upgrade Node.js actions as GHA recommends.

## How to review

## PR checklist

- ~[ ] Add, expand, or update documentation.~ Just updating tests
- ~[ ] Update release notes.~ Just updating tests

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -21,14 +21,14 @@ jobs:
 
     steps:
     - name: Check out ixmp
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         repository: iiasa/ixmp
         path: ixmp
         fetch-depth: ${{ env.depth }}
 
     - name: Check out message_ix
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         path: message_ix
         fetch-depth: ${{ env.depth }}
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -33,7 +33,7 @@ requires-python = ">=3.8"
 dependencies = [
   "click",
   "ixmp >= 3.8.0",
-  "genno[pyam] >= 1.20",
+  "genno[pyam] >= 1.20, < 1.23",
```
khaeru, line 36: In the past I've done this in the workflow file, so the commit history of pyproject.toml doesn't contain changes that won't be released. See for instance https://github.com/iiasa/message_ix/commit/76bdd30297d5e5eff31a5ac5f7b42c048a4e63b2

I guess this way is also fine, but it does need a comment (like the one in the above example) to indicate it's a temporary work-around, what relevant issue is, and that it should be removed before a release.

khaeru (changes requested):
Seems good! One change and one question, maybe a change.

pyproject.toml
glatterf42: I've dropped the commit changing pyproject.toml and instead included one that install `genno < 1.23` in the pytest CI workflow. The nightly workflow seems to work just fine with `genno == 1.23`.

khaeru (approved):

Issue #601, opened 2022-05-06
`message_ix.Scenario.rename()` does not work as expected


behnam-zakeri:
`message_ix.Scenario.rename()` is supposed to rename a set element, i.e., replacing the old name with a new name in all `message_ix` sets and parameters. However, this is not the case, i.e., not all sets and parameters are updated with a new name correctly, and there is no error or warning thrown stating this incompleteness. One can for example run the following code and check this. 

It seems the issue is that under this method in `message_ix.core.py`, the check for finding a set element in a parameter is based on `scenario.idx_names()`. This means, e.g., when renaming a "node", the code only finds parameters that have index names of "node", but missing other relevant parameters that have index names related to node but not "node" itself, e.g., "node_loc", "node_rel" etc. This can be resolved relatively easily by using `scenario.idx_sets()` and with a little tweak of the code, and should be done possibly soon, as this is a very useful feature.
```
import ixmp as ix
import message_ix
mp = ix.Platform()
# Load a scenario and clone it
scen = message_ix.Scenario(mp, "Westeros_electrified", "baseline")
test = scen.clone(scenario="test", keep_solution=False)
# Rename an element from set "node" and keeping the original member
test.rename("node", {"Westeros": "Essos"}, keep=True)
# Check if the data exists in the original and renamed format
df = test.par("output", {"node_loc": "Westeros"})
assert not df.empty
df = test.par("output", {"node_loc": "Essos"})
assert not df.empty

**Comment by behnam-zakeri** [comment]
@khaeru and @glatterf42, following the demo in the TIGER workshop, I created this simple test related to this issue. Do you think this is good enough to be added to our test suite?
```
import pytest
from ixmp import Platform
from message_ix.testing import make_westeros


testdata = [
    ("technology", "coal_ppl", "coal_powerplant"),
    ("node", "Westeros", "Tessos"),
]


@pytest.mark.parametrize("set_name,old,new", testdata)
def test_rename(test_mp: Platform, set_name: str, old: str, new: str) -> None:
    """Test if the utility rename works as expected
    Can resolve issue #601 (github.com/iiasa/message_ix/issues/601)
    """
    scen_ref = make_westeros(test_mp, quiet=True)
    scen_ref.solve()

    # Clone to do renaming and tests
    scen = scen_ref.clone(keep_solution=False)

    # Rename old name to a new name
    scen.check_out()
    scen.add_set(set_name, new)
    scen.rename(set_name, {old: new})

    # Check if the new member has been added
[...]

PR #784, opened 2024-01-22
Make tests robust & independent

khaeru:
This PR is to collect changes related to #776. As we notice tests "flaking", we can adjust them on this branch, and then merge once we feel we've addressed all observed flaky tests. The PR will thus remain open for a while until we're convinced we've reached that point.

- Job [1](https://github.com/iiasa/message_ix/actions/runs/7592300001/job/20681496117#step:9:439): `test_add_bound_activity_up_all_modes` fails with `assert np.isclose(300.0, 332.5)`.

## How to review

- Read the diff and note that the CI checks all pass.

## PR checklist

- [x] Continuous integration checks all ✅
  
- [x] Add or expand tests; coverage checks both ✅
- [x] Add, expand, or update documentation.
- [x] Update release notes.
  
- [x] FAILED message_ix/tests/test_feature_bound_activity_shares.py::test_add_bound_activity_up - assert False
     +  where False = <function isclose at 0x105ed3f70>(300.0, 150.0)
     +    where <function isclose at 0x105ed3f70> = np.isclose
    On: macos-latest-py3.8
- [x] FAILED message_ix/tests/test_feature_bound_activity_shares.py::test_add_bound_activity_up_all_modes - ixmp.model.base.ModelError: GAMS errored with return code 3:
    There was an execution error

    For details, see the terminal output above, plus:
    Listing   : /Users/runner/work/message_ix/message_ix/message_ix/model/MESSAGE_run.lst
    Input data: /Users/runner/work/message_ix/message_ix/message_ix/model    /data/MsgData_Canning_problem_(MESSAGE_scheme)_test_add_bound_activity_up_all_modes_cloned.gdx
    On: macos-latest-py3.8, macos-latest-py3.11, macos-latest-py3.12
- [x] FAILED message_ix/tests/test_core.py::test_excel_read_write - assert False
     +  where False = <function isclose at 0x107384eb0>(155.02499389648438, 153.6750030517578)
     +    where <function isclose at 0x107384eb0> = np.isclose
    On: macos-latest-py3.12
- [x] FAILED message_ix/tests/test_integration.py::test_run_clone - RuntimeError: unhandled Java exception: The index set 'year' does not have an element '2010'!
On: macos-latest-py3.11 (x2)
- [x] FAILED message_ix/tests/test_legacy_version.py::test_solve_legacy_scenario - RuntimeError: unhandled Java exception: The index set 'year' does not have an element '1963'!
On: windows-latest-py3.11, windows-latest-py3.10
[...]

PR #791, opened 2024-02-15
Improve `Scenario.rename()` to address #601

behnam-zakeri:
This PR improves the utility `Scenario.rename()` to resolve the bug identified in issue #601.

## How to review
- Copy the unit test introduced in this PR (see the file `test_core.py` to your message_ix `main` branch. The test should fail with the existing code.
- Checkout to the branch of this PR, test should pass.

## PR checklist
- [x] New test is added ✅
- [x] Update release notes.

### DIFF
--- a/message_ix/core.py
+++ b/message_ix/core.py
@@ -751,8 +751,12 @@ class Scenario(ixmp.Scenario):
             if isinstance(ix_set, pd.DataFrame):
                 if name in self.idx_sets(item) and not ix_set.empty:
                     for key, value in mapping.items():
-                        columns = [x for x in self.idx_names(item) if
-                                   self.idx_sets(item)[self.idx_names(item).index(x)] == name]
+                        columns = [
+                            x
+                            for x in self.idx_names(item)
+                            if self.idx_sets(item)[self.idx_names(item).index(x)]
+                            == name
+                        ]
                         for col in columns:
                             df = ix_set[ix_set[col] == key]
                             if not df.empty:
@@ -768,8 +772,11 @@ class Scenario(ixmp.Scenario):
             if name not in self.idx_sets(item):
                 continue
             for key, value in mapping.items():
-                columns = [x for x in self.idx_names(item) if
-                           self.idx_sets(item)[self.idx_names(item).index(x)] == name]
+                columns = [
+                    x
+                    for x in self.idx_names(item)
+                    if self.idx_sets(item)[self.idx_names(item).index(x)] == name
+                ]
                 for col in columns:
                     df = self.par(item, filters={col: [key]})
                     if not df.empty:
--- a/message_ix/core.py
+++ b/message_ix/core.py
@@ -749,12 +749,15 @@ class Scenario(ixmp.Scenario):
         for item in self.set_list():
             ix_set = self.set(item)
             if isinstance(ix_set, pd.DataFrame):
-                if name in ix_set.columns and not ix_set.empty:
+                if name in self.idx_sets(item) and not ix_set.empty:
                     for key, value in mapping.items():
-                        df = ix_set[ix_set[name] == key]
-                        if not df.empty:
-                            df[name] = value
-                            self.add_set(item, df)
+                        columns = [x for x in self.idx_names(item) if
+                                   self.idx_sets(item)[self.idx_names(item).index(x)] == name]
+                        for col in columns:
+                            df = ix_set[ix_set[col] == key]
+                            if not df.empty:
+                                df[columns] = df[columns].replace({key: value})
+                                self.add_set(item, df)
             elif ix_set.isin(keys).any():  # ix_set is pd.Series
                 for key, value in mapping.items():
                     if ix_set.isin([key]).any():
@@ -762,13 +765,16 @@ class Scenario(ixmp.Scenario):
 
         # search for from_tech in pars and replace
         for item in self.par_list():
-            if name not in self.idx_names(item):
+            if name not in self.idx_sets(item):
                 continue
             for key, value in mapping.items():
-                df = self.par(item, filters={name: [key]})
-                if not df.empty:
-                    df[name] = value
-                    self.add_par(item, df)
+                columns = [x for x in self.idx_names(item) if
+                           self.idx_sets(item)[self.idx_names(item).index(x)] == name]
+                for col in columns:
+                    df = self.par(item, filters={col: [key]})
+                    if not df.empty:
+                        df[columns] = df[columns].replace({key: value})
+                        self.add_par(item, df)
 
         # this removes all instances of from_tech in the model
         if not keep:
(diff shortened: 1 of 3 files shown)


behnam-zakeri:
@glatterf42 the tests don't pass on python 3.9. Is this related to the way `pytest.mark.parameterize()` is unpacked between different versions of python?

glatterf42:
There are many different error messages which should not be there if some of the tests are passing at all like 
```shell
FAILED message_ix/tests/test_core.py::TestScenario::test_rename[node-Westeros-Tessos-True] - assert 173795.09375 == (173795.09375 * 2)
```
or
```shell
FAILED message_ix/tests/test_core.py::TestScenario::test_rename[technology-coal_ppl-coal_powerplant-False] - RuntimeError: unhandled Java exception: The index set 'node' does not have an element 'Tessos'!
```
If these things are working for some tests, there should be no reason for them to suddenly fail for others. 
If tests fails but then suddenly succeed on a re-run, these tests are called flaky. Ideally, we want to get to the root cause of their flakiness and remove it. @khaeru has recently opened a PR with precisely that aim: https://github.com/iiasa/message_ix/pull/784
In that PR, the `make_dantzig()` fixture is adapted to be more robust. So I'm wondering whether we can fix these flaky tests by applying the [exact same changes](https://github.com/iiasa/message_ix/pull/784/files) to `make_westeros()`.

glatterf42:
`make_westeros()` already has an optional `request` argument, so I just passed a `request` from the fixture to it, let's see how this does. 

BTW: another continent in the world of Game of Thrones is called Essos, which is tantalizingly close to the Tessos you used initially, so I took the freedom to rename it if you don't mind :)

glatterf42:
One fail remains, but that is one of our usual flaky suspects for which we haven't yet found the root cause, so I'll just re-run it and rest assured that all your tests are passing as expected :)

glatterf42 (approved):
Looks good to me :)

behnam-zakeri:
Thanks @glatterf42 for reviewing this and the nice suggestion for the name Essos (I somehow mixed that). I see your point. It seems some subsequent tests loaded a scenario that was already manupulated by earlier tests. So, giving explicit scenario names using the "request" feature was very useful.

message_ix/tests/test_core.py
```diff
@@ -48,10 +48,12 @@ def test_rename(
...
 
         if not keep:
-            # Check if OBJ value remains unchanged when old is removed (keep=False)
+            # Check if "old" is removed (keep=False)
+            assert old not in set(scen.set(set_name))
+            # Check if OBJ value remains unchanged when "old" is removed (keep=False)
             assert scen.var("OBJ")["lvl"] == scen_ref.var("OBJ")["lvl"]
         elif set_name == "node":
-            # Check if OBJ value is as twice when old "node" is kept (keep=True)
+            # Check if OBJ value is as twice when "old" node is kept (keep=True)
```
glatterf42, line 56:
```suggestion
            # Check if OBJ value is twice as high when "old" node is kept (keep=True)
```

glatterf42:
One more minor suggestion (I think that's a typo), then we can merge this :)

khaeru:
@behnam-zakeri @glatterf42 if it's alright, I will jump in here to make the added code a little more concise and probably performant.

message_ix/core.py
```diff
@@ -749,26 +749,39 @@ def rename(self, name, mapping, keep=False):  # noqa: C901
...
+                columns = [
+                    x
+                    for x in self.idx_names(item)
+                    if self.idx_sets(item)[self.idx_names(item).index(x)] == name
+                ]
+                for col in columns:
+                    df = self.par(item, filters={col: [key]})
+                    if not df.empty:
+                        df[columns] = df[columns].replace({key: value})
+                        self.add_par(item, df)
```
khaeru, line 784: Specifically here, because of the 3 nested `for:` loops, this results in:
- Multiple add_par() calls when renaming e.g. a `node` element in parameter `input`, which has two columns 
- Multiple add_par() calls if `mapping` has > 1 element.
- Both of these effects combined.

This might be slow in our large models (like our global models) with the JDBCBackend.

khaeru:
Done, see the first added commit for details.

I also expanded the test and documentation. Particularly:
- I find that the method does not work for the "year" set, even though this is an index set. Or rather, it works, but the test fails because the resulting scenario is not feasible according to GAMS. I added a note to the docstring and a test case to confirm. I would suggest to make this a follow-up enhancement.
- I added tests and docs for exceptions that come up if one calls the method with invalid arguments: name of an indexed set, name of a parameter, name of neither.
- The method already adds the new elements to the named set. I think this was not clear from the docstring, because Behnam's new test called `scen.add_set()` unnecessarily. So added documentation to clearly state that the method does this.

behnam-zakeri:
Thanks a lot, @khaeru. A learning class for avoiding for loops and if-statements, even when dealing with different item types (i.e., "par" and "set").

message_ix/core.py
```diff
@@ -725,56 +725,85 @@ def add_macro(
...
+
+            # Construct filters for add_{par,set} and an argument for pd.*.replace()
+            if len(names) == 0:
+                # This is the set `name` itself → data will be pd.Series
+                filters = None
+                to_replace: Mapping[str, Union[str, Mapping]] = mapping
+            else:
+                # An indexed set or parameter → data will be pd.DataFrame
+                # Filters: only the values to be replaced
+                filters = {n: mapping.keys() for n in names}
```
behnam-zakeri, line 785: @khaeru this line creates a mapping of filters for all `names` at the same time. But this should be per each name in `names`. For example, if the code is used to rename "node", this filter will be `{"node_loc": "Westeros", "node_origin": "Westeros"}` for parameter `input`, i.e., the union of both "node_loc" and "node_origin". But the intention should be to retrieve and replace any entry in parameter `input` which has "node_loc" **or** "node_origin" of "Westeros". In the Westeros example as it is, we don't have such a case and as such the tests pass. But in the global model, we have many entries in `input` with "node_loc" being a node, e.g., "R11_AFR" and "node_origin" being another node, usually "R11_GLB". The code you suggested would not retrieve/replace these cases, if existed, `{"node_loc": "Westeros", "node_origin": <another node>}`. I hope I'm not missing something out.

khaeru: Thanks for checking carefully! I believe you're correct.

- The built-in `filters=` functionality of ixmp indeed only provides intersection/AND logic, not union/OR.
- We can extend the test cases here by adding 1+ rows that would meet an OR but not an AND condition, and then confirm they are modified as expected. This test will fail at the moment.
- I will next think of a fix. I think either we'll need to reintroduce a second loop (call .par() once with filters for each named dimension) or retrieve all data for the parameter and then pass it through something like `DataFrame.query()`. Ideally these would perform equally well, but in practice I think they might not.

khaeru: Please see the added test and let me know if the code now appears to cover the case you describe.

behnam-zakeri: Thanks @khaeru for reacting quickly. The updated code resolves the issue I mentioned. The new test, `test_rename2`, captures the case very nicely.

glatterf42:
The CI seems fine, the errors are likely flaky, but I'll still confirm this before merging since we don't run the checks again on a merge anymore :)

PR #798, opened 2024-03-12
Update Westeros multinode tutorial for hints

behnam-zakeri:
This PR updates the multinode tutorial for two reasons:
- shortening the code after resolving issue #601 
- adding hints for the in-depth questions

## How to review

Please run Westeros multinode tutorial and make sure the hints provided at the end will work. You can copy-paste those code-based hints at the end of the tutorial and solve the illustrated scenario.

## PR checklist

- [x] Continuous integration checks all ✅
- ~[ ] Add or expand tests; coverage checks both ✅~ Just adding hints to one tutorial
- [X] Add, expand, or update documentation.
- [X] Update release notes.

PR #801, opened 2024-03-14
Adjust tutorials for pandas copy-on-write

khaeru:
Use khaeru/genno#133 to work around has2k1/mizani#38.

Depending on changes made upstream in dask-expr, mizani, or others, this change could be later reverted; it's only for CI to pass.

## How to review

## PR checklist

- ~Add, expand, or update documentation.~ N/A
- ~Update release notes.~


glatterf42 (approved):
Looks good to me, but I don't think it's necessary anymore with the release 0.13.2 of plotnine.

To be clear, though: I'm not sure I like the solution plotnine is using:
```python
        self.pd_option_context = pd.option_context(
            "mode.chained_assignment", None
            "mode.chained_assignment",
            None,
            "mode.copy_on_write",
            False,
        )
```
They are simply disabling this option completely, which does not seem future-proof. Maybe there's no need to comply with this option yet, but there will be and I think I'd like to see warnings before running into this wall.
So maybe we should plan for some time to be allocated to manually enable CoW again and check that everything on our side still works as intended. We might have to pin plotnine to 0.13.1 for this.

khaeru:
> …which does not seem future-proof. Maybe there's no need to comply with this option yet, but there will be and I think I'd like to see warnings before running into this wall.
> So maybe we should plan for some time to be allocated to manually enable CoW again and check that everything on our side still works as intended.

Totally agree with the thinking here.

Looking at khaeru/genno#133, notice the new fixture parametrize_copy_on_write. I used it via this setting:
https://github.com/khaeru/genno/blob/57411be974d21f92b388f5d07913237e6fd58d91/pyproject.toml#L100-L102

So the steps to test ixmp or others would be:
1. Ensure `pytest_plugins = (…, "genno.testing", …)` in conftest.py.
2. Add a similar line to pyproject.toml.
3. Run tests.
4. Comment out or remove the line from (2).

CoW does indeed cause errors in our stack; for instance at iiasa/message_data#545 I found out that the change_technical_lifetime utility has code that does not work with CoW and will need an update. So I think it is prudent to do such checks.

> We might have to pin plotnine to 0.13.1 for this.

I also notice now that plotnine itself has capped pandas < 3.0.0: https://github.com/has2k1/plotnine/blame/main/pyproject.toml#L27 —this could lead to the same kind of problem we've had with pyam, where a pin/cap introduced in *recent* versions causes pypi to resolve a dependency with an *older* version (without the pin/cap), leading to indirect weirdness that's hard to diagnose and work around. (Maybe there needs to be an essay like “Upper bounds on Python package dependencies [considered harmful](https://en.wikipedia.org/wiki/Considered_harmful)”.) We should not willingly create this kind of headache for ourselves and others, so where certain bounds on dependency versions are needed temporarily we should continue to handle those within CI workflows, not in pyproject.toml/releases.

As for this PR, I have not thought through all the user implications, but I guess:
- If the user will be using pandas 3.0, they will also be using a future version of plotnine with appropriate migration applied.
- If the user is using pandas < 3.0 and dask-expr, which causes these errors, then they are likely (now) to get 0.13.2 rather than 0.13.1 or earlier, and then the mitigation here is not needed.

So based on this logic I'd be okay to close the PR without merging.

glatterf42:
We might want to cherry-pick [d7c5f4f](https://github.com/iiasa/message_ix/pull/801/commits/d7c5f4f721948a26ea1a62588e2a6438c01f8b83) at least. Should I add it to #798?

khaeru:
> We might want to cherry-pick [d7c5f4f](https://github.com/iiasa/message_ix/pull/801/commits/d7c5f4f721948a26ea1a62588e2a6438c01f8b83) at least. Should I add it to #798?

Please do.

glatterf42:
Done for now, until we come back via #804.

PR #805, opened 2024-03-21
Introduce some additional labels in docs

glatterf42:
For https://github.com/iiasa/message-ix-models/pull/157, we want to link to the message_ix docs via intersphinx. That works fine for the most part, but we are missing labels for some subsections, so they are not yet reachable via intersphinx. This PR adds these labels.

## How to review

- Check that the documentation is built without error.

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ Only adding docs labels
- ~[ ] Update release notes.~ Only adding docs labels

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -229,6 +229,8 @@ Next:
     $ pytest -m rmessageix
 
 
+.. _common-issues:
+
 Common issues
 =============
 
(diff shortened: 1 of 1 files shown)


khaeru (approved):
Seems good!

One thing I sometimes think of is "Is there a chance that I will want to use a similar label in a different part of the same docs?"

- For "system-dependencies" —probably not.
- For "common-issues" —also probably not. But within, say, message-ix-models there might be "common issues with running MESSAGEix-Nexus", "common issues with running MESSAGEix-Materials", and so we might choose to add a prefix like "water-common-issues".

Again, I don't think that motivates any change for these two.

glatterf42:
The macos-py3.8 test is just flaky, so I'd merge without rerunning it to save some minutes.

PR #793, opened 2024-02-21
Add dac

ywpratama:
**Required:** write a single sentence that describes the changes made by this PR.

## How to review

**Required:** describe specific things that reviewer(s) must do, in order to ensure that the PR achieves its goal.
If no review is required, write “No review:” and describe why.

## PR checklist

- [ ] Continuous integration checks all ✅
  
- [ ] Add or expand tests; coverage checks both ✅
- [ ] Add, expand, or update documentation.
- [ ] Update release notes.

PR #807, opened 2024-03-21
Dac tutorial message models

gidden:
This is a fresh PR where we will implement DACCS related tutorials. It depends on https://github.com/iiasa/message-ix-models/pull/158 where additional features for implementing DACCS in scenarios are developed.

This supercedes #793 and #806

FYI @ywpratama 

**Required:** write a single sentence that describes the changes made by this PR.

## How to review

**Required:** describe specific things that reviewer(s) must do, in order to ensure that the PR achieves its goal.
If no review is required, write “No review:” and describe why.

## PR checklist

- [ ] Update release notes.


message_ix/util/tutorial.py
```diff
@@ -3,17 +3,19 @@
 from functools import partial
 
 from message_ix import Scenario
-from message_ix.report import Key, Reporter, operator
+from message_ix.report import Key, Reporter, computations
```
gidden, line 6: FYI I took this from @ywpratama's active PR. Were there recent changes to the core `Reporter` api? (including file name changes?)

I would not expect all of the change lines here based on what is in https://github.com/iiasa/message_ix/pull/806/files#diff-c26536ce84986d4689980ad63dd90f92b8012bdfcb3bd55ed3df380f71d0c56a

I'll note also this tutorial ran for me locally, so I'm confused if `message_ix.report` not longer has a `computations`. FYI @glatterf42

glatterf42:
Thanks for keeping this work up-to-date, @gidden :)
However, I'm not sure we should be adding tutorials here that depend on code in message-ix-models. AFAIK, code in message-ix-models can rely on code in message_ix, but not the other way round. In theory, message_ix is a general tool and message-ix-models is about a specific application. 
@khaeru please correct me if I'm wrong.

message_ix/util/tutorial.py
glatterf42: Yes, there were changes in v3.8.0 to this exact module. See our [release notes](https://docs.messageix.org/en/stable/whatsnew.html#v3-8-0-2024-01-12). You should be using `from message_ix.report import operator` now instead of `... computations`.

khaeru:
> AFAIK, code in message-ix-models can rely on code in message_ix, but not the other way round. In theory, message_ix is a general tool and message-ix-models is about a specific application.

Yes, that's the order of our stack.

I do acknowledge that we have already a nice, inviting landing location in message-ix for new tutorials: there's a docs page, tests, etc. In message-ix-models, there are as yet no tutorials, so we lack corresponding items. However, we can pretty easily set that up by reusing or copying bits from message-ix. It would be great to have this as the first message-ix-models-specific tutorial.

gidden:
Thanks both - would it be possible to get the skeleton of the docs pages in message-ix models and then we can add these there?

gidden:
Hi all - after further discussions with @ywpratma we decided to make his `add_dac` tool more generic and will open a different PR for that

PR #388, opened 2020-08-19
Fix/Revise the maths formulations

Jihoon:
I find the documentation on the core math formulation is not quite standalone by itself and ends up being quite confusing to external users. It requires filling in or fixing some missing parts to better communicate the formulation doc. I open this PR, but I don't really know answers to all these.

===
Some examples I found:

1. Equation COST_ACCOUNTING_NODAL

- The line explaining n^L is not relevant to the equation. Move it where it first appears, and define 'sub-node'

2. Auxiliary COMMODITY_BALANCE

- Is it missing 'variable' in the title? i.e. "Auxiliary variable COMMODITY_BALANCE"?
- Define h^A below the equation

3. I didn't check all, but multiple equations contain unexplained subscripts, which can only be guessed. (e.g. h^{OD})

4. Sometimes T^R or L^R is used to mean T^{REN} or L^{REN}.

===
Also, I am not sure whether it is a standard notation in the modeling community, but denoting subsets like N(n_hat) is confusing because it usually means functions. Why don't we use N^{n_hat}, just like it does for T^{REN}.

PR #811, opened 2024-03-29
Document mathematical notation

khaeru:
Following on the 2024-03-28 MESSAGE meeting (https://github.com/iiasa/message_ix/pull/388#issuecomment-2024738439), this PR proposes a style of mathematical notation to be used in the docs.

This style is:
- Intended to be more explicit and easier to read and edit than the existing style.
- Documented on a new page.
  The previous/extant style was not documented fully or in any user-visible location.

This PR is not intended to include revision of all equations to match the style, only to establish it. That revision would be accomplished in 1 or more follow-up PRs by a team of contributors.

## How to review

- Look at the [‘Notation’ page](https://iiasa-energy-program-message-ix--811.com.readthedocs.build/en/811/notation.html) in the preview build of the docs.
- Comment below.

## PR checklist

- [ ] Discuss and implement feedback.
- [ ] Complete the section on LaTeX source.
- [ ] Maybe expand the Sphinx [LaTex/MathJax config](https://github.com/iiasa/message_ix/blob/1fca82ff8eadbf7bcb383c099c49ef3b4ba4710d/doc/conf.py#L184-L205).
  In particular, we could define LaTeX commands like `\output` to expand to `\text{output}_{c h h^D l m n^D n^L t y^A y^V}`, since these are almost always written the same way. This would make the LaTeX source of the equations much easier to read and edit.
- [ ] ~Add or expand tests;~ coverage checks both ✅
- [ ] Update release notes.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -196,9 +196,8 @@ intersphinx_mapping = {
 
 # See https://github.com/iiasa/message_ix/pull/721#pullrequestreview-1497907368:
 # prefer to write \text{} explicitly
-# TODO read at least some of these from message_ix.models
-# TODO complete list
-# TODO also add these to a LaTeX preamble
+# TODO Populate these by reference to message_ix.models
+# TODO Also add these to a LaTeX preamble for PDF output
 text_macros = """ACT
 STORAGE
 STORAGE_CHARGE
@@ -212,6 +211,10 @@ mathjax3_config = dict(
     tex=dict(
         macros={k.replace("_", ""): r"\text{" + k + "}" for k in text_macros.split()},
     ),
+)
+# Example only for #811
+mathjax3_config["tex"]["macros"].update(
+    output=r"\text{output}_{c h h^D l m n^D n^L t y^A y^V}",
 )
 
 # -- Options for sphinx.ext.napoleon ---------------------------------------------------
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -71,6 +71,7 @@ the core MESSAGEix mathematical implementation.
    :hidden:
    :caption: Mathematical Specification
 
+   notation
    model/MESSAGE/sets_maps_def
    time
    model/MESSAGE/parameter_def
--- a/doc/notation.rst
+++ b/doc/notation.rst
@@ -7,7 +7,7 @@ This page explains the notation used in this mathematical formulation section of
 
    & \forall \, c, h, \ l \notin L^{\text{RES}} \cup L^{\text{REN}} \cup L^{\text{STOR}}, n = n^D = n^O, y = y^A: \\&
    \sum_{h m n^L t ; y^V \leq y}{\left(
-     \text{output}_{c h h^D l m n^D n^L t y^A y^V}
+     \output
      \cdot \text{duration_time_rel}_{h^1=h^D;h^2=h}
      \cdot \text{ACT}_{h m n^L t y^A y^V}
    \right)} \\&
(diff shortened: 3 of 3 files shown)


Jihoon:
Thanks, Paul, for putting this together. I tried to go through it but then find it still quite hard to follow the details. 

Some observations:
1. If you sum over h on LHS, RHS shouldn't have subscript h.
2. Using y and y^A interchangeably is confusing. The summation mark has y, while there is no y inside the parentheses.
3. I don't quite get the 'alignment' part and the meaning of this. image.  'Aligned' means identical?
4. In this sense, indexing was a bit better to follow in the old setting. 
5. For the math equation, why don't we use parenthesis within brackets instead of nesting parentheses in parenthesis? (differently from usual coding style)
6. I would assume there is a separate table or something to explain what each index means (like c/h/l). All in one place.
7. Putting =,>, or < on a separate line makes it look also very hidden. What about keeping it the old way: what's being defined goes to LHS, and the elaborative or the specification go RHS. For the given example, putting `COMMODITY_BALANCE` on LHS and the rest aligned to the right of the equality sign, and at each +/- sign, line changes.
8. So what is n^L in the end?

Thanks again. This is a VERY helpful effort for readers but a very challenging task for the documentor..

khaeru:
Thanks @Jihoon for the close read! These are all important points to consider.

> 1. If you sum over h on LHS, RHS shouldn't have subscript h.
> 2. Using y and y^A interchangeably is confusing. The summation mark has y, while there is no y inside the parentheses.

These are both errors I think. For (2) I guess it should be y^A on the summation mark for consistency. I'm not sure about (1).
 
> 3. I don't quite get the 'alignment' part and the meaning of this. image. 'Aligned' means identical?

Yes. This is supposed to reflect that some parameters have dimension n^D, others n^O, others n. If one is e.g. "R12_AFR", then the others are the same.

This is for instance _not_ the case for n^L: n^L can have one value while n^D, n^O, and n have a different one (the same value for all three).

IMHO this is a tricky concept *per se*, and hard to convey it concisely and clearly through notation.

> 5. For the math equation, why don't we use parenthesis within brackets instead of nesting parentheses in parenthesis? (differently from usual coding style)

In LaTeX it is possible to write `\left( \left( \right) \right)` and the LaTeX compiler will make progressively smaller or larger parentheses, which makes it easy to visually match them. MathJAX doesn't seem to support this.

I would be fine for describing an order (e.g. () innermost, followed by [], then alternate…or something like that) but it would be a bit harder to apply consistently. Will consider this further.

> 6. I would assume there is a separate table or something to explain what each index means (like c/h/l). All in one place.

Yes, these are already given on the page for [Sets and mappings](https://docs.messageix.org/en/latest/model/MESSAGE/sets_maps_def.html). I think this could be improved, but didn't want to get into that yet.

> 7. Putting =,>, or < on a separate line makes it look also very hidden. What about keeping it the old way: what's being defined goes to LHS, and the elaborative or the specification go RHS. For the given example, putting `COMMODITY_BALANCE` on LHS and the rest aligned to the right of the equality sign, and at each +/- sign, line changes.

This would produce more line wraps and possibly longer lines that overflow.
I would be okay with always starting with the side that is more brief or a single symbol and keeping the (in)equality on the same line:

```
COMMODITY_BALANCE =
\sum{
- …
+ …
```

My one worry with this is that it may reverse the way some equations are named and actually implemented in the code. For example, if a equation is called BLAH_GT or BLAH_LT but then we end up with respectively the symbol ≤ ("less than or equal") or ≥ ("greater than or equal"), the mismatch is potentially confusing. But I will try and see how it looks.

> 8. n^L is not defined in the end?

It should be defined at (6), maybe isn't.

PR #870, opened 2024-07-26
Migrate issue templates to forms

glatterf42:
This PR migrates the templates we currently use for issue creation to forms. Forms have the advantage of being more specific, for example, with the forms below, people would not have to skip the point `What happened?` and anything they post in the `Code Sample` field would be formatted as python code automatically, etc. This should help streamlining issues and increasing their quality.

For the moment, I have translated our current templates as directly as possible. These changes have been tested on my fork (see below), where one can try opening an issue using these forms. The one exception is that I'm trying to still include a comment here that links to the original inspiration. This might not be possible, but I still want to try, credit where credit's due.

Please note that forms are currently still in beta, so subject to change.

## How to review

- Look at https://github.com/glatterf42/message_ix/issues/new/choose and open one of the issue types. See if you like the layout.

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ Just affects GitHub pages.
- ~[ ] Add, expand, or update documentation.~ Just affects GitHub pages.
- ~[ ] Update release notes.~ Just affects GitHub pages.


khaeru (approved):
Looks great. Let's merge and then iterate based on whether people appear to have any hiccups using this initial set of forms.

Issue #867, opened 2024-07-25
Extend historical parameters to track all VAR when shifting `firstmodelyear`


OFR-IIASA:
Currently, only select VAR is tracked via historical parameters, meaning that when shifting the firstmodelyear a lot of information for example PRICES and total installed capacity are lost and cannot be traced.

Issue #543, opened 2022-01-25
Parameter "historical_emission" no populated.


OFR-IIASA:
The parameter `historical_emission` should automatically be tracked with the value from the variable `EMISS` when shifting the `firstmodelyear`, as was explained here in issue #314. 

A simple test using the Westeros tutorial (https://github.com/iiasa/message_ix/blob/main/tutorial/westeros/westeros_emissions_bounds.ipynb) showed that the parameter is indeed not populated.

This can be replicated by running the above mentioned tutorial. At the end, clone "scen" and shift the `firstmodelyear` in the process to 710. Retrieving the parameter "historical_emission" returns an empty dataframe, despite original scenario containing entries in the `var(""EMISS)`.

khaeru:
This is also linked to #254: the functionality to shift the first model year is currently embedded in the Java code of the ixmp JDBCBackend. This is both a "stack level violation" and means that the code is not currently maintained or updated with improvements in this package.

A fix will involve creating utility code (basically a single function) here that shifts the first model year in a message_ix.Scenario; then the Java code can be bypassed/never used. That function should address this bug.

khaeru:
I've relabeled this. While it _is_ an intuitive and straightforward thing to expect given the brief/vague description of this argument, the docs also don't promise that this should happen, it's not a bug.

Issue #542, opened 2022-01-25
Add parameter "historical_total_capacity"


OFR-IIASA:
Currently, the framework currently only allows the tracking of historically "new" installed capacity, via the parameter `historical_new_capacity`. The information about early-retired capacity is therefore lost when shifting the `firstmodelyear`. This means, that for the purpose of reporting, is not possible to replicate the amount of total installed capacity for year `y`, when shifting the `firstmodelyear` to `y+1`.

As their is currently no dynamic growth constraint associated with total capacity, this issue does not impact model results.

PR #873, opened 2024-08-12
Add `.models.shift_period()`

khaeru:
This PR adds a function `shift_period()` that implements, purely in Python, the behaviour of `Scenario.clone(…, shift_first_model_year)`. Because the latter is currently implemented in the Java code underlying `ixmp.JDBCBackend`, it cannot be extended in parallel with changes to the implementation of MESSAGE in GAMS (or, in the future, other implementations).

- Partly addresses #254.
- Also due to limitations of `ixmp.JDBCBackend`, is it not supported to set values on model data items of type "variable" or "equation" directly. Thus, this PR can't be completed without iiasa/ixmp#552, iiasa/ixmp#400, or similar.
- cf. also iiasa/ixmp#424. While the behaviour is currently provided as an argument to the `clone()` command, the new function is distinct.
- This PR does not directly address, but will make it much easier to address:
  - #542
  - #543
  - #867
- The operation of shifting the first model period *forward* in time is *lossy*, because parameters like `historical_activity` have fewer dimensions than the corresponding variables (e.g. `ACT`). For the opposite operation—shifting the first model period *backwards* in time—essentially the scenario must be solved again to generate the corresponding solution data in e.g. `ACT`.

**More to be added**

## How to review

**To be added**

## PR checklist

- [ ] Complete the implementation
  - [ ] Invoke `shift_period()` from `message_ix.Scenario.clone(…)` if the corresponding argument is given.
- [ ] Update release notes.

### DIFF
--- a/message_ix/message.py
+++ b/message_ix/message.py
@@ -1,6 +1,7 @@
 import logging
 from collections.abc import MutableMapping
 from functools import partial
+from typing import TYPE_CHECKING
 from warnings import warn
 
 import ixmp.model.gams
@@ -17,6 +18,10 @@ from .common import (
     _filter_log_initialize_items,
     _item_shorthand,
 )
+
+if TYPE_CHECKING:
+    from message_ix.core import Scenario
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


glatterf42:
This will address https://github.com/iiasa/ixmp/issues/75.

PR #877, opened 2024-09-12
Make more tests robust and independent

### LINKED ISSUE #797: Make more tests robust & independent
Unfortunately, even though #784 improved the tests a lot, we still seem to see flaky behaviour. So let's continue the same kind of work from #784 on these tests:

- [ ]
  ```python
  FAILED message_ix/tests/test_feature_bound_activity_shares.py::test_add_share_mode_lo - assert False
   +  where False = <function isclose at 0x111d84700>(0.5416666666666666, 0.525)
   +    where <function isclose at 0x111d84700> = np.isclose
  ```

glatterf42:
Closes #797.

This is a follow-up PR to #784 with the aim of bringing even more test stability into our CI suite.

For the moment (at least while it's free), this PR experiments with [codecov's new flaky-test-detection feature](https://about.codecov.io/blog/find-failing-and-flaky-tests-with-codecov-test-analytics/).

## How to review

- Read the PR comments and ensure the reasoning makes sense.

## PR checklist

- [x] Update tests; coverage checks both ✅
- [x] Update release notes.

### DIFF
--- a/message_ix/cli.py
+++ b/message_ix/cli.py
@@ -34,10 +34,10 @@ ixmp.cli.ScenarioClass = message_ix.Scenario
 @click.option(
     "--source-dir",
     type=click.Path(file_okay=False),
-    help="Optional directory to copy file from.",
+    help="Optional directory to copy files from.",
 )
 @click.argument("path", type=click.Path(file_okay=False))
-def copy_model_cmd(path, source_dir, overwrite, set_default, quiet):
+def copy_model_cmd(path, overwrite, set_default, quiet, source_dir):
     """Copy the MESSAGE GAMS files to a new PATH.
 
     To use an existing set of GAMS files, you can also call:
@@ -46,7 +46,7 @@ def copy_model_cmd(path, source_dir, overwrite, set_default, quiet):
     """
     from message_ix.util import copy_model
 
-    copy_model(path, source_dir, overwrite, set_default, quiet)
+    copy_model(path, overwrite, set_default, quiet, source_dir=source_dir)
 
 
 @main.command()
--- a/message_ix/cli.py
+++ b/message_ix/cli.py
@@ -65,7 +65,7 @@ def dl(branch, tag, path):
     from urllib.request import Request, urlopen
 
     if tag and branch:
-        raise click.BadOptionUsage("Can only provide one of `tag` or `branch`")
+        raise click.UsageError(message="Can only provide one of `tag` or `branch`")
     elif branch:
         # Construct URL and filename from branch
         zipname = f"{branch}.zip"
--- a/message_ix/cli.py
+++ b/message_ix/cli.py
@@ -31,8 +31,13 @@ ixmp.cli.ScenarioClass = message_ix.Scenario
 )
 @click.option("--overwrite", is_flag=True, help="Overwrite existing files.")
 @click.option("--quiet", is_flag=True, help="Don't print paths of copied files.")
+@click.option(
+    "--source-dir",
+    type=click.Path(file_okay=False),
+    help="Optional directory to copy file from.",
+)
 @click.argument("path", type=click.Path(file_okay=False))
-def copy_model_cmd(path, overwrite, set_default, quiet):
+def copy_model_cmd(path, source_dir, overwrite, set_default, quiet):
     """Copy the MESSAGE GAMS files to a new PATH.
 
     To use an existing set of GAMS files, you can also call:
@@ -41,7 +46,7 @@ def copy_model_cmd(path, overwrite, set_default, quiet):
     """
     from message_ix.util import copy_model
 
-    copy_model(path, overwrite, set_default, quiet)
+    copy_model(path, source_dir, overwrite, set_default, quiet)
 
 
 @main.command()
--- a/message_ix/tests/test_cli.py
+++ b/message_ix/tests/test_cli.py
@@ -49,36 +49,32 @@ def test_copy_model(
 
 
 @pytest.mark.parametrize(
-    "opts",
+    "opts, exit_code",
     [
-        "",
-        "--branch=main",
-        "--tag=v1.2.0",
+        ([""], 0),
+        (["--branch=main"], 0),
+        (["--tag=v1.2.0"], 0),
         # Nonexistent tag
-        pytest.param("--tag=v999", marks=pytest.mark.xfail(raises=ValueError)),
+        pytest.param(["--tag=v999"], 0, marks=pytest.mark.xfail(raises=ValueError)),
         # Can't use both --tag and --branch
+        # TODO Why is this not actually raising a click.UsageError?
         pytest.param(
             ["--branch=main", "--tag=v1.2.0"],
+            2,
             marks=pytest.mark.xfail(raises=click.UsageError),
         ),
     ],
 )
-def test_dl(message_ix_cli, opts, tmp_path):
-    if isinstance(opts, list):
-        r = message_ix_cli("dl", *opts, str(tmp_path))
-        # TODO Why is this not actually raising a click.UsageError?
-        assert r.exit_code == 2, (r.exception, r.output)
-    else:
-        r = message_ix_cli("dl", opts, str(tmp_path))
+def test_dl(message_ix_cli, opts, exit_code, tmp_path):
+    r = message_ix_cli("dl", *opts, str(tmp_path))
+    assert r.exit_code == exit_code, (r.exception, r.output)
 
-        assert r.exit_code == 0, (r.exception, r.output)
+    if opts == "":
+        # Guess what the latest release will be from GitHub, using __version__.
+        # This string is provided by setuptools-scm based on the most recent
+        # Git tag, e.g. if the tag is 'v3.0.0' it may be '3.0.1.devN+etc'.
+        major = message_ix.__version__.split(".")[0]
 
-        if opts == "":
-            # Guess what the latest release will be from GitHub, using __version__.
-            # This string is provided by setuptools-scm based on the most recent
-            # Git tag, e.g. if the tag is 'v3.0.0' it may be '3.0.1.devN+etc'.
-            major = message_ix.__version__.split(".")[0]
-
-            # 'message-ix dl' defaults to the latest release
-            pattern = re.compile(rf"Default: latest release v{major}\.\d+\.\d+")
-            assert pattern.match(r.output)
+        # 'message-ix dl' defaults to the latest release
+        pattern = re.compile(rf"Default: latest release v{major}\.\d+\.\d+")
+        assert pattern.match(r.output)
--- a/message_ix/tests/test_cli.py
+++ b/message_ix/tests/test_cli.py
@@ -1,5 +1,4 @@
 import re
-import shutil
 from pathlib import Path
 
 import pytest
@@ -8,7 +7,9 @@ import message_ix
 from message_ix import config
 
 
-def test_copy_model(monkeypatch, message_ix_cli, tmp_path, tmp_env, request):
+def test_copy_model(
+    monkeypatch, message_ix_cli, tmp_path, tmp_env, request, tmp_model_dir
+):
     # Use Pytest monkeypatch fixture; this ensures the original value is restored at the
     # end of the test
     monkeypatch.setattr(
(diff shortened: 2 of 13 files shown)


glatterf42:
Most items from #797 seem addressed now. Mainly, I'm not yet sure why the tests in `test_feature_bound_activity_shares` are flaky. The only thing I can think of is that despite our previous efforts (identifying scenario names via `request.node.name`), some scenarios could still get mixed up so that bounds would be applied sort of randomly instead of as intended. So for now, I have doubled down on this approach with the help of [pytest-xdist's recommended way of uniquely identifying each test run](https://pytest-xdist.readthedocs.io/en/stable/how-to.html#uniquely-identifying-the-current-test-run).
For now, I've only used it in `test_feature_bound_activity_share`, but if it works (and we don't intend to drop pytest-xdist), we could use this feature elsewhere, too, to replace `request.node.name`.
However, in general, I thought `request.node.name` was uniquely identifying scenarios already, so I'm not quite convinced this new approach works any better.

glatterf42:
I have to admit: while I did make changes to `test_feature_bound_activity_shares`, I was not sure they would address the flakiness seen in `test_add_share_mode_lo` or `test_commodity_share_lo`. However, I have re-run all tests five times now and have not observed a single flaky test, so maybe one of the changes I made *did* resolve that flakiness.

For future reference: I have tried replicating the numbers reported in https://github.com/iiasa/message_ix/issues/797#issue-2178453258: called `make_dantzig` locally, defined `calc_share` as it is in the test file and I find:
```python
>>> 1.05 * calc_share(scen1)
0.56875
>>> calc_share(scen1)
0.5416666666666666
```
This seems odd since `0.56875` is not one of the reported values when flakiness occurs, seemingly suggesting either that the numbers are reported from a different scenario altogether or that the scenario in question had also been changed by another test (in the same module; since all of them use the module-scoped fixture `test_mp`). For example, maybe the `test_add_share_mode_up` frequently added a `0.95` multiplier to the scenario, resulting in the `test_add_share_mode_lo` scenario finding the "normal" `0.5416...` value (due to its reverse-factor of 1 / 0.95 = 1.05). 
In either of those cases, the changes I made (ensuring each clone gets its own unique scenario name) could well be the changes needed to eliminate the flakiness.

glatterf42:
Today, I'm seeing these flaky errors for the first time:
```python
 __________________________________ test_dl[] ___________________________________
[gw1] linux -- Python 3.9.20 /opt/hostedtoolcache/Python/3.9.20/x64/bin/python

message_ix_cli = <bound method message_ix_cli.<locals>.Runner.invoke of <message_ix.tests.conftest.message_ix_cli.<locals>.Runner object at 0x7fc27797a2b0>>
opts = ''
tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw1/test_dl__0')

    @pytest.mark.parametrize(
        "opts",
        [
            "",
            "--branch=main",
            "--tag=v1.2.0",
            # Nonexistent tag
            pytest.param("--tag=v999", marks=pytest.mark.xfail(raises=AssertionError)),
        ],
    )
    def test_dl(message_ix_cli, opts, tmp_path):
        r = message_ix_cli("dl", opts, str(tmp_path))
    
>       assert r.exit_code == 0, (r.exception, r.output)
E       AssertionError: (URLError(TimeoutError(110, 'Connection timed out')), '')
E       assert 1 == 0
E        +  where 1 = <Result URLError(TimeoutError(110, 'Connection timed out'))>.exit_code

message_ix/tests/test_cli.py:66: AssertionError
____________________________ test_dl[--tag=v1.2.0] _____________________________
[gw0] linux -- Python 3.9.20 /opt/hostedtoolcache/Python/3.9.20/x64/bin/python

message_ix_cli = <bound method message_ix_cli.<locals>.Runner.invoke of <message_ix.tests.conftest.message_ix_cli.<locals>.Runner object at 0x7f3e3f6a8e20>>
opts = '--tag=v1.2.0'
tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw0/test_dl___tag_v1_2_0_0')

    @pytest.mark.parametrize(
        "opts",
        [
            "",
            "--branch=main",
            "--tag=v1.2.0",
            # Nonexistent tag
            pytest.param("--tag=v999", marks=pytest.mark.xfail(raises=AssertionError)),
        ],
    )
    def test_dl(message_ix_cli, opts, tmp_path):
        r = message_ix_cli("dl", opts, str(tmp_path))
    
>       assert r.exit_code == 0, (r.exception, r.output)
E       AssertionError: (URLError(TimeoutError(110, 'Connection timed out')), '')
E       assert 1 == 0
E        +  where 1 = <Result URLError(TimeoutError(110, 'Connection timed out'))>.exit_code

message_ix/tests/test_cli.py:66: AssertionError
 ____________________________ test_dl[--branch=main] ____________________________
[gw1] linux -- Python 3.9.20 /opt/hostedtoolcache/Python/3.9.20/x64/bin/python

message_ix_cli = <bound method message_ix_cli.<locals>.Runner.invoke of <message_ix.tests.conftest.message_ix_cli.<locals>.Runner object at 0x7fc27797a2b0>>
opts = '--branch=main'
tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw1/test_dl___branch_main_0')

    @pytest.mark.parametrize(
        "opts",
        [
            "",
            "--branch=main",
            "--tag=v1.2.0",
            # Nonexistent tag
            pytest.param("--tag=v999", marks=pytest.mark.xfail(raises=AssertionError)),
        ],
    )
    def test_dl(message_ix_cli, opts, tmp_path):
        r = message_ix_cli("dl", opts, str(tmp_path))
    
>       assert r.exit_code == 0, (r.exception, r.output)
E       AssertionError: (URLError(TimeoutError(110, 'Connection timed out')), 'Retrieving https://github.com/iiasa/message_ix/archive/main.zip
E         ')
E       assert 1 == 0
E        +  where 1 = <Result URLError(TimeoutError(110, 'Connection timed out'))>.exit_code

message_ix/tests/test_cli.py:66: AssertionError
```

Unfortunately, they don't provide too much information. If I'm not mistaken, the first two tests timed out [retrieving the `tags_info`](https://github.com/iiasa/message_ix/blob/c66994489fdf92574e6b791e4d502541ea79f7a8/message_ix/cli.py#L89-L90) from GitHub, while the last test apparently timed out [retrieving the data from GitHub](https://github.com/iiasa/message_ix/blob/c66994489fdf92574e6b791e4d502541ea79f7a8/message_ix/cli.py#L114-L115). In both cases, I'm not sure what we can do to avoid the flakiness (other than re-running tests on fails): there's just one GitHub address we need to address and it's the same for all tests and we're retrieving all data to temporary directories (that should be unique for each test run), anyway. So this is mainly documenting for future reference.

Lastly, I noticed that we didn't test calling the CLI command `dl` with both `--tag` and `--branch`, so I added a test for that. Unfortunately, click handles the exception somewhat awkwardly, so that the test is for now `XPass`ing: it is triggering the right `click.UsageError`, but this gets printed to stdout and raises a `SystemExit`, which doesn't make it through our `message_ix_cli` fixture: this still returns `r`, only now it has an `.exit_code != 0`. While this is not ideal, I didn't want to spend too much time on this right now.

.github/workflows/pytest.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
         pytest message_ix \
           -m "not nightly and not tutorial" \
           -rA --verbose --color=yes --durations=20 \
-          --cov-report=xml \
+          --cov --junitxml=junit.xml \
```
khaeru, line 101: As I understand, the --cov-report=xml file (coverage.xml) includes the line coverage, whereas the --junitxml file includes the results of tests.

I guess the latter is needed for Codecov to detect flaky tests. But doesn't this change also turn off generation of the former? How does Codecov then receive info on line coverage?

message_ix/util/__init__.py
```diff
@@ -14,28 +15,31 @@
 
 
 def copy_model(
-    path: Path, overwrite: bool = False, set_default: bool = False, quiet: bool = False
+    path: Path,
+    source_dir: Optional[Path] = None,
```
khaeru, line 19: If our own downstream code, or user code, has for instance:
```py
copy_model("/foo/bar", False, False, True)
```
…then inserting a new positional-or-keyword argument before other positional-or-keyword arguments will break that code.

In point of fact, such code likely doesn't exist. But in general, good practice is to follow a sequence like:
1. Add new arguments at the end of the list. Possibly add them as keyword-only:
   ```py
   ...
   quiet: bool = False,
   *,
   source_dir: Optional[Path] = None,
   ...
   ```
2. Possibly, later make certain other arguments keyword-only. (This is a pattern pandas has used, for instance.)
   - First deprecate and warn that `overwrite`, `set_default`, `quiet` will not be accepted as positional arguments.
   - Then in a later version remove the deprecated functionality.

Can we at least do (1) here?

khaeru: Also, the added argument and CLI option are changes in the user API, so they should be mentioned in the release notes.

message_ix/util/__init__.py
```diff
@@ -46,28 +50,29 @@ def output(str):
     else:
         output = print
 
-    # Identify the source directory using importlib.resources
```
khaeru, line 49: Just recording for posterity that:
- `importlib.resources` was used in #784 because it's in the Python standard library, thus appears to be recommended/preferred over `pathlib.Path(__file__)`.
- However, there are [strong dissenting views](https://discuss.python.org/t/easy-and-recommended-way-to-get-path-of-datafile-within-package/20581/4) from Python users.
- It was decided to switch to using `pathlib.Path(__file__)`. This reverses changes made in the earlier PR.
- We don't know if this has any impact on flakiness of tests.

message_ix/cli.py
```diff
@@ -31,8 +31,13 @@
 )
 @click.option("--overwrite", is_flag=True, help="Overwrite existing files.")
 @click.option("--quiet", is_flag=True, help="Don't print paths of copied files.")
+@click.option(
+    "--source-dir",
+    type=click.Path(file_okay=False),
+    help="Optional directory to copy file from.",
```
khaeru, line 37: “files” plural?

message_ix/tests/test_cli.py
```diff
@@ -57,20 +55,30 @@ def test_copy_model(monkeypatch, message_ix_cli, tmp_path, tmp_env):
...
-    if opts == "":
-        # Guess what the latest release will be from GitHub, using __version__.
-        # This string is provided by setuptools-scm based on the most recent
-        # Git tag, e.g. if the tag is 'v3.0.0' it may be '3.0.1.devN+etc'.
-        major = message_ix.__version__.split(".")[0]
-
-        # 'message-ix dl' defaults to the latest release
-        pattern = re.compile(rf"Default: latest release v{major}\.\d+\.\d+")
-        assert pattern.match(r.output)
+    if isinstance(opts, list):
```
khaeru, line 67: In order to avoid this branch & duplication, use a list of (possibly 0- or 1-element) lists in the `parametrize()` call:
```python
@pytest.mark.parametrize(
    "opts",
    [
        [],
        ["--branch=main"],
        ["--tag=v1.2.0"],
        # Nonexistent tag
        pytest.param(["--tag=v999"], marks=pytest.mark.xfail(raises=AssertionError)),
        pytest.param(["--tag=v999"], marks=pytest.mark.xfail(raises=ValueError)),
        # Can't use both --tag and --branch
        pytest.param(
            ["--branch=main", "--tag=v1.2.0"],
            marks=pytest.mark.xfail(raises=click.UsageError),
        ),
    ],
)
```

khaeru (approved):
Thanks for the thorough tracking and many fixes to flaky tests! The use of isolation across the test suite now exhibits a more consistent style that can be followed by others making additions to the code.

I put some notes inline, and a few small requested changes. With those, this will be good to merge.

message_ix/tests/test_cli.py
glatterf42: Yes, I thought about that, too. This would probably be nicer if the last case really raised a `click.UsageError`, but I'll include this in a commit, anyway.

.github/workflows/pytest.yaml
glatterf42: Happy to drop this commit since flaky-test-detection is going to become a paid feature sooner or later, anyway. I mainly wanted to see if it could be useful for this PR, no need to keep these changes :)

PR #880, opened 2024-10-14
Drop outdated pin of JPype on CI

glatterf42:
Today's [CI failures](https://github.com/iiasa/message_ix/actions/runs/11321261760) somewhat irritate me since we install anaconda and miniconda versions for Python 3.10, and `conda info` even shows this, but `message-ix show-versions` then shows that Python version 3.8.20 is used. 3.8 has just reached its end of life, so we can probably drop testing it and the main difference to the runs using newer versions is the pin of JPype 1.2.1.
This version was released almost four years ago now (4.1.2021), so I think it's safe to remove this pin entirely rather than trying to debug the action.

## How to review

Please note: this PR contains one commit that temporarily enables the action for this PR; that should be dropped before merging.

## PR checklist

- ~[ ] Add or expand tests; coverage checks both ✅~ Just CI update.
- ~[ ] Add, expand, or update documentation.~ Just CI update.
- ~[ ] Update release notes.~ Just CI update.


khaeru:
> 3.8 has just reached its end of life, so we can probably drop testing it and the main difference to the runs using newer versions is the pin of JPype 1.2.1.

Rather than “we can probably drop testing [3.8]”, it is our versioning policy to drop support for EOL Pythons and add support for new ones ASAP. This made me check and realize that we don't yet have a PR on message_ix parallel to iiasa/ixmp#544. Is it possible to expand this one to include parallel changes?

glatterf42:
True, good catch. Most of the required changes should be here now :)
If the tests are successful, I can also declare the 3.13-tests as required instead of 3.12.

glatterf42:
The current failures arise from failing to build Python-3.13 wheels for pyarrow, which should be fixed [once they release v18.0.0](https://github.com/apache/arrow/issues/43519#issuecomment-2408925458), but this has yet to happen.

khaeru (approved):
Good to go, including a rebase to drop the "TEMPORARY" commit.

PR #886, opened 2024-11-11
Replace resolved numpy workaround with dask workaround

glatterf42:
This PR tries to resolve the CI test failures by not using version 2024.11.0 of dask, which is causing [issues with genno](https://github.com/khaeru/genno/issues/149). 

The most preferable solution would be to adapt genno to this dask version, but it is unclear how much effort this would be and how long this would take. So for now, I'm simply imposing that we don't use that particular version (note: 2024.11.1 would be allowed, hoping that this might resolve the incompatibilities).
If this workaround is successful and we agree we want to employ it for now, we should also adopt it in the nightly workflow.

## How to review

## PR checklist

- [x] Update CI tests; coverage checks both ✅

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -170,9 +170,9 @@ jobs:
         pip install --upgrade "ixmp @ git+https://github.com/iiasa/ixmp.git@main"
         pip install .[tests]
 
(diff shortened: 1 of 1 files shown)


glatterf42:
Superseded by #887. Apparently, GitHub will not simply execute a workflow file from a fork, which seems reasonable.

PR #887, opened 2024-11-11
Replace resolved numpy workaround with dask workaround

glatterf42:
This PR tries to resolve the CI test failures by not using version 2024.11.0 of dask, which is causing [issues with genno](https://github.com/khaeru/genno/issues/149). 

The most preferable solution would be to adapt genno to this dask version, but it is unclear how much effort this would be and how long this would take. So for now, I'm simply imposing that we don't use that particular version (note: 2024.11.1 would be allowed, hoping that this might resolve the incompatibilities).
If this workaround is successful and we agree we want to employ it for now, we should also adopt it in the nightly workflow.

## How to review

## PR checklist

- [x] Update CI tests; coverage checks both ✅

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -66,9 +66,9 @@ jobs:
         (cd ixmp; pip install .[tests])
         (cd message_ix; pip install .[tests])
 
-        # TEMPORARY Work around unionai-oss/pandera#1685;
-        # see https://github.com/khaeru/genno/issues/140
-        pip install "numpy < 2"
+        # TEMPORARY Work around dask v2024.11.0;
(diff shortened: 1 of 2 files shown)


khaeru (approved):
Thanks!

I would only suggest to use `dask < 2024.11.0` instead of `dask != 2024.11.0`.

The reason is that, until or unless we make a report to dask upstream and they choose to fix it, we should expect that future releases of dask (e.g. 2024.11.1, 2024.12.0, 2025.1.0) will also cause the same errors.

Using `<` avoids any need to adjust the workaround.

glatterf42:
Thanks, sounds good :)
I can open a PR similar to this one on message-ix-models, too.

PR #603, opened 2022-05-10
attempt a nightly workflow only on approved PRs

gidden:
**Required:** write a single sentence that describes the changes made by this PR.

Copies nightly workflow, changes to evoke on `pull_request_review`, and checks that review state is `approved`. 

Related to #591.

## How to review

**Required:** describe specific things that reviewer(s) must do, in order to ensure that the PR achieves its goal.
If no review is required, write “No review:” and describe why.

Approve the PR, and confirm that the nightly tests are executed and result as expected.

## PR checklist

- [ ] Continuous integration checks all ✅
  
- [ ] Add or expand tests; coverage checks both ✅
- [ ] Add, expand, or update documentation.
- [ ] Update release notes.

PR #892, opened 2024-11-21
Bump GAMS version in nightly CI worklow

glatterf42:
Similar to https://github.com/iiasa/message-ix-models/pull/250#pullrequestreview-2450593364, I'm hoping today's nightly CI test failure also just arises from an incompatibility of using an older GAMS version but a newer GAMS license for this kind of run. This PR thus bumps the GAMS version in the nightly workflow and temporarily enables them to make sure the fix works.

## How to review

## PR checklist

- [x] Update tests; coverage checks both ✅
- ~[ ] Add, expand, or update documentation. ~ Just CI.
- ~[ ] Update release notes. ~ Just CI.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -11,7 +11,10 @@ on:
   - cron: "0 5 * * *"
 
 env:
-  GAMS_VERSION: 29.1.0
+  # Version used until 2024-11-20; disabled
+  # GAMS_VERSION: 29.1.0
+  # First version including a macOS arm64 distribution
+  GAMS_VERSION: 43.4.1
   # See description in lint.yml
   depth: 100
 
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -14,6 +14,8 @@ on:
       - '**.gms'
       - '**.gpr'
       - '**.gdx'
+    types:
+     - labeled
 
 env:
   # Version used until 2024-11-20; disabled
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -9,6 +9,11 @@ on:
   schedule:
   # 05:00 UTC = 06:00 CET = 07:00 CEST
   - cron: "0 5 * * *"
+  pull_request_target:
+    paths:
+      - '**.gms'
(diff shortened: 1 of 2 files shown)


glatterf42:
I've been looking at events that trigger CI runs and I've stumbled upon [pull_request_target](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request_target-workflow-based-on-files-changed-in-a-pull-request). If I'm not mistaken, this can close #603 since the nightly tests will now automatically run on every PR that changes any GAMS file (`.gms`, `.gpr`, and `.gdx` for now).

message_ix/tests/test_nightly.py
glatterf42, line 1: This test seems to not be considered for our coverage report, so new lines here are considered untested. I'd argue this should not stop this PR, though.

.github/workflows/nightly.yaml
```diff
@@ -9,9 +9,19 @@ on:
   schedule:
   # 05:00 UTC = 06:00 CET = 07:00 CEST
   - cron: "0 5 * * *"
+  pull_request_target:
+    paths:
+      - '**.gms'
+      - '**.gpr'
+      - '**.gdx'
+    types:
+     - labeled
```
glatterf42, line 18: Without this, our secrets could be extracted by any workflow changing GAMS files.

.github/workflows/pytest.yaml
```diff
@@ -6,6 +6,9 @@ on:
   # 05:00 UTC = 06:00 CET = 07:00 CEST
   schedule:
   - cron: "0 5 * * *"
+  pull_request_target:
+    types:
+      - labeled
```
glatterf42, line 11: This is trying to enable fork-workflows for the normal tests, too.

.github/workflows/nightly.yaml
glatterf42: Note: the tests passed before, so I disabled them again. They won't run when we apply a label to this PR because it would need to clear both filters, i.e. change some GAMS files *and* get a label to run.

khaeru (approved):

PR #893, opened 2024-11-21
Allow moderated secrets usage on PRs from forks

khaeru:
This PR implements the approach recommended by [this article](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) from Github's ‘Security Lab’ to allow the usage of secrets on PRs from forks of the `message_ix` repository.

- Add a "receive" CI workflow.
  - Run on all PRs into `main`.
  - Fail if (a) the PR is from a fork and (b) it does not have a certain label, currently "safe to test".
- Adjust the "pytest" CI workflow:
  - Add a `workflow_run:` event trigger. Invoke this trigger every time the "receive" workflow completes (success or failure).
  - Skip the main `pytest` and `tutorials` jobs if the "receive" workflow failed.
  - Invoke actions/checkout with the right options to use the fork PR branch.

With this merged, it would be up to some user with ‘write’ permissions on the current repo to give a first look-over of any PR from and then add the "safe to test" label. (The label could also be removed again at any time.)

## How to review

[Per the documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run):

> This [workflow_run] event will only trigger a workflow run if the workflow file is on the default branch.

This means we will need to:
- Merge this PR.
- Create or update another PR (#770 was suggested).
- Confirm that the `pytest` workflow is invoked and runs successfully; if not, open a follow-up PR to adjust the workflow(s).

## PR checklist

- ~Add or expand tests; coverage checks both ✅~ N/A; CI changes only
- ~Update release notes.~

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -6,6 +6,9 @@ on:
   # 05:00 UTC = 06:00 CET = 07:00 CEST
   schedule:
   - cron: "0 5 * * *"
+  workflow_run:
+    workflows: [ "Receive pull request" ]
+    types: [ completed ]
 
 # Cancel previous runs that have not completed
 concurrency:
@@ -23,6 +26,10 @@ env:
 
 jobs:
   pytest:
+    if: >
+      github.event_name != 'workflow_run' ||
+      github.event.workflow_run.conclusion == 'success'
+
     strategy:
       matrix:
         os:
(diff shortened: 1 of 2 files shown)


glatterf42 (approved):
Thanks, this looks promising to me :)

khaeru:
@glatterf42 FYI this appears not to have worked exactly as intended 🥲 

When I opened #894:
- [This](https://github.com/iiasa/message_ix/actions/runs/12066855919) workflow run on `main` was triggered —which should not have happened, or it should have exited early because that is not a PR from a fork.
- It failed, because this value https://github.com/iiasa/message_ix/blob/433868d874533b0b6741c8f54d18b037890265a6/.github/workflows/pytest.yaml#L82
  …appears to be a JSON structure rather than the desired string "iiasa/message_ix".

I'll make a note to come back and revisit this.

PR #894, opened 2024-11-28
Test against ixmp4

khaeru:
This **began** as a minimal PR to check that the message_ix test suite runs using ixmp4 storage (i.e. without the ixmp.JDBCBackend and underlying Java code).

It was then substantially expanded by @glatterf42 such that (most of) the message_ix test suite and all of the tutorials are confirmed to run using the IXMP4Backend.

Some things not in scope for this PR:
- Changes to the behaviour/API of message_ix.Scenario.
- Migrating large chunks of the `ixmp` code (including its test suite) to `message_ix`.
- User-facing documentation explaining the changes and how to use the new backend with message_ix; this will be handled by #939 and a follow-up PR.

## Details

In order to achieve the testing, it:

- Installs from ~the branch for iiasa/ixmp#552~ the ixmp `main` branch that:
  - Provides a minimal 'shim' layer between the behaviour of ixmp.Platform, ixmp.TimeSeries, ixmp.Scenario and the underlying ixmp4 API.
  - Ensures that all of the behaviour of those 3 classes continues to pass the ixmp test suite when using ixmp4 storage. (message_ix and downstream/user code rely on that behaviour.)
  - Will (but does not yet, as of 2024-11-28) [parametrize](https://docs.pytest.org/en/stable/how-to/parametrize.html) the `ixmp.testing.test_mp` fixture to yield either the current ixmp.JDBCBackend-backed test Platform, or an ixmp4-backed test Platform.
- Confirms all of the behaviour of message_ix.Scenario continues to pass the `message_ix.tests` suite, with *either* storage option. (message-ix-models and downstream/user code rely on this behaviour.)

The branch is used by iiasa/message-ix-models#257.

Implementation/to dos:
- As a consequence of parametrizing `ixmp.testing.test_mp`, all tests in `message_ix.tests` that use the `test_mp` fixture will run *twice*: once with JDBCBackend, and once with ixmp4.
- Some other tests may not be parametrized automatically by the upstream changes:
  - [x] Identify these.
  - [x] Parametrize them manually.
- [x] Rebase this branch frequently.

## How to review

- Read the diff.
- Note the CI checks all pass.

## PR checklist

- [x] Merge upstream PRs including iiasa/ixmp#552, iiasa/ixmp#575.
- ~Add, expand, or update documentation.~ See above.
- [x] Update release notes.


glatterf42:
I tried adapting the branch used for ixmp in the Pytest workflow file. While the changes in https://github.com/iiasa/ixmp/pull/557 are not enough to pass ixmp's tests yet, I wanted to see how well they fare in the message_ix tutorials. Unfortunately, the pytest run uses the workflow file from `main`, it seems, so we can't use the CI here to test against the changes in ixmp.

khaeru:
> Unfortunately, the pytest run uses the workflow file from `main`, it seems, so we can't use the CI here to test against the changes in ixmp.

That's why we have these lines:
https://github.com/iiasa/message_ix/blob/50725215eea5d49b244895aca05fcb535d243380/.github/workflows/pytest.yaml#L4-L6

I believe if these are uncommented then the workflow from this branch will be used. Since the branch is first-party, it will have access to secrets.

khaeru:
At iiasa/ixmp#557, @glatterf42 asked:

> For example: `ixmp4.Run.clone()` doesn't support `first_model_year` […]. How crucial are these features?

To expand on https://github.com/iiasa/ixmp/pull/557#issuecomment-2618951852: I briefly played around with #873 last year. Via that PR or something else, there are multiple possible solutions. I'm not sure right now which achieves our goal of minimizing breaking changes:
1. For one `message_ix.Scenario.clone()` can check if it's talking to JDBCBackend or IXMP4Backend:
   - If JDBCBackend, it passes on the `first_model_year` argument to the Java code, preserving the current behaviour exactly.
   - If IXMP4Backend, it asks for a simple clone, then invokes `shift_period()` to modify the data.
2. Another possibility is that `message_ix.Scenario.clone()` *never* passes down a first_model_year parameter—even to JDBCBackend—and *always* invokes `shift_period()` itself.

While (1) is not 100% 'clean' in the sense that message_ix.Scenario *shouldn't* need to know or care what backend it is talking to, the limitations of the Java code (see the other PR) might make (2) unworkable, in which case (1) would be, I think, the simplest compromise option.

glatterf42:
> I believe if these are uncommented then the workflow from this branch will be used. Since the branch is first-party, it will have access to secrets.

Even with these lines uncommented, the workflow from `pull_request` got cancelled, so only `pull_request_target` ran, which used the `main` version of ixmp again. However, this is just FYI, don't spend too much time on this; I'm testing the tutorials locally, anyway.

glatterf42:
This is slowly taking shape. All new functions have docstrings now and should be covered by some tests, though notably I have not tested some edge cases yet because the functions we're calling here are mostly for internal use and should not need that.
Before the tests here can pass, we'll need a proper parametrization of `test_mp` in ixmp, I think, for JDBC and IXMP4Backend. We do have one already, but right now, this simply means all tests requesting `test_mp` are running twice. So I'll take a look at that (and some cleanup for the ixmp branches) next.

For this PR, we mainly miss docs pages and release notes afterwards, I think :)
I'm not sure if our changes her should get there own dedicated docs page (I assume they do) and which form this should take.
Also, we should decide on what to do with the data in `scenario_data`. I invested a few hours to copy all information contained in ixmp_source thoroughly, but now we're not using that (preferring `models.py` instead). If we don't need that (and similarly several additions in `Scenario.__init__()`), we can do some more cleanup.

glatterf42:
Our workflow file currently reads:

```yaml
  pull_request:
    branches: [ main ]
  # pull_request_target:
  #   branches: [ main ]
  #   types: [ labeled, opened, reopened, synchronize ]
```

For commit https://github.com/iiasa/message_ix/pull/894/commits/42e12075c17baa688bc9c2df6a61993f313d9ffb (TEMPORARY Add ixmp4 dependencies to RTD requirements), this was running as expected: running on the `pull_request` trigger and installing the desired branch of ixmp4. For https://github.com/iiasa/message_ix/pull/894/commits/77f162bdcac22d762e815d792bb8b2928022b69e (Bump version of gams used for CI), though, the tests ran on the `pull_request_target` trigger again and failed because the latest release of ixmp4 doesn't contain some of the necessary changes.
I'm not entirely sure why this is apparently determined at random with a race condition.

message_ix/tests/test_integration.py
```diff
@@ -70,9 +77,13 @@ def test_run_remove_solution(test_mp, request):
     )
 
 
-def test_shift_first_model_year(test_mp, request):
+# TODO IXMP4Backend doesn't support clone(shift_first_model_year=...) yet
```
khaeru, line 80: I don't think this should be the responsibility of ixmp, as that is a stack level violation. See #873 and #254.

message_ix/tests/test_integration.py
```diff
@@ -49,9 +51,14 @@ def test_run_clone(tmpdir, request):
     )
 
 
-def test_run_remove_solution(test_mp, request):
+# TODO ixmp4-remove_solution() doesn't seem to remove the timeseries values for
```
khaeru, line 54: If I read the test correctly, the time-series/IAMC-structured data should not be changed when model solution data is removed. So I don't think this requires an upstream adjustment.

message_ix/core.py
```diff
@@ -59,10 +71,58 @@ def __init__(
         # Scheme returned by database
         assert self.scheme == "MESSAGE", self.scheme
 
+        if sys.version_info >= (3, 10):
+            from ixmp.backend.ixmp4 import IXMP4Backend
+
+            from message_ix.util.scenario_setup import add_default_data
+
+            if isinstance(self.platform._backend, IXMP4Backend):
```
khaeru, line 79: In general here I'd favour (and will add) a utility function that collapses these two conditionals into one.

khaeru:
> Since the main goal for 3.11 is that the tutorials run with IXMP4Backend, I think we ought to test this.

Now that this is running, the test logs (e.g. [here](https://github.com/iiasa/message_ix/actions/runs/14955064145/job/42009609120?pr=894#step:10:332)) point to at least 2 changes needed upstream. So I'll open a new PR in ixmp for those adjustments.

khaeru:
I'm also running into aforementioned issue where GHA still seems to be triggering the 'pull_request_target' event even though we've supposedly disabled it on this branch. For now I am cancelling those runs and allowing/restarting the 'pull_request'-triggered runs.

khaeru (commented):
Okay this is "approved" (even though, as the initiator, I can't actually approve it)! [Here](https://github.com/iiasa/message_ix/actions/runs/14968744495) is a passing run of the whole test suite.

Before merging, I'll rebase to drop the TEMPORARY commit, but not wait for the CI to run again.

Some points to follow up on:
- We will make 1+ PRs to document the state of ixmp4 support in the coming release, how to handle this during installation.
- The added files tests/data/reporter*graph.txt can be reduced a little because one is a strict subset of the other.
- We can likely do the following:
  - Split `message_ix.models` into submodules like `message_ix.models.common`, `message_ix.models.message`, `message_ix.models.macro` —with the latter each containing one class and associated utility functions.
  - Consolidate structural information from `.util.scenario_data` and behaviour from `.util.scenario_setup` into `message_ix.models.message`, since these are the structure/behaviour of a "MESSAGE-scheme Scenario" (as opposed to, say, a MACRO Scenario).
- We can use pytest-profiling to investigate the causes of slower runtimes for various ixmp4-backed tests versus their jdbc-backed equivalents.

PR #898, opened 2024-12-09
Bump Python version for tutorials

glatterf42:
We're seeing the same error in our CI again as we did previously when GitHub runner-images first changed `ubuntu-latest` to Ubuntu 24.04. They are rolling out the same change again now. However, we have adopted Python 3.13 in the meantime and in the ixmp test suite, this seems to work, while Python 3.12 produces the same error with Ubuntu 24.04. 3.11 works with that image, too, so the error might be specific to Python 3.12. Thus, this PR bumps the Python version we use for the tutorial tests in hopes of enabling them to run again on `ubuntu-latest`.

If this works, we might be able to adjust the ixmp test suite to only use Ubuntu 22.04 with Python 3.12.

See https://github.com/IRkernel/IRkernel/issues/747 and https://github.com/r-lib/actions/issues/932 for the issues I opened about this back then.

## How to review

## PR checklist

- [x] Update test suite; coverage checks both ✅
- ~[ ] Add, expand, or update documentation.~ Just CI bump.
- ~[ ] Update release notes.~ Just CI bump.

### DIFF
--- a/message_ix/tests/test_report.py
+++ b/message_ix/tests/test_report.py
@@ -204,30 +204,34 @@ def test_reporter_as_pyam(caplog, tmp_path, dantzig_reporter):
     assert not any(c in df2.columns for c in ["h", "m", "t"])
 
     # Variable names were formatted by the callback
-    reg_var = pd.DataFrame(
-        [
-            ["seattle", "Activity|canning_plant|production"],
-            ["seattle", "Activity|transport_from_seattle|to_new-york"],
-            ["seattle", "Activity|transport_from_seattle|to_chicago"],
-            ["seattle", "Activity|transport_from_seattle|to_topeka"],
-            ["san-diego", "Activity|canning_plant|production"],
-            ["san-diego", "Activity|transport_from_san-diego|to_new-york"],
-            ["san-diego", "Activity|transport_from_san-diego|to_chicago"],
-            ["san-diego", "Activity|transport_from_san-diego|to_topeka"],
-        ],
-        columns=["region", "variable"],
-    ) if sys.version_info >= (3, 10) else pd.DataFrame(
-        [
-            ["san-diego", "Activity|canning_plant|production"],
-            ["san-diego", "Activity|transport_from_san-diego|to_chicago"],
-            ["san-diego", "Activity|transport_from_san-diego|to_new-york"],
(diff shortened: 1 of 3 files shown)


glatterf42:
All tutorial tests are now failing, but notably _after_ the step that is failing in our CI suite. This seems to confirm that using Python 3.13 with Ubuntu 24.04 resolves that issue :)

The reason the tests are still failing is likely that pyam doesn't officially support Python 3.13 yet, which it can't until ixmp4 does. For ixmp4, all that is needed for that is a new release, which I will ask about right away. I even prepared a PR for pyam already.

glatterf42:
I have now also opened https://github.com/actions/runner-images/issues/11137 to see if this bug can be resolved on the runner-image level.

glatterf42:
As expected: with the release of pyam 3.0.0, which supports Python 3.13, the tutorial tests are passing again (using Python 3.13 for all OSs). Support for Pyam 3.13 requires a small adaptation in our test suite, which this PR also provides now.

measrainsey (approved):
LGTM, thanks @glatterf42 ! :)

message_ix/tests/test_report.py
```diff
@@ -203,18 +204,34 @@ def add_tm(df, name="Activity"):
...
+                ["seattle", "Activity|transport_from_seattle|to_chicago"],
+                ["seattle", "Activity|transport_from_seattle|to_topeka"],
+                ["san-diego", "Activity|canning_plant|production"],
+                ["san-diego", "Activity|transport_from_san-diego|to_new-york"],
+                ["san-diego", "Activity|transport_from_san-diego|to_chicago"],
+                ["san-diego", "Activity|transport_from_san-diego|to_topeka"],
+            ],
+            columns=["region", "variable"],
+        )
+        if sys.version_info >= (3, 10)
```
khaeru, line 221: Just to note for posterity—the approach I took to a similar failure in genno was this:
https://github.com/khaeru/genno/pull/155/files

Some advantages I'd claim for that approach:
- The `if` condition and comment identify that the difference here is attributable to behaviour of specific versions of `pyam-iamc`. The underlying version of Python is not directly responsible; only indirectly insofar as it leads to different versions of `pyam-iamc` being available.
- Instead of duplicating the same expected data, `.sort_values()` is called, so the code is shorter.

PR #904, opened 2025-01-08
Fix repository name in CI workflow

glatterf42:
This PR aims to eventually fix how we run our CI workflows from forks. We want the tests to run only once and to be able to access our main repo's secrets if labelled correctly. 

First, I'll not add any label to this PR, so it should not be able to access the secrets, which should make the tests fail. 
Then I'll add the label and push something, which should trigger exactly one workflow: the "Receive PR" workflow should complete successfully, triggering one run of the tests with access to our secrets.

I'm not convinced this will work already because the trigger for `pytest.yaml` is also "on PR to `main`", so I'd have to look into combining this with "and not from within the head repo".

## How to review

Look at the triggered CI runs and check they align with our intentions.

## PR checklist

### DIFF
--- a/message_ix/tests/test_report.py
+++ b/message_ix/tests/test_report.py
@@ -1,7 +1,7 @@
 import logging
 import re
-import sys
 from functools import partial
+from importlib.metadata import version
 from pathlib import Path
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


glatterf42:
So the "Receive PR" workflow fails as expected since it's missing the label. However, I still triggered the "Pytest" workflow because the target is `iiasa:main`. This makes me think: the "Receive PR" workflow should be the only one with a trigger of "on PR to `main`". In this case, it would fail, but with a label or from within the IIASA repo, it should pass, triggering the "Pytest" workflow.

glatterf42:
On it. What confuses me more is that the tests are actually passing, even though they should not have access to our secrets. Could this be because one doesn't need a license to access the cached GAMS installation?

glatterf42:
Pushing a commit without labeling the PR now triggers only the "Receive PR", "license/cla", and "Build package" tests.

glatterf42:
Adding the label gets "Receive PR" to pass, but the view of the tests here does not show that any of the other tests would be running now.

The commit I used to trigger the update was the improvement @khaeru suggested [here](https://github.com/iiasa/message_ix/pull/898#discussion_r1905113582).

glatterf42:
Maybe these statuses are only reported back once the whole workflow finishes?
But even that seems to have happened already: https://github.com/iiasa/message_ix/actions/runs/12668365243/job/35303644902
It looks to me like the "Receive PR" workflow triggers the "Pytest" workflow as it is on `IIASA:main`, which still fails due to `github.event.workflow_run.repository` being an object (where a string is expected by the checkout action).

khaeru:
We developed the alternate approach at #909.

PR #808, opened 2024-03-25
Add an option to solve MACRO concurrently for all regions

daymontas1:
This PR introduces the option to solve for all regions concurrently in MACRO. A parameter (''solve_param'') has been added to ''MESSAGE-MACRO'' to control the solving mechanism for MACRO. Setting ''solve_param'' to 1 (the default value) triggers sequential solving across regions, whereas setting it to 2 enables concurrent solving.

An if statement has also been introduced in ''macro_solve'' that dictates whether sequential or concurrent solving is applied in MACRO, based on the value of the ''solve_param'' parameter defined in ''MESSAGE-MACRO_run''. The default solving is sequential (''solve_param''=1), so if the user doesn't update the solving mechanism, MACRO is solved as it used to be. In terms of efficiency, the concurrent solving is subtly more efficient. 

For the concurrent solving mechanism (''solve_param''=2), all nodes are activated from the beginning, thereby enabling simultaneous solving for all regions. This setting allows for factoring in interregional interactions into the optimization process. We want this concurrent optimization option, as we intend to incorporate interregional feedback mechanisms into MACRO, e.g., interregional investments. Given that in the current version of MACRO regions are independent, both solving mechanisms should produce the same results.

## How to review
**Required:** It must be verified that both versions produce the same results (considering there are no interregional interactions in the current version of MACRO).

If the changes are accepted, the model's documentation must be updated to include the concurrent option and instructions on how to trigger it. Additionally, it may be worthwhile to consider making the concurrent solving mechanism for MACRO the default, in case it reduces solving times. 

## PR checklist

- [ ] Continuous integration checks all ✅
  
- [x] Add or expand tests; coverage checks both ✅
- [x] Add, expand, or update documentation.
- [x] Update release notes.

PR #905, opened 2025-01-08
Fix workflow runs from forks

glatterf42:
As discovered in https://github.com/iiasa/message_ix/pull/904#issuecomment-2577300711, it seems we need to bring to `main` a version of the "Pytest" workflow that doesn't run on all PRs to `main` and that uses the correct data for the checkout action when being dispatched from the "Receive PR" workflow. This PR cherry-picks the two commits from #904 that achieve this. 

## How to review

- ~Read the diff and note that the CI checks all pass.~
- Note that this PR should trigger the "Pytest" workflow only once, after successfully completing the "Receive PR" workflow.

## PR checklist

- ~Continuous integration checks all ✅~ These fail because of the underlying issue. Merging the changes will address the issue.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -79,7 +79,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         fetch-depth: ${{ env.depth }}
-        repository: ${{ github.event.workflow_run.repository }}
+        repository: ${{ github.event.workflow_run.repository.fullname }}
         ref: ${{ github.event.workflow_run.head_branch }}
 
     - name: Fetch tags (for setuptools-scm)
@@ -166,7 +166,7 @@ jobs:
       if: github.event_name == 'workflow_run'
       uses: actions/checkout@v4
       with:
-        repository: ${{ github.event.workflow_run.repository }}
+        repository: ${{ github.event.workflow_run.repository.fullname }}
         ref: ${{ github.event.workflow_run.head_branch }}
 
     - uses: actions/setup-python@v5
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -1,8 +1,6 @@
 name: Test
 
 on:
-  pull_request:
-    branches: [ main ]
   # 05:00 UTC = 06:00 CET = 07:00 CEST
   schedule:
   - cron: "0 5 * * *"


glatterf42:
Okay, even though this PR is from within IIASA's repo, the only "Pytest" run that is triggered uses the version currently on `main`, so it's still failing. I find it curious that these failing tests are not registered here, instead showing the required tests as "expected" still. This makes me wonder if they will update to "successful" once all of them are.

glatterf42:
From [this GitHub blog](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/), it sounds like we might have to save things like coverage and test failure rate (or so) as artifacts and comment them on the PR manually:

> The workflow processing the PR should then store any results like code coverage or failed/passed tests in artifacts and exit. The following workflow then starts on workflow_run where it is granted write permission to the target repository and access to repository secrets, so that it can download the artifacts and make any necessary modifications to the repository or interact with third party services that require repository secrets (e.g. API tokens).

This sounds less than ideal. Alternatively, I found [this suggestion](https://michaelheap.com/access-secrets-from-forks/): use the `triggering_actor` to decide if the workflow has access to secrets. This way, just anyone (without write access to the main repo) would open the PR and the workflow fails (because it doesn't have access to secrets), but once re-run from someone with the right permissions, the workflow is allowed to access the secrets. 
This is still prone to human error, but should be fine in our case, I think. We don't often have time-critical contributions, almost never ones from outside IIASA. So we could double and triple check any suspicious code before re-running a workflow.

khaeru:
> it sounds like we might have to save things like coverage and test failure rate (or so) as artifacts and comment them on the PR manually

Urgh, that's definitely too complicated.

> Alternatively, I found [this suggestion](https://michaelheap.com/access-secrets-from-forks/): use the `triggering_actor` to decide if the workflow has access to secrets.

This seems like a good alternative. If I understand, a maintainer would need to trigger a re-run each time, right? I think that is an acceptable compromise, given that we will not use this so frequently: we can continue to encourage teammates to use branches within the iiasa repo, so it would be only outside collaborators.

glatterf42:
Not even that: all colleagues from the MESSAGEix Devs team have write access, too. So even PRs from their forks would just run immediately, given their permissions. This would only affect PRs from people not on the team (or otherwise having write access to this repo), if I understand correctly.

khaeru:
Okay, that sounds great. Should we try to implement it via this PR?

glatterf42:
I tried doing that just now. I think for this to work as intended, the `github.event.workflow_run.head_branch` needs to point to the fork's branch, and I'm not sure it does that. It could also point to `main` as the head branch triggering the "Pytest" workflow (seeing that the "Receive" workflow is executed on `main`), in which case we would only be testing `main`, which does not seem meaningful. I'm not sure how best to test this (other than merging something to some kind of `main`, which could also be one of our fork's main, I guess).

khaeru:
As discussed, we'll drop the 3rd commit re: triggering_actor here; merge; check the effects; and then continue in a new PR.

khaeru (approved):

khaeru:
Okay, after doing this, I rebased both #904 and #808 on main, triggering CI to run. What I see:
- The "receive" workflow runs and does not fail.
- "Test" workflow runs are triggered. [Per the list](https://github.com/iiasa/message_ix/actions/workflows/pytest.yaml), they do not appear associated with the respective PRs.
- On the two PR pages, the required checks/jobs are all showing "Expected — Waiting for status to be reported". So they are not associated with the jobs in the triggered workflow.
- The "Test" workflow runs fail in the step "Check out message_ix (workflow_run)", for instance [here](https://github.com/iiasa/message_ix/actions/runs/12883550924/job/35918026294#step:3:56):
```
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=100 origin +refs/heads/macro_update*:refs/remotes/origin/macro_update* +refs/tags/macro_update*:refs/tags/macro_update*
  The process '/usr/bin/git' failed with exit code 1
  Waiting 14 seconds before trying again
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=100 origin +refs/heads/macro_update*:refs/remotes/origin/macro_update* +refs/tags/macro_update*:refs/tags/macro_update*
  The process '/usr/bin/git' failed with exit code 1
  Waiting 11 seconds before trying again
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=100 origin +refs/heads/macro_update*:refs/remotes/origin/macro_update* +refs/tags/macro_update*:refs/tags/macro_update*
  Error: The process '/usr/bin/git' failed with exit code 1
```
- I guess this is occurring because, e.g. there is no `remotes/origin/macro_update` in the iiasa/message_ix repo; within that repo, the particular remote (glatterf42/message_ix or daymontas1/message_ix) is not known as 'origin'.

I guess this means back to the drawing-board! Will link to this comment from any follow-up PR(s).

Issue #580, opened 2022-04-13
Support autogeneration of RES data/diagram


gidden:
# Request

In many contexts, understanding the underlying reference energy system (RES) structure of a given scenario is needed.

Given the RES structure in a dataframe, we can then use this directly and generate visuals to support debugging and other concerns.

# Example

Generally, this can be done by intelligently interrogating `input` and `output` parameters for a scenario.

Below is an example of two functions that could help

```
def find_inp_link(s, tec, filters):
    filters = filters.copy()
    filters['technology'] = tec
    df = s.par('input', filters=filters)
    df = df[['commodity', 'level']].drop_duplicates()
    return df

def find_outp_link(s, inp_link, filters):
    filters = filters.copy()
    filters = dict(**filters, **inp_link)
    df = s.par('output', filters=filters)
    df = df[['technology']].drop_duplicates()
    return df
```

Consider the example of `LNG_exp` in current versions of the global model

```
filters = {'node_loc': 'R15_NAM', 'year_act': 2020}
tec = 'LNG_exp'
inp_link = find_inp_link(s, tec, filters).iloc[0].to_dict()
>>> {'commodity': 'LNG', 'level': 'primary'}
```
then
```
find_outp_link(s1, inp_link, filters).iloc[0].to_dict()
>>> {'technology': 'LNG_prod'}
```
and so on

khaeru:
[Graphviz](https://www.graphviz.org/gallery/) and [NetworkX](https://networkx.org/documentation/stable/auto_examples/index.html) (including via graphviz) provide useful features. The Python graphviz package is what's used under the hood by dask (https://docs.dask.org/en/latest/graphviz.html#visualize-task-graphs), thus also by genno and `{ixmp,message_ix}.reporting`.


To produce the kind of visualization that we often use, the code would need to, among other things:
- Receive configuration setting the *order* of items in the `level` set. (The ixmp backend does not make strong promises about the ordering of set item; I don't think we can count on the items being in order in our scenarios; and in any case the notion of order is ambiguous once we have some levels pertaining to energy, and others to water, materials, etc.)
- Group the (commodity, level) nodes into multiple sets based on (level).
- Group the (technology) nodes into multiple sets based on which input/output levels they use. E.g. (technology) nodes that all input from (commodity, "primary") and output to (commodity, "secondary") would need to be grouped.
[...]

PR #906, opened 2025-01-10
Add .tools.graph_viz_tool

macflo8:
Following on the 2024-12-12 MESSAGE meeting and #580, this PR aims to add a tool that can visualise a MESSAGEix model structure as a graph.

The initial commit adds the standalone tool developed by @macflo8, which can serve as a basis for a tool that is fully integrated in the `message_ix` code base. The standalone tool offers two visualisation options:
- a static "RES-style" visualisation that relies on the `graphviz` package that is similar to the VBA tool mentioned in https://github.com/iiasa/message_ix/issues/580#issuecomment-2538468377
- an interactive graph that relies on the `pyvis` package that is more suited to visualise a subset of the model

## How to review

- The general methodology should be reviewed to make sure the tool is computationally efficient.
- The design should be reviewed to make sure it fits with the patterns used in `message_ix`.
- The ease of use should be reviewed, to increase the probability that current (and future) `message_ix` users will actually use it. 

## PR checklist

- [ ] Update release notes.

### DIFF
--- a/message_ix/tools/graph_viz_tool/dynamic_graph.py
+++ b/message_ix/tools/graph_viz_tool/dynamic_graph.py
@@ -1,0 +1,106 @@
+from typing import List
+
+import pandas as pd
+import pyvis as pv
+from message_ix_models.util import load_package_data
+
+pyvis_hierarchy_config = {
+    "enabled": True,
+    "direction": "LR",
+    "blockShifting": True,
+    "edgeMinimization": False,
+    "parentCentralization": False,
+    "levelSeparation": 150,
+    "treeSpacing": 200,
+}
+
+T_D_TECS_COLOR = "#f7cf26"
+
+
+def load_technology_annotations():
+    tec_codes = load_package_data("technology")
+    tec_descriptions = {k: v.get("description") for k, v in tec_codes.items()}
+    return tec_descriptions
+
+
+class DynamicGraph:
+    def __init__(self, commodities: List[str]):
+        self.graph = pv.network.Network(
+            layout="reingold_tilford",
+            select_menu=True,
+            directed=True,
+            heading="commodity flow graph",
+        )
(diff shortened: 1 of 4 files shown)


khaeru:
Thanks! Let's work together on this branch to address the TODOs from the description.

Some thoughts before jumping in:
1. At the 2025-01-09 MESSAGE team meeting, @OFR-IIASA reminded of a suite of checks that he and @gidden developed for common issues in constructing MESSAGEix models. I mentioned #506, still a draft. One of the examples we discussed was the case where there is (potentially) input from/output to a (level, commodity, node, time) that does not have a corresponding source/sink. (This is not _necessarily_ an error, as some such technologies may never need to be active.) This check can be implemented by inspection of the _same_ abstract graph data structure that is used here to make a graphical visualization. 
2. In #770, the data used is a similar abstract graph, except a *weighted* graph where the edge weights are the magnitudes of the respective flows. There it is expressed in the IAMC structure particular to pyam.
3. As revealed by #770, the RES structure can potentially differ by MESSAGE node and year (period). (This is potentially confusing as a modeling practice, thus discouraged, but the framework does not prevent it.) So we may need to provide similar options here, e.g. select one (node, year) or sum across many. 
4. We should be careful in documentation (comments, docstrings, docs, files) to distinguish the "MESSAGE node [set/dimension]" from a "graph node", the common mathematical concept.

khaeru:
See also [this Slack thread](https://iiasa-ece.slack.com/archives/CD0GBHHA4/p1740668613230309?thread_ts=1738919098.704119&cid=CD0GBHHA4) in which @Tyler-lc has shared scripts to create interactive Sankey diagrams using the Plotly and Holoviz packages. This adds up to 3 different methods/diagrams that could potentially be integrated here, in addition to the one already merged/released.

PR #912, opened 2025-01-22
Fix `PRICE_EMISSION` definition

glatterf42:
In preparation of the 3.10 release, @khaeru and I noticed that the [original PR](https://github.com/iiasa/message_ix/pull/726) to fix `PRICE_EMISSION` (address #723) has become a little cluttered with discussions of how to create a proper test for the calculation update. Meanwhile, @volker-krey suggested that we should just bring the fix to `main` and open a follow-up issue to create the test later. Thus, this PR adds all changes from #726 that seem to be necessary to fix the calculation and make the tests pass as they are. 

>[!CAUTION]
>The error with the current calculation of `PRICE_EMISSION` went undetected so long because our tests were not stringent enough. Hence our initial insistence that #726 contain the exact kind of test to prevent the issue from coming up again. Such a test is missing from this PR on purpose, but that makes it incomplete in that sense. We need to maintain the sense of urgency that this test should be added as quickly as possible. 
> The only reason we are considering merging this without the test is because both @OFR-IIASA and @yiyi1991 manually confirmed that this fix works on sufficiently complex scenarios.

## How to review

  does in the future.
- If you know of further places in our docs where the `PRICE_EMISSION` calculation is explained and needs updating, please share that.

## PR checklist

- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -1,8 +1,24 @@
+.. _v3.10.0:
+
 Next release
 ============
 
-GitHub community guidelines
----------------------------
+Migration note
+---------------
+
+.. _v3.10.0-migrate-1:
+
+1. For scenarios with :doc:`periods </time>` that have 2 or more different ``duration_period``, users should expect that values for the solution variable ``PRICE_EMISSION`` will change compared to version 3.9.0 and earlier.
+
+   **Only** such scenarios are affected.
+   For example, if ``duration_period`` is 5 years for some periods in the ``year`` set, and 10 years for others, then ``PRICE_EMISSION`` values will change.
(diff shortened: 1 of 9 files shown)


khaeru:
Thanks for excerpting a clean version of these changes.

> - [ ] Update release notes.

I will join to try to write an appropriate summary that alerts users to the nature of the change and its expected impact.

Per the commit messages, I'm not sure yet that adopting [“conventional commits”](https://www.conventionalcommits.org/en/v1.0.0/)—either optionally or as mandatory for all contributions—has benefit/cost > 1.

Can we either:
- Modify this text to explain what alternate format is accepted https://github.com/iiasa/message_ix/blob/4c1447333b5c10ba6e0396c9ceaf4aa20d95978f/doc/contributing.rst?plain=1#L185
- or, rewrite the messages for now?

PR #919, opened 2025-02-24
Fix "install from conda forge" CI  broken by release 3.10.0

glatterf42:
Unfortunately, our release 3.10.0 broke our own "install from conda forge" CI _again_. This PR will eventually fix that.

We see this error upon test collection in the CI run:
```python
 ImportError while importing test module 'D:\a\_actions\iiasa\actions\main\setup-conda\Anaconda3\envs\testenv\Lib\site-packages\message_ix\tests\tools\test_sankey.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
D:\a\_actions\iiasa\actions\main\setup-conda\Anaconda3\envs\testenv\Lib\site-packages\message_ix\tools\sankey.py:5: in <module>
    from pyam.str import get_variable_components
E   ModuleNotFoundError: No module named 'pyam.str'

During handling of the above exception, another exception occurred:
D:\a\_actions\iiasa\actions\main\setup-conda\Anaconda3\envs\testenv\Lib\importlib\__init__.py:88: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
D:\a\_actions\iiasa\actions\main\setup-conda\Anaconda3\envs\testenv\Lib\site-packages\message_ix\tests\tools\test_sankey.py:7: in <module>
    from message_ix.tools.sankey import map_for_sankey
D:\a\_actions\iiasa\actions\main\setup-conda\Anaconda3\envs\testenv\Lib\site-packages\message_ix\tools\sankey.py:7: in <module>
    from pyam.utils import get_variable_components
E   ImportError: cannot import name 'get_variable_components' from 'pyam.utils' (D:\a\_actions\iiasa\actions\main\setup-conda\Anaconda3\envs\testenv\Lib\site-packages\pyam\utils.py)
```

We introduced the sankey tool in this release and it already includes a clause that tries to handle older versions of pyam (that adjusted the location from which to import `get_variable_components`. However, the CI run currently installs pyam 0.7.0, which was released more than 4.5 years ago, for some reason. 

Unfortunately, conda doesn't seem to offer any option to explain its dependency resolution process. I didn't find any reason when [comparing version 0.7.0 and 0.8.0 of pyam](https://github.com/IAMconsortium/pyam/compare/v0.7.0...v0.8.0): the pytest dependency received a pin, but that didn't make it to the conda-force build recipe. In fact, the [current recipe](https://github.com/conda-forge/pyam-feedstock/blob/main/recipe/meta.yaml) only contains a pin of numpy to < 2.0, which I would not expect to effect anything from four years ago. 

I'll keep on looking for a reason, but in the meantime, this PR attempts to install pyam 3.0.0 explicitly in the hopes that this will either work or yield some details on why not.

## How to review

## PR checklist

- ~[ ] Add, expand, or update documentation.~ Just CI
- [x] Rebase removing "TEMPORARY" commits.

### DIFF
--- a/.github/workflows/conda.yaml
+++ b/.github/workflows/conda.yaml
@@ -87,7 +87,9 @@ jobs:
 
         # Also install pytest and packages required for testing. pip/PyPI
         # supports this via message_ix[testing], but conda does not.
-        conda create --quiet --name testenv message-ix pytest asyncssh sphinx
+        # Pin Python version to enable pyam versions relied on by the sankey tool;
+        # remove once pyam supports the latest Python
+        conda create --quiet --name testenv message-ix pytest asyncssh sphinx python=3.12
         conda list --name testenv
 
     - name: Check CLI commands and run test


glatterf42:
https://github.com/conda-forge/pyam-feedstock/pull/69 will fix this issue, after which we can use this PR or the schedule tests to confirm that everything works again. Note that everything should work fine without any change to `main`, so this PR should not be merged either way.

glatterf42:
Through re-running the tests, I found that the same problems exists for the ixmp4 feedstock, which still blocks pyam 3.0.0 to be used. https://github.com/conda-forge/ixmp4-feedstock/pull/29 should fix that.

glatterf42:
Still need to drop the "TEMPORARY" commit if we're satisfied, but the CI is passing again :)
I pinned Python to 3.12 for "install from conda forge" because even with pyam-feedstock and ixmp4-feedstock fixed, pyam 3.0.0 could not be installed because it pins numpy to < 2.0, which excludes Python 3.13. I hope this is fine in this case.
As soon as pyam gets another release (and they don't forget to adapt their feedstock dependencies), the numpy pin will be removed on their side and we can remove our Python pin here.

khaeru (approved):
Thanks. As we discussed, users in the interim can work around by:
- Installing message_ix & co from PyPI with Python 3.13.
- Using conda with Python 3.12, as the workflow does.

PR #922, opened 2025-03-10
Fix Code Quality test after pandas-stubs v2.2.3.250308

glatterf42:
v2.2.3.250308 of pandas-stubs broke our Code Quality CI check with an error about indexing a `pandas.DataFrame`. This PR makes the check pass again.

## Details

The error we received was

```python
message_ix/tools/add_year/__init__.py:84: error: Invalid index type "list[str]" for "_LocIndexerFrame[DataFrame]"; expected type "slice[Any, Any, Any] | ndarray[Any, dtype[integer[Any]]] | Index[Any] | list[int] | Series[int] | <6 more items>"  [index]
```

and the code in question was

```python
def mask_df(df: pd.DataFrame, index: str, count: int, value) -> None:
    """Create a mask for removing extra values from *df*."""
    df.loc[
        index,
        df.columns
        > (df.loc[[index]].notnull().cumsum(axis=1) == count).idxmax(axis=1).values[0],
    ] = value
```

I think I put in the type hint of `index: str` as I understood it, but this is not correct. All usage of `mask_df()` within our code calls the function with `index` being one element of an iteration over `year_df.index`, which is a `MultiIndex`, so `index` is actually a `tuple`. I have confirmed as much through running

```python
    print(index)
    print(type(index))
    reveal_type(index)
```

immediately before the `df.loc[...]` call. So first I tried changing the type hint to `index: tuple`, which yields the same error but with `index type "list[tuple[Any, ...]]"`. Through looking at the `print(index)` for our tests, it seems that the types of the tuple entries can only be `int` or `str`, so I tried specifying that. 
Unfortunately, this yields the same error, and now even the pylance extension in my code editor can't tell me what `.notnull()` etc are, but the code is working fine. In fact, making the second appearance of `index` just `index` (instead of `[index]`), results in an error.

So I tried to specify the type correctly as best I can and the code is working as it is, but mypy is not happy. I assume this is because pandas-stubs is not allowing our `index` type, even though that seems to be working fine. So for now, it seems like adding a `type: ignore` is the best way to handle this. I should also open an issue with pandas-stubs, though I don't know if I can set up a minimal example to reproduce this.

## How to review

## PR checklist

### DIFF
--- a/message_ix/tools/add_year/__init__.py
+++ b/message_ix/tools/add_year/__init__.py
@@ -76,12 +76,16 @@ def slice_df(
     return df.set_index(idx)
 
 
-def mask_df(df: pd.DataFrame, index: str, count: int, value) -> None:
+def mask_df(
+    df: pd.DataFrame, index: tuple[Union[int, str], ...], count: int, value
+) -> None:
     """Create a mask for removing extra values from *df*."""
     df.loc[
         index,
         df.columns
-        > (df.loc[[index]].notnull().cumsum(axis=1) == count).idxmax(axis=1).values[0],
+        > (df.loc[[index], :].notnull().cumsum(axis=1) == count)
+        .idxmax(axis=1)
+        .values[0],
     ] = value
 
 


khaeru:
Here we suffer a little from lack of documentation in `.add_year` (#494 or similar could help).

My impression:
- Generally, with `df = pd.DataFrame(...)`, one can select a subset of columns with `df[["a", "b", "c"]]` —in this case `["a", "b", "c"]: list[str]` and one expects there to be columns with names "a", "b", and "c".
- One can also use a `list[tuple]` or `list[Any]` like `["a", "b", ("c", 1)]`. This works because `pd.Index` works with most Hashables, so it's possible to have an Index in which one element is the tuple `("c", 1)`.
- In the first bullet above, suppose one did `df[("a", "b", "c")]`. This wouldn't work if `df` has columns named "a", "b", and "c", because it's not using those names; it's looking for a single column named `("a", "b", "c")`.
- In the code, perhaps whatever is calling `mask_df()` gets inferred by mypy as producing something like `index=("a", "b", "c")` but intends those as the names of 3 columns. In this case the fix would be to call `list(index)` to coerce it to the correct type, giving `df[["a", "b", "c"]]` or similar. This should also satisfy mypy.

glatterf42:
I'm not sure about this. I have come up with a minimal working example to reproduce this error message for pandas-stubs, but I also tried your suggestion just now: when I leave the type hint as `index: tuple[Union[int, str], ...]`, mypy doesn't care if I write `df.loc[[index]]` or `df.loc[list(index)]` because it doesn't like `list[int | str]` either. But if I change the type hint to just `index: tuple`, mypy is satisfied. However, pytest then fails with this error:

```python
FAILED message_ix/tests/tools/test_add_year.py::test_add_year[jdbc] - KeyError: "['tec' 'comm' 'mode' 2025 'level' 'year' 'year' '-'] not in index"
```

So it doesn't look like `list(index)` resolves this error.

khaeru:
Hm, I see. So `index` is more like a tuple of values for each level in a pd.MultiIndex, meant to identify just a single element in that MultiIndex?

Let me poke around a little in the calling code to see if that clarifies.

khaeru:
> So `index` is more like a tuple of values for each level in a pd.MultiIndex, meant to identify just a single element in that MultiIndex?
> 
> Let me poke around a little in the calling code to see if that clarifies.

It seems so. This code is probably not very performant, so IMO we should get rid it—perhaps also the entire containing function `interpolate_2d()`—rather than spending much or any time to type hint it correct. That's what #494 is for. But since we don't necessarily want to bring that back to the front of the queue right now, I'll approve the PR as-is.

AFAICT it would also satisfy mypy to simply drop the type hint on `mask_df(…, index=…)` altogether. Up to you which to choose.

khaeru (approved):

glatterf42:
Agreed that #494 would be great to have. Luckily, one of the pandas-stubs also came up with a solution: specifying both row and column indexes :)

glatterf42:
Merging without the macos checks; no idea why they weren't picked up by the runners. Tomorrow's CI will confirm that they are passing, too.

Issue #929, opened 2025-03-31
EMISSION_EQUIVALENCE is created for all `type_tec` irrespective of their purpose


OFR-IIASA:
### What happened?

The equation `EMISSION_EQUIVALENCE` is generated for all type_tec irrespective of their purpose. This means, that the type_tec, which are also used for the definition of the share constraints are also included. What this results in is that if a technology has an emission factor and that if this technology is included in a `type_tec`, then the variable `EMISS` is created for all `type_tec` entries, resulting potentially in a very large table which potentially creates memory issues.

The GAMS code related to shares has been needs to be cleaned up. Basically `type_tec` is used for multiple purposes without distinguishing them in a clean way. In principle the sets `type_tec_shares` and `type_tec_total` should be used for this but at present they are just defined as aliases of type_tec (in set_maps_def.gms). With minimal modifications they could be turned into proper sets that can then be used for filtering in the generation of equations without making the code hard to read.

The following alterations are needed to circumvent the over-generation of entries into the `EMISSION_EQUIVALENCE` formulation (thanks to @volker-krey for the suggestions; testing is still required):
`set_maps_def.gms` (lines 313-314):

```
*Alias(type_tec,type_tec_share);
*Alias(type_tec,type_tec_total);
Sets type_tec_share(type_tec)
     type_tec_total(type_tec) ;
```

`data_load.gms` (line 85, i.e. after loading the gdx file):
```
type_tec_share(type_tec) $ (SUM((shares, node_share, node, mode, commodity, level), map_shares_commodity_share(shares,node_share,node,type_tec,mode,commodity,level))) = YES ;
type_tec_total(type_tec) $ (SUM((shares, node_share, node, mode, commodity, level), map_shares_commodity_total(shares,node_share,node,type_tec,mode,commodity,level))) = YES ;
```
After this, the two sets include only type_tec entries that have share and total entries, respectively. Now, limiting the generation of EMISSION_EQUIVALENCE to type_tec entries that are not share related is required.
```
`model_core.gms` (line 1839, definition of equation):
EMISSION_EQUIVALENCE(node,emission,type_tec,year) $ (NOT type_tec_share(type_tec) AND NOT type_tec_total(type_tec))..
```

### Code Sample

```python

```

### What did you expect to happen?

`type_tec` used for definition of technologies related to share constraints should not be excluded from the equaiton `EMISSION_EQUIVALENCE`

### Versions

```shell

```

volker-krey:
[...]

PR #924, opened 2025-03-18
Fix historic vintages in map_tec_lifetime

macflo8:
This PR fixes the incorrect generation of a subset of map_tec_lifetime, that represents the pre first model year technology vintages, raised in #923.

The `map_tec_lifetime` set is generated in 3 steps:

https://github.com/iiasa/message_ix/blob/1edda554fc38d439073f6bdbd6195015cb29fb4d/message_ix/model/MESSAGE/data_load.gms#L122-L135

The last condition in the third block is (incorrectly) only checking the `technical_lifetime` of the first period instead of the respective historical year. 

### Background/workaround
This problem was circumvented in the default MESSAGEix-GLOBIOM parametrisation by adding `input` and `output` coefficients for `year_act` in historic years. If this is done, the `map_tec_lifetime` entries for the historic years are added in the second block since these entries are then included in `map_tec`.
This is also done in the Westeros tutorial by passing `in_horizon=False` to scenario.vintage_and_active_years().

### Things added in this PR

The initial version of the branch used modifications to two tutorial notebooks to highlight the bug, as follows:
> This PR removes this parametrisation from Westeros to highlight this bug. In addition to setting `in_horizon=True`, `westeros_historical_new_capacity.ipynb` is modified implement the example given in #923:
>
> - The `technical_lifetime` of coal_ppl is reduced to 10 years in [700, 710, 720], **while keeping it at 20 years for the historic year 690**.
> - The capacity additions in year 700 bounded to 0 with a `bound_new_capacity_up`
>
> Those two changes make the Westeros model of `westeros_historical_new_capacity.ipynb.ipynb` infeasible, even though `historical_new_capacity` is parametrized to satisfy the demand in the first model year without additional capacity. Thus, tests will fail (for now).

The final version does the following:

- A minimal test to reproduce the issue.
- Adopt the **proposed solution**, changing the third term of `map_tec_lifetime` to:
  ```
  * mapping of technology lifetime to all periods 'year_all' which were built prior to the beginning of the model horizon
  map_tec_lifetime(node,tec,historical,year_all)$( map_tec(node,tec,year_all) AND map_period(historical,year_all)
      AND historical_new_capacity(node,tec,historical)
      AND duration_period_sum(historical,year_all) < technical_lifetime(node,tec,historical) ) = yes ;
  ```
- Add explicit documentation of this parameter.
- Closes #932. See https://gist.github.com/khaeru/59e362941f3eb84afcf08d66d3f36c72 for discussion.

## How to review

[...]

PR #930, opened 2025-03-31
Refactor use of set type_tec in EMISSION_EQUIVALENCE

OFR-IIASA:
This PR addresses the issue described in issue #929 

## How to review

**Required:** describe specific things that reviewer(s) must do, in order to ensure that the PR achieves its goal.
If no review is required, write “No review:” and describe why.

## PR checklist

- [ ] Update release notes.

### DIFF
--- a/message_ix/model/MESSAGE/data_load.gms
+++ b/message_ix/model/MESSAGE/data_load.gms
@@ -81,6 +81,8 @@ fixed_extraction, fixed_stock, fixed_new_capacity, fixed_capacity, fixed_activit
 storage_initial, storage_self_discharge, time_order
 ;
 
+type_tec_share(type_tec) $ (SUM((shares, node_share, node, mode, commodity, level), map_shares_commodity_share(shares,node_share,node,type_tec,mode,commodity,level))) = YES ;
+type_tec_total(type_tec) $ (SUM((shares, node_share, node, mode, commodity, level), map_shares_commodity_total(shares,node_share,node,type_tec,mode,commodity,level))) = YES ;
 
 *----------------------------------------------------------------------------------------------------------------------*
 * ensure that each node is mapped to itself                                                                            *
--- a/message_ix/model/MESSAGE/model_core.gms
+++ b/message_ix/model/MESSAGE/model_core.gms
@@ -1836,7 +1836,7 @@ ACTIVITY_SOFT_CONSTRAINT_LO(node,tec,year,time)$( soft_activity_lo(node,tec,year
 *                   \text{ if } \widehat{t} \in \widehat{T}^{LAND} \Bigg)
 *
 ***
-EMISSION_EQUIVALENCE(node,emission,type_tec,year)..
+EMISSION_EQUIVALENCE(node,emission,type_tec,year) $ (NOT type_tec_share(type_tec) AND NOT type_tec_total(type_tec))..
     EMISS(node,emission,type_tec,year)
     =E=
     SUM(location$( map_node(node,location) ),
--- a/message_ix/model/MESSAGE/sets_maps_def.gms
+++ b/message_ix/model/MESSAGE/sets_maps_def.gms
@@ -310,8 +310,10 @@ Sets
     time_relative(time)                     flag for treating unit of ACT in sub-annual time slices relative to parent 'time' (activating parameter 'duration_time_rel')
 ;
 
-Alias(type_tec,type_tec_share);
-Alias(type_tec,type_tec_total);
+*Alias(type_tec,type_tec_share);
+*Alias(type_tec,type_tec_total);
+Sets type_tec_share(type_tec)
+     type_tec_total(type_tec) ;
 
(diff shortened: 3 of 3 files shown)


adrivinca:
Hi, to me it looks good apart from the last commit. I think the map_shares_commodity should stay there, otherwise the equation has no linkages to the commodity we are applying the shares to.
reverting the last commit the test seems to pass again

```
SHARE_CONSTRAINT_COMMODITY_UP(shares,node_share,year,time)$( share_commodity_up(shares,node_share,year,time) )..
* activity by type_tec_share technologies with map_shares_generic_share entries and a specific mode
    SUM( (node,location,type_tec_share,tec,vintage,mode,commodity,level,time2)$(
        ( map_shares_commodity_share(shares,node_share,node,type_tec_share,mode,commodity,level) OR
          map_shares_commodity_share(shares,node_share,node,type_tec_share,'all',commodity,level) ) AND
        cat_tec(type_tec_share,tec) AND
        map_tec_act(location,tec,year,mode,time2) AND
        map_tec_lifetime(location,tec,vintage,year)
    ),
        (
            output(location,tec,vintage,year,mode,node,commodity,level,time2,time) +
            input(location,tec,vintage,year,mode,node,commodity,level,time2,time)
        ) *
        duration_time_rel(time,time2) *
        ACT(location,tec,vintage,year,mode,time2)
    )
    =L=
    share_commodity_up(shares,node_share,year,time) * (
* total input and output by `type_tec_total` technologies mapped to respective commodity, level and node
    SUM( (node,location,type_tec_total,tec,vintage,mode,commodity,level,time2)$(
        ( map_shares_commodity_total(shares,node_share,node,type_tec_total,mode,commodity,level) OR
           map_shares_commodity_total(shares,node_share,node,type_tec_total,'all',commodity,level) ) AND
        cat_tec(type_tec_total,tec) AND
        map_tec_act(location,tec,year,mode,time2) AND
        map_tec_lifetime(location,tec,vintage,year)
    ),
        (
            output(location,tec,vintage,year,mode,node,commodity,level,time2,time) +
            input(location,tec,vintage,year,mode,node,commodity,level,time2,time)
        ) *
        duration_time_rel(time,time2) *
        ACT(location,tec,vintage,year,mode,time2)
    ) ) ;
```
not sure why we had always the map written twice with "mode" and "all". But that's separate
` map_shares_commodity_share(shares,node_share,node,type_tec_share,mode,commodity,level) OR`
`map_shares_commodity_share(shares,node_share,node,type_tec_share,'all',commodity,level) ) AND`

Also, do we need a test to check that the emission equation (or variable) is properly populated? Or at least it is not defined for the shares sets? I did not think yet how

OFR-IIASA:
@adrivinca the reason why the check takes place twice, once for `mode` and once for `all` is because when parametrizing, there is an option to apply the share constraint mode specifically or to all modes for which "all" is used, therefore allowing the user to pass the mode "all" which automatically applies the constraint across all modes instead of having to pass the modes individually. see for example https://docs.messageix.org/en/latest/model/MESSAGE/model_core.html#equation-activity-bound-all-modes-up ; interesting that this is also usable for shares; i didnt see any documentation for this explicitly.

adrivinca:
thanks, I see. But this way, with the OR condition, this will exclude the possibility to have the equation binding just for a specific mode, and a user might not be aware of it. It is probably a very minor case.

Anyway, if you agree that the map_shares can go back in the equation I revert the commit and also clean up the set definition file.

Still there is something about the test that does not convince me. It seems the test uses the commodity shares to impose mode shares, which it is not the proper use of those equations.
See line [245 of test_feature_bound_activity_shares.py](https://github.com/iiasa/message_ix/blob/8c0ca67e0d5adddf47a5a082c12f1f7e7447bb1a/message_ix/tests/test_feature_bound_activity_shares.py#L245)

usually the SHARE_CONSTRAINT_COMMODITY_UP/LO should set a share between multiple commodities and a subset of them (or similar reasoning in technologies defined in cat_tec that relate to dummy or real commodities).
The test only makes one commodity, do not impose shares between commodities. instead it plays with modes. I wonder if it derives from a time before the `SHARE_CONSTRAINT_MODE_UP`.

@khaeru you seem to be the last having edited the test, maybe not written it. But if someone confirms my doubts I can try edit it.
Cheers

adrivinca:
I was looking at the tests again, and they are actually ok. not as I would have framed at first though, but they are checking the right thing.
The test on share commodity did not have any emission set.
So I added a test within `test_commodity_share_up` that adds an emission factor, and check that the EMISS variable does not contain any of the type_tec defined for the shares.
This test fails under the previous versions and passes with this PR.

Hope it looks ok.
Thumbs up form my side for the GAMS changes and the PR

khaeru (approved):
Thanks both for these changes!

I made a number of clean-ups, also somewhat informed by #924, and added some migration notes that I hope make the impact clear. It seems unlikely that there will be many users in a position to be affected, but in any case we give them suggestions of what to do.

I'll merge as long as checks pass, or after resolving any failures.

PR #965, opened 2025-07-22
Fix: Corrected capitalization to GDXCOMPRESS in aux file

Tyler-lc:
## Short explanation 
Corrected capitalization in auxiliary_settings.gms, that prevented output files from being compressed in UNIX systems.

## Longer explanation
On case-sensitive file system (MacOS and Linux), GAMS is not compressing the output files as expected. This results in output gdx files that can be 3-4 times larger than on machines running Windows (which does not have a case sensitive file system).

The problem seems to be in a capitalization in the file "auxiliary_settings.gms". Currently we have:
"$SETENV GdxCompress 1"

However, it should be in all-caps:
"$SETENV GDXCOMPRESS 1"
The typo causes the command to be ignored on MacOS and Linux.

## How to review

- Possibly test if new capitalization works on Windows as well.

## PR checklist

- [ ] Update release notes.


khaeru:
> this PR doesn't really require a change in the documentation

Agreed! That's usually the case for bug fixes: the documentation already describes what the code *should do*; the bug is that the code *is not doing that*. So by fixing the bug, the existing docs again become a true statement of the behaviour.

> I'm still unsure how to update the release note.

Thanks for asking, and sorry the docs are not clear on that.

The file to be modified is RELEASE_NOTES.rst, [here](https://github.com/iiasa/message_ix/blob/main/RELEASE_NOTES.rst?plain=1). You can click ["🕑 History"](https://github.com/iiasa/message_ix/commits/main/RELEASE_NOTES.rst) in the top right to see the commit history for that file, which includes commits like 3e366ef99684e187695247780e6e510f6c3f053d. Try to make a similar commit with a similar message.

Some tips are:
- Put the entry in the section for "Next release" and subsection for "All changes".
- Use the same kind of grammar and wrapping as existing entries.
- Follow the existing style for linking to this PR at the end of the first sentence.
- Check that the entry looks OK in the built docs. 2 ways to do this:
  - Push to this PR branch and wait for ReadTheDocs to build the docs.
  - Follow the instructions to run `make html` in your local doc/ directory.

glatterf42 (approved):
Thanks for this fix :)
I rebased the branch on `main` and adapted the second commit message so that it starts capitalized. Now looks good to me 🚀

RELEASE_NOTES.rst
```diff
@@ -47,6 +47,7 @@ All changes
 
 - Document the :ref:`minimum version of Java <install-java>` required for :class:`ixmp.JDBCBackend <ixmp.backend.jdbc.JDBCBackend>` (:pull:`962`).
 - Improve type hinting (:pull:`963`).
+- Fix capitalization in in auxiliary_settings.gms to enable GDX output file compression on MacOS and Linux. (:pull:`965`)
```
khaeru, line 50: The word "in" appears twice here.

glatterf42: Thanks for spotting this :)

PR #983, opened 2025-10-21
Add modelinstance support 

Wegatriespython:
This is an experimental branch connected to [IXMP PR](https://github.com/iiasa/ixmp/pull/605) adds support for the [Gamsmodelinstance API](https://www.gams.com/50/docs/apis/python/classgams_1_1control_1_1execution_1_1GamsModelInstance.html)which provides a persistent model object for use.

Right now this is quite experimental, hence the code is not up to quality. The API for allowing user access is also not final. Putting it here for reference and feedback. For most models, model generation times dominate solving times. This means when a sequence of scenarios are to be evaluated in an ensemble a large fraction of the time is spent in the generation of the model matrix. Modelinstance provides a persistent object with an in-memory database for modifiable parameters via sync_db.

In theory this allows for gains = 1 + time_build/time_solving. For the baseline model time_build is ~500s and solve is 200s which yields a gain of 3.5X for large number of runs. Additionally the model instance passes the previous solution's basis to the next solution. In theory then a barrier followed by Simplex methods with advanced basis can approach even higher performance gains. But this requires dynamic solver parametrization maybe ML methods could help here.

## How to review

At this point feedback and usage would be appreciated 

## PR checklist
Experimental so this will not apply till at least an early Alpha stage is reached. 

~~- [ ] Continuous integration checks all ✅~~

~~- [ ] Add or expand tests; coverage checks both ✅~~
~~- [ ] Add, expand, or update documentation.~~
~~- [ ] Update release notes.~~

### DIFF
--- a/message_ix/core.py
+++ b/message_ix/core.py
@@ -760,6 +760,36 @@ class Scenario(ixmp.Scenario):
         """
         super().solve(model=model, solve_options=solve_options, **kwargs)
 
+    def create_model_instance(self, model="MESSAGE", **model_options):
+        """Create a persistent GAMS model instance for efficient resolving.
+
+        This is a thin wrapper around :meth:`ixmp.Scenario.create_model_instance`
+        that defaults to the MESSAGE model.
+
+        Parameters
+        ----------
+        model : str, optional
+            Model to compile. Default is "MESSAGE".
+        model_options :
+            Additional keyword arguments passed to the model.
+
+        Returns
+        -------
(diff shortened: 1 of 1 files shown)


glatterf42:
Same as https://github.com/iiasa/ixmp/pull/605#issuecomment-3637206761, I'd like to make sure that we don't create incompatibilities in our stack by having one official `main` version and an outdated version that we direct people to use (such as this branch). Do you know when you'll be able to continue here?

PR #881, opened 2024-10-15
Update Python support for 3.13

glatterf42:
As suggested by @khaeru in #880, we should drop support for the now outdated version 3.8 of Python and add support for the new 3.13. This PR does just that.

Please also see the analogous https://github.com/iiasa/ixmp/pull/544, which details some issues that need to be closed upstream before we can support 3.13.

## How to review

- Read the diff and note that the CI checks all pass.

## PR checklist

- [x] Continuous integration checks all ✅
- [x] Update tests; coverage checks both ✅
- [x] Add, expand, or update documentation.
- [x] Update release notes.

PR #985, opened 2025-10-24
Confirm support for Python 3.14 and drop 3.9

khaeru:
Python 3.14 [was released 2024-10-07](https://peps.python.org/pep-0745/). This PR is to:

- [x] Check/confirm that message_ix is compatible with the new version by updating CI workflows.
  - [x] Update the GitHub ruleset config to require 3.14 jobs to pass, instead of 3.13.
- [x] Check/confirm the existence of compatible releases of upstream dependencies. If those releases are not immediately available, the PR can aid to monitor and track upstream issues.
  - [x] iiasa/ixmp4#209 and further upstream. At first glance ([here](https://github.com/iiasa/message_ix/actions/runs/18774447220/job/53565787905?pr=985#step:10:23863)), there are 116 FAIL and 56 ERROR results due to this. The same mark `ixmp4_209` and handling used in iiasa/ixmp#602 are used to mark these as XFAIL.
- Drop support for Python 3.9, which reached end-of-life (no more updates, even security updates) [on 2025-10-07](https://peps.python.org/pep-0569/#lifespan).
  - [x] Remove any code marked as for Python 3.9 compatibility only. Depending on these changes, part or all of the functionality may remain usable with Python 3.9 if users have extreme need.
  - [x] Make any general clean-ups or improvements possible with [features available from Python 3.10](https://docs.python.org/3/whatsnew/3.10.html).
    - [x] Use the type union operator instead of typing.Optional and typing.Union. This is achieved by running `ruff check --select F401,UP007,UP045 --fix` or similar multiple times, plus some small manual clean-ups. See [the ruff docs](https://docs.astral.sh/ruff/rules/#pyupgrade-up).
    - [x] Use `match`/`case` instead of some `if` statements with multiple `elif` blocks.

Analogous to #881.

## How to review

- Read the diff,
- Note that the CI checks all pass.

## PR checklist

- [x] Update release notes.

After approval, before merge:
- [ ] Rebase, dropping TEMPORARY commit(s).

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -22,6 +22,8 @@ Adjust any imports like the following:
 All changes
 -----------
 
+- :mod:`message_ix` is tested and compatible with `Python 3.14 <https://www.python.org/downloads/release/python-3140/>`__ (:pull:`985`).
+- Support for Python 3.9 is dropped (:pull:`985`), as it has reached end-of-life.
 - Add representation of commodity flows associated with construction and retirement of technology capacity (:pull:`451`).
 
   - New parameters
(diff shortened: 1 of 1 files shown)


glatterf42 (approved):
Thanks for these cleanups! I think the test failures are flaky, but I've triggered a rerun to be sure. Since all tests are passing here, I think this PR includes the fixes for our nightly CI failures, so it would be great to merge this today :)

PR #986, opened 2025-11-21
Fix `Scenario` errors in `add_year()`

### LINKED ISSUE #578: `add_year()` as implemented results in scenario errors
Observation: Using `add_year()` on the current main branch results in scenarios with misaligned `input`, `output` and `technical_lifetime parameters`. This can result in completely unexpected results. Below, for example, is the result of a high-carbon-tax scenario applied to an existing 5-year timestep model and the same model after using `add_year()` to add 1-year timesteps in the first 5-year period.

(this is an updated image originally published in a comment below but is relevant for new readers of this issue)
image

There is a believed to be correctly working version on [this branch](https://github.com/behnam-zakeri/message_ix/tree/test-add-year/message_ix/tools/add_year).

Immediate solution: Raise a `NotImplementedError` in `main`'s implementation

Eventual solution: Migrate the correctly-working version into the `main` branch.

glatterf42:
Closes #578.
In that issue, @gidden mentioned that [this branch by Behnam](https://github.com/behnam-zakeri/message_ix/tree/test-add-year/message_ix/tools/add_year) contained a fix for the issue. However, that branch has become severely outdated over the years. So after rebasing #494, I created a new branch here that captures the changes contained in @behnam-zakeri's branch. 

I'm not sure what's still missing in #494, but I think this could be a good opportunity to move things along. And the assessments in #578 made me think that things need to move along to address these known bugs.

## How to review

## PR checklist

- [ ] Update release notes.

### DIFF
--- a/message_ix/tools/add_year/__init__.py
+++ b/message_ix/tools/add_year/__init__.py
@@ -790,7 +790,7 @@ def interpolate_2d(  # noqa: C901
     value_col: str = "value",
     extrapolate: bool = False,
     extrapol_neg: float | None = None,
-    year_diff: list[int] | None = None,
+    year_diff: list[int] = [],
     bound_extend: bool = True,
 ):
     """Interpolate parameters with two dimensions related year.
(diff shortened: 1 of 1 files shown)


glatterf42:
Ideally, this PR would add a test that confirms that the fix here actually does what we intend it to do.

PR #1023, opened 2026-06-10
Fix slow ACT.lo assignment during model generation

### LINKED ISSUE #1022: Diagnose super-linear increase in model generation time at high node count
### What is this about?

When running NEXUS module which adds basins per `R12` region, the model generation time grows disproportionately, 

 Measured by running `MESSAGE_run.gms` directly on saved input GDX files (no Python/ixmp in the loop), perfect-foresight, `profile=1`, solving is intentionally skipped : 

| Scenario | Nodes | LP size | GDX read | execution phase | matrix generation | total to solver |
|---|---|---|---|---|---|---|
| R12 global scenario | ~14 | 1.85 M rows, 8.7 M nz | ~5 s | **~60–80 s** | ~35–43 s | ~2–2.5 min |
| Same + water-nexus module, full basin resolution | ~225 | 21.2 M rows, 29.8 M nz | ~6 s | **~6,330–6,480 s** | ~190–285 s | ~111–115 min |

Using the profile summary for the large case we get 

```
---- Profile Summary (large case, total 6,668 s)
  6358.048   1.457GB      2621 Assignment ACT (697719)
   102.699   8.245GB      4912 Solver     MESSAGE_LP
    46.882   8.047GB      4401 Equation   EMISSION_EQUIVALENCE (19306980)
```
the majority of the time is spent on this statement : 

`message_ix/model/MESSAGE/model_core.gms` lines 186–187, the inter-vintage-arbitrage guard:

```gams
ACT.lo(node,tec,year_all,year_all2,mode,time)$( map_tec_lifetime(node,tec,year_all,year_all2)
    AND map_tec_act(node,tec,year_all2,mode,time) AND bound_activity_lo(node,tec,year_all2,mode,time) >= 0 ) = 0 ;
```
 The culprit appears to be the relational condition: `bound_activity_lo(...) >= 0` is satisfied by *absent* records (0 ≥ 0), so GAMS cannot drive the assignment from sparse storage and falls back to enumerating a much larger space that blows up with the node dimension.

A similar state : the assignment a few lines below (lines 195–196) has the identical two-set shape without the relational test does not appear in the profile at all (< 1 s at `profiletol=1`) in either case.

```gams
ACT.lo(node,inv_tec,year_all,year_all2,mode,time)$( map_tec_lifetime(node,inv_tec,year_all,year_all2)
    AND map_tec_act(node,inv_tec,year_all2,mode,time) ) = 0 ;
```
Environment: observed on GAMS 48.6 (Linux); also reproduced on GAMS 46.5; `message_ix` current `main`.

Wegatriespython:
Fix the `ACT` non-negativity default assignment so GAMS model generation no longer scales super-linearly with node count. Closes #1022.

The inter-vintage-arbitrage guard set `ACT.lo = 0` under the condition `bound_activity_lo(...) >= 0`. On production inputs GAMS evaluates that condition over the full declared index space rather than the `map_tec_lifetime` × `map_tec_act` intersection (see the review discussion for what triggers this). The condition is now split: the default is assigned over the set intersection alone, and tuples with an explicitly negative `bound_activity_lo` are reset to `-INF` in a second assignment. Under `AUX_BOUNDS` those tuples receive `-AUX_BOUND_VALUE`, as before.

On the inputs from #1022, the assignment drops from 6,358 s to under 10 s at ~225 nodes and from 56 s to under 2 s at R12; total time to solver handoff at ~225 nodes drops from ~111 min to ~5 min. The generated model is unchanged: identical row, column and nonzero counts at both scales, `gdxdiff`-identical `ACT` bounds (459,751 records at R12, including 323 negative `bound_activity_lo` entries, with and without `AUX_BOUNDS`), and identical objective values.

## How to review

- Read the diff in `model_core.gms` and confirm the assignments reproduce the previous semantics: `ACT.lo = 0` where `bound_activity_lo` is zero or absent, `-INF` where it is negative, and `-AUX_BOUND_VALUE` for the negative case under `AUX_BOUNDS`.

## PR checklist

- ~Add, expand, or update documentation.~ Only a code comment changed; the `*` comment block is not rendered in the docs.
- [x] Update release notes.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -22,6 +22,7 @@ Adjust any imports like the following:
 All changes
 -----------
 
+- Reduce GAMS model generation time for scenarios with many nodes by assigning the non-negativity default on ``ACT`` through sparse set iteration (:pull:`1023`).
 - :mod:`message_ix` is tested and compatible with `Python 3.14 <https://www.python.org/downloads/release/python-3140/>`__ (:pull:`985`).
 - Support for Python 3.9 is dropped (:pull:`985`), as it has reached end-of-life.
 - :mod:`message_ix` is tested and compatible with `Pandas 3.0.0 <https://pandas.pydata.org/pandas-docs/stable/whatsnew/v3.0.0.html>`_,
--- a/message_ix/model/MESSAGE/model_core.gms
+++ b/message_ix/model/MESSAGE/model_core.gms
@@ -182,9 +182,18 @@ Variables
 %AUX_BOUNDS%    AND map_tec_act(node,tec,year_all2,mode,time) ) = %AUX_BOUND_VALUE% ;
 
 * to avoid "inter-vintage arbitrage" (across different vintages of technologies), all activities that
-* have positive lower bounds are assumed to be non-negative
+* have positive lower bounds are assumed to be non-negative; the test on bound_activity_lo is kept out
+* of the assignment condition because a relational test cannot drive sparse iteration
+Set is_bound_activity_lo_negative(node,tec,year_all,mode,time) "flag for explicitly negative lower bounds on activity";
+is_bound_activity_lo_negative(node,tec,year_all,mode,time)$( bound_activity_lo(node,tec,year_all,mode,time)
+    AND bound_activity_lo(node,tec,year_all,mode,time) < 0
+    AND map_tec_act(node,tec,year_all,mode,time) ) = yes ;
 ACT.lo(node,tec,year_all,year_all2,mode,time)$( map_tec_lifetime(node,tec,year_all,year_all2)
-    AND map_tec_act(node,tec,year_all2,mode,time) AND bound_activity_lo(node,tec,year_all2,mode,time) >= 0 ) = 0 ;
+    AND map_tec_act(node,tec,year_all2,mode,time) ) = 0 ;
+ACT.lo(node,tec,year_all,year_all2,mode,time)$( map_tec_lifetime(node,tec,year_all,year_all2)
+    AND is_bound_activity_lo_negative(node,tec,year_all2,mode,time) ) = -INF ;
+%AUX_BOUNDS% ACT.lo(node,tec,year_all,year_all2,mode,time)$( map_tec_lifetime(node,tec,year_all,year_all2)
+%AUX_BOUNDS%    AND is_bound_activity_lo_negative(node,tec,year_all2,mode,time) ) = -%AUX_BOUND_VALUE% ;
 * previous implementation using upper bounds
 * ACT.lo(node,tec,year_all,year_all2,mode,time)$( map_tec_lifetime(node,tec,year_all,year_all2)
 *    AND map_tec_act(node,tec,year_all2,mode,time)


khaeru (commented):
Thanks for the change here and the measurements of the original performance problem (#1022)/improvement on a full-size model.

Do you have a specific reference to the GAMS documentation for "sparse set iteration"? I've re-read the pages:
- [Conditional Expressions, Assignment and Equations](https://www.gams.com/54/docs/UG_CondExpr.html)
- [Data Manipulation with Parameters](https://www.gams.com/latest/docs/UG_Parameters.html)

…and don't find explicit description of this.

The reason is I'm wondering if the implementation can be simpler and thus more easily documented and understood.

Consider the prior code (1). I label each of the conditions:
```
ACT.lo(node,tec,year_all,year_all2,mode,time)$(
  map_tec_lifetime(node,tec,year_all,year_all2)             * 1A
  AND map_tec_act(node,tec,year_all2,mode,time)             * 1B
  AND bound_activity_lo(node,tec,year_all2,mode,time) >= 0  * 1D
) = 0 ;
```

On this branch you've written (2):
```
is_bound_activity_lo_negative(node,tec,year_all,mode,time)$(
  map_tec_act(node,tec,year_all,mode,time)                * 2B
  AND bound_activity_lo(node,tec,year_all,mode,time)      * 2C
  AND bound_activity_lo(node,tec,year_all,mode,time) < 0  * 2D
) = yes ;
...
ACT.lo(node,tec,year_all,year_all2,mode,time)$(
  map_tec_lifetime(node,tec,year_all,year_all2)           *2A
  AND is_bound_activity_lo_negative(node,tec,year_all2,mode,time)
) = -INF ;
```

This should be *functionally* equivalent to (3):
```
ACT.lo(node,tec,year_all,year_all2,mode,time)$(
  map_tec_lifetime(node,tec,year_all,year_all2)           * 3A
  AND map_tec_act(node,tec,year_all,mode,time)            * 3B
  AND bound_activity_lo(node,tec,year_all,mode,time)      * 3C
  AND bound_activity_lo(node,tec,year_all,mode,time) < 0  * 3D
) = -INF ;
```

Shouldn't (3) and (2) also be equally *performant*?

If not, then the way I would understand this (again, in the absence of a specific description in the docs) is:
- AND conditions are evaluated in order, processing stops once a FALSE is obtained.
- Thus 2C/3C are only evaluated if 2A/3A and 2B/3B are TRUE.
- 2C/3C are FALSE except in the very small number of cases (n, t, y_a, m, t) where bound_activity_lo has a defined value, so in most cases the overall condition is FALSE.
- The numerical comparison (2D/3D) is only evaluated in the small number of cases where 2C/3C are TRUE.
- The flaw in (1) is that there is no condition 1C, and so 1D is evaluated for every case where 1A and 1B, inserting 0 for the (many, many) missing values in bound_activity_lo.

If this is a correct understanding, then I think (3), without a new set, would be more clear.


While studying this, I also notice that the earlier/unmodified code in this section has some redundancy. For instance:
```
* include upper and lower bounds (to avoid unbounded models)
%AUX_BOUNDS% ACT.lo(node,tec,year_all,year_all2,mode,time)$( 
```
…has the same conditional as the modified line that sets 0 values for ACT.lo. So the former are always overwritten and can be removed.

I will push a commit or two making these changes and perhaps adding a test. Then if you could confirm that the fix is effective, we can merge.

khaeru:
It seems we need changes on message_ix corresponding to iiasa/message-ix-models#550. I will make those changes in a separate PR, and then rebase this one.

khaeru:
> I will make those changes in a separate PR, and then rebase this one.

Done via #1035.

Wegatriespython:
Thanks for the review. I tested the changes and found your version to be as performant: on the R12 input behind #1022, the inline form and the flag-set form both run in about 0.1 s against 35 s for the original `>= 0` statement, and generation time on the full ~225-node input is unchanged between them. That meant the operating hypothesis in the PR was wrong, so I tried removing the bare `bound_activity_lo(...)` guard as well; the perf boost stayed (0.07 s without it, 0.11 s with it). So I will reword the title and body.

Checking further, the actual culprit is more specific: it is the `-INF` values stored in `bound_activity_lo` for the `*_TCE` emission-accounting technologies in mainline scenarios. If you rewrite those to -1 and keep everything else identical, the original `>= 0` statement drops from 30 s to 0.08 s. This also explains why the slowdown was hard to reproduce without real scenario data.

When the parameter holds a `-INF` and the condition contains a comparison that the parameter's default satisfies (`>= 0`; `NOT (x < 0)` behaves identically), GAMS stops iterating the ~386k tuples the two maps hold and walks the full declared six-index product. Clause order makes no difference.

For the `-INF` values themselves (incidental connection to the inf discussion on  iiasa/ixmp4#266), the other bounds use `is_bound_*` flag sets, but `bound_activity_lo` however defaults to 0, so "no lower bound" for the TCE technologies can only be stored as an explicit `-INF`. That is harmless in ACTIVITY_BOUND_LO, but the ACT.lo assignment reads the parameter directly in its condition. Maybe worth extending the flags so the assignment conditions on sets only, then it is defended the same way the equations are. The behaviour is identical on GAMS 46.5.0, 48.6.0 and 51.4.0; given the evidence, this looks like a GAMS oddity around special values rather than a general rule about relational tests.

I see the rewritten block has two other changes separate from performance. When re-running the checks from the PR against your branch, the ACT.lo bound comparison on the R12 input is no longer identical: 32 tuples differ, all `loil_ppl` in R12_LAM, R12_CHN and R12_RCPA at year 2020 across vintages, `0` on main and `-INF` on your branch. They are the three records in this input with a small negative `bound_activity_lo` on an `inv_tec`; on main the trailing `ACT.lo(node,inv_tec,...) = 0` ran last and re-imposed the zero floor on them, and your branch drops that line. The new test cannot catch it because make_dantzig has no inv_tec.

Second, the PR's `AUX_BOUNDS` check no longer runs on your branch: enabling it fails with "Set has not been initialized" at the `AND map_tec_act(...)` line of block (4), and the error is absent with it off. It looks like the `map_tec_lifetime(n,t,vintage,y_all)` operand went missing from that condition.

khaeru:
It's not clear from this reply whether you read the question near the start of my review. I don't see a direct response.

> Checking further, the actual culprit is more specific: [...] this looks like a GAMS oddity around special values rather than a general rule about relational tests.
>
> all loil_ppl in R12_LAM, R12_CHN and R12_RCPA at year 2020

This sounds like it is about specific MESSAGEix-GLOBIOM scenarios created with message-ix-models or other code. To be clear, per separation of concerns, message_ix, ixmp, and/or ixmp4 are not the place to handle these; they are rather modeling errors or dependencies on ultimately broken or buggy behaviour. It *is* appropriate that we note and warn users (e.g. in a "migration notes" section in the docs, e.g. [here](https://docs.messageix.org/en/latest/whatsnew.html#id7)) about possible implications of the GAMS code changes.

The GAMS documentation ("Logical Conditions", "Extended Range Arithmetic", and "Logical Functions") seems to be explicit and consistent that a not-set value in any parameter is treated as 0 for the purposes of numerical comparisons, and that explicit EPS, +INF, and -INF are treated as not zero. This has been the behaviour for as long as I've been aware of GAMS.

> Second, the PR's AUX_BOUNDS check no longer runs on your branch:

Which check? AFAICT we have [no tests that turn on AUX_BOUNDS](https://github.com/search?q=repo%3Aiiasa%2Fmessage_ix+AUX_BOUNDS&type=code), and the PR does not contain tests other than the one I added for you.

Wegatriespython:
To answer the first question, no. The term was the working hypothesis, not based on specific gams documentation. The hypothesis was based on a now falsified model of how gams works. The nearest documentation is the User's Guide performance page, which recommends "using dollar conditions and filtering sets" to restrict assignments to the relevant cases [Finding and Fixing Execution Errors and Performance Problems](https://www.gams.com/latest/docs/UG_ExecErrPerformance.html#UG_ExecErrPerformance_ReducingGAMSExecutionTime). GAMS's seems to call the mechanism for faster execution as "sparse execution", per the[ 51.1.0 release notes ](https://www.gams.com/51/docs/RN_51.html#g5110_GAMSSYSTEM). 
 
Your forms (2) & (3) are equally as fast. 

Semantics notwithstanding, GAMS execution path logic is still opaque. 

Second on the separation of concerns point. In principle that makes sense, in practice the current GAMS code guards on input data in many places. 

But the dropping `ACT.lo(node,inv_tec,...) = 0`, would effect every current clone workflow R12 scenario. A migration guide while correct in principle, might not get adopted and if users see `message_ix` not support their scenario the default response is to roll back to an older (incorrect) version of `message_ix`. In the absence of clearing iiasa/message-ix-models#551, most scenarios are not created from code. Fixing the errant code alone will not resolve the database wide issue. More critically even if all scenarios database were to be fixed, there would be no way of checking errant code unless we had a strict policy merged code only can produce a scenario. I think restoring that line would be more pragmatic than not. 

Last, the sentence on AUX_BOUNDS check referred to the sentence in the PR body, which addressed that the change did not modify gams behaviour unintentionally. 

You can reproduce the check by running : 

```
cd message_ix/model
printf '$setglobal AUX_BOUNDS\n$include MESSAGE_run.gms\n' > aux_on.gms
gams aux_on.gms --in=data/MsgData_<scenario>.gdx --out=output/MsgOutput_<scenario>.gdx
```

It will show 

```
ACT.lo(n,t,vintage,y_all,m,time)$(
  AND map_tec_act(n,t,y_all,m,time)
****       $140        $8,36
****   8  ')' expected
****  36  '=' or '..' or ':=' or '$=' operator expected
**** 140  Unknown symbol`
```

